PR #4, opened 2018-11-19
all: use New() constructor function for simplicity

deadprogram:
Use New() constructor function for devices for simplicity sake. Resolves https://github.com/aykevl/tinygo-drivers/issues/1


aykevl:
Thank you!

PR #5, opened 2018-11-20
Add BlinkM I2C RGB LED

deadprogram:
This PR adds support for the BlinkM I2C RGB LED.

### DIFF
--- a/blinkm/blinkm.go
+++ b/blinkm/blinkm.go
@@ -21,10 +21,10 @@ func New(bus machine.I2C) Device {
 }
 
 // Version returns the version of firmware on the BlinkM.
-func (d Device) Version() ([]byte, error) {
+func (d Device) Version() (major, minor byte, err error) {
 	version := []byte{0, 0}
 	d.bus.ReadRegister(Address, GET_FIRMWARE, version)
-	return version, nil
+	return version[0], version[1], nil
 }
 
 // SetRGB sets the RGB color on the BlinkM.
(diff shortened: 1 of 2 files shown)


blinkm/blinkm.go
```diff
@@ -0,0 +1,54 @@
...
+// This function only creates the Device object, it does not touch the device.
+func New(bus machine.I2C) Device {
+	return Device{bus}
+}
+
+// Version returns the version of firmware on the BlinkM.
+func (d Device) Version() ([]byte, error) {
+	version := []byte{0, 0}
+	d.bus.ReadRegister(Address, GET_FIRMWARE, version)
+	return version, nil
```
aykevl: This causes a heap allocation. You can avoid that by returning the two numbers itself (`return version[0], version[1], nil`) or as a single integer (`return uint16(version[0]) << 8 | uint16(version[1]), nil`).
Many libraries expose a version number as a single number, for example, FastLED is at version 3002001 at the moment (meaning 3.2.1). Such a version number is also easier to compare.

deadprogram: Oddly the return values are actually letters example `a e` not numbers. But probably the same idea still applies:

```
func (d Device) Version() (major, minor byte, err error)
```

What do you think?

aykevl: Sounds good!

deadprogram: Done.

aykevl:
Merged in 9d863fdf15222c8e4309b197a5b1f6c9754fe38a.

PR #6, opened 2018-12-18
initial support for PCD8544 LCD controller 

conejoninja:


aykevl:
Thank you for the pull request! I will take a proper look soon.

pcd8544/pcd8544.go
```diff
@@ -0,0 +1,119 @@
...
+		rstPin: rstPin,
+		scePin: scePin,
+		buffer: [504]byte{},
+	}
+}
+
+// Init initializes the LCD with default configuration
+func (d *Device) Init() {
+	d.rstPin.Low()
+	time.Sleep(500 * time.Millisecond)
```
aykevl: I don't think this is necessary on most chips, and 500ms is certainly way more than required. The required reset time is at least 100ns. As peripheral writes are generally 2 cycles on Cortex-M, this means that not sleeping at all between .Low() and .High() translates into at least 125ns for a 16MHz chip (possibly more depending on the compiler and optimization level).
To future proof this library, I would recommend simply sleeping for 100ns literally, so `time.Sleep(100 * time.Nanosecond)`. Or maybe a little bit more if you want to be sure. That avoids half a second of startup time.

pcd8544/pcd8544.go
```diff
@@ -0,0 +1,119 @@
...
+		bus:    bus,
+		dcPin:  dcPin,
+		rstPin: rstPin,
+		scePin: scePin,
+		buffer: [504]byte{},
+	}
+}
+
+// Init initializes the LCD with default configuration
+func (d *Device) Init() {
```
aykevl: Can you rename this function to `Configure()`? That's the same name that is used by the other drivers, for consistency.

pcd8544/pcd8544.go
conejoninja: I did not appreciate any difference between using more or less time (or any at all), it looked too much for me too but Adafruit's pcd8544 library has a 500ms delay. Will change this.

conejoninja:
Thanks for the suggestions. I tried to find a balance between a generic API and not wasting too much resources, but I see how useful could be to have the same API for different displays. Will make it as compatible as possible with the one of periph, even it's a bit customized too. Something as complete as Adafruit's GFX will be a different beast.

examples/pcd8544/setbuffer/main.go
```diff
@@ -0,0 +1,49 @@
...
+
+	dcPin := machine.GPIO{4}
+	dcPin.Configure(machine.GPIOConfig{Mode: machine.GPIO_OUTPUT})
+	rstPin := machine.GPIO{5}
+	rstPin.Configure(machine.GPIOConfig{Mode: machine.GPIO_OUTPUT})
+	scePin := machine.GPIO{17}
+	scePin.Configure(machine.GPIOConfig{Mode: machine.GPIO_OUTPUT})
+
+	machine.SPI0.Configure(machine.SPIConfig{
+		Frequency: 328125,
```
deadprogram: I think this baud rate will be ignored and instead use the default of `500000`.

aykevl:
Also, a more generic comment about this PR: I'm not sure that the driver should keep a buffer with the display data. Ideally, drivers only do as much as necessary to interface with the hardware. You can build a library (like GFX) on top of that which implements more advanced things and supports all displays implementing a given interface, possibly using typeasserts to check for special features.

conejoninja:
I wasn't sure either, the library I used as a base had the setPixel/getPixel methods that needed the buffer. It would be needed if we want to implement something like [ssd1306.Draw(r image.Rectangle, src image.Image, sp image.Point)](https://godoc.org/periph.io/x/periph/devices/ssd1306#Dev.Draw) or Scroll() without artifacts or erasing the rest of the image.

Don't have much experience interfacing with displays, so I'm not sure which functions you should expect, or if set/getPixel functions are useful at all, for complex animations you probably are going to implement your own buffer. I'm open to suggestions.

deadprogram:
It seems like this PR needs a bunch of updating to bring it in line with the latest TinyGo, as well as conversation about display interfaces. But there is also a bunch of good and useful code here. Probably worth the effort to update and land it.

conejoninja:
It's on my todo list, I did not forget about this. I was waiting for a more defined display interface, which we have now

deadprogram:
Any reason why we should not merge this now? Seems like it is ready.

deadprogram:
Actually there is one thing still. I recommend rebase master, then add a smoke test like https://github.com/tinygo-org/drivers/blob/master/.circleci/config.yml#L22

conejoninja:
make the commit clearer from a recent _master_. Added smoke tests and the device to the README's list

deadprogram:
Thanks, @conejoninja you are the display expert now, for sure!

Merging.

PR #7, opened 2018-12-22
apa102: add support for APA102 SPI controlled LEDs

deadprogram:
This PR adds support for APA102 SPI controlled LEDs, such as the Adafruit dotStar and other similar devices. Here is a sample program that creates a holiday lights display using a strip of 60 LEDS:

```go
// Connects to an APA102 SPI RGB LEDs.
package main

import (
	"image/color"
	"machine"
	"time"

	"github.com/aykevl/tinygo-drivers/apa102"
)

func main() {
	machine.SPI0.Configure(machine.SPIConfig{
		Frequency: 500000,
		Mode:      0})

	a := apa102.New(machine.SPI0, 60)
	leds := make([]color.RGBA, 60)
	rg := false

	for {
		rg = !rg
		for i := range leds {
			rg = !rg
			if rg {
				leds[i] = color.RGBA{R: 0xff, G: 0x00, B: 0x00, A: 0x77}
			} else {
				leds[i] = color.RGBA{R: 0x00, G: 0xff, B: 0x00, A: 0x77}
			}
		}

		a.WriteColors(leds)
		time.Sleep(100 * time.Millisecond)
	}
}
```

### DIFF
--- a/apa102/apa102.go
+++ b/apa102/apa102.go
@@ -22,14 +22,12 @@ const (
 // Device wraps APA102 SPI LEDs.
 type Device struct {
 	bus   machine.SPI
-	count int
 	Order int
 }
 
-// New returns a new APA102 driver. Pass in a fully configured SPI bus, and the count of
-// APA102 LEDs that are connected together.
-func New(b machine.SPI, count int) Device {
-	return Device{bus: b, count: count, Order: BGR}
+// New returns a new APA102 driver. Pass in a fully configured SPI bus.
+func New(b machine.SPI) Device {
+	return Device{bus: b, Order: BGR}
 }
 
 // WriteColors writes the given RGBA color slice out using the APA102 protocol.
@@ -59,7 +57,7 @@ func (d Device) WriteColors(cs []color.RGBA) (n int, err error) {
 		}
 	}
 
-	d.endFrame()
+	d.endFrame(len(cs))
 
 	return len(cs), nil
 }
@@ -68,17 +66,21 @@ func (d Device) WriteColors(cs []color.RGBA) (n int, err error) {
 func (d Device) Write(buf []byte) (n int, err error) {
 	d.startFrame()
 	d.bus.Tx(buf, nil)
-	d.endFrame()
+	d.endFrame(len(buf) / 4)
 
 	return len(buf), nil
 }
 
+// startFrame sends the start bytes for a strand of LEDs.
 func (d Device) startFrame() {
 	d.bus.Tx([]byte{0x00, 0x00, 0x00, 0x00}, nil)
 }
 
-func (d Device) endFrame() {
-	for i := 0; i < (d.count+15)/16; i++ {
+// endFrame sends the end frame marker with one extra bit per LED so
+// long strands of LEDs receive the necessary termination for updates.
+// See https://cpldcpu.wordpress.com/2014/11/30/understanding-the-apa102-superled/
+func (d Device) endFrame(count int) {
+	for i := 0; i < count/16; i++ {
 		d.bus.Tx([]byte{0xff}, nil)
 	}
 }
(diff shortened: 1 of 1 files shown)


aykevl:
Unfortunately, the interface of this LED strip is not the same as the one of the ws2812 driver. I'm now trying to unify them: using color.RGBA is a good idea.

apa102/apa102.go
```diff
@@ -0,0 +1,75 @@
...
+			d.tx[i*4+2] = byte(c.G)
+			d.tx[i*4+3] = byte(c.R)
+		}
+	}
+
+	return d.Write(d.tx)
+}
+
+// Write the raw bytes using the APA102 protocol.
+func (d Device) Write(buf []byte) error {
```
aykevl: This does not implement the io.Writer interface: the return value must be `(int, error)`.
(I know the ws2812 driver has a similar error, I'll send a PR to fix it there as well).

apa102/apa102.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// WriteColors writes the given RGBA color slice out using the APA102 protocol.
+// The A value (Alpha channel) is used for brightness, set to 0xff (255) for maximum.
+func (d Device) WriteColors(cs []color.RGBA) error {
+	for i, c := range cs {
+		d.tx[i*4] = 0xe0 | (c.A >> 3) // brightness is scaled to 5 bit value
+		switch d.Order {
+		case BRG:
+			d.tx[i*4+1] = byte(c.B)
```
aykevl: `byte` and `uint8` are aliases, so this conversion is not necessary.

apa102/apa102.go
```diff
@@ -0,0 +1,75 @@
...
+	bus   machine.SPI
+	tx    []byte
+	count int
+	Order int
+}
+
+// New returns a new APA102 driver. Pass in a fully configured SPI bus, and the count of
+// APA102 LEDs that are connected together.
+func New(b machine.SPI, count int) Device {
+	t := make([]byte, count*4)
```
aykevl: I would really like to see this allocation gone, as RAM is scarce and allocations are expensive. I think it can be done at the cost of a bit of code duplication. Or is there a reason why a buffer has to be allocated first?

apa102/apa102.go
deadprogram: Good idea. Change made.

apa102/apa102.go
deadprogram: Done.

apa102/apa102.go
deadprogram: OK I made change to accommodate this idea. This will certainly help for the case that larger numbers of LEDs are being used.

deadprogram:
Please see updates to accommodate suggested changes.

apa102/apa102.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return len(buf), nil
+}
+
+func (d Device) startFrame() {
+	d.bus.Tx([]byte{0x00, 0x00, 0x00, 0x00}, nil)
+}
+
+func (d Device) endFrame() {
+	for i := 0; i < (d.count+15)/16; i++ {
```
aykevl: You can avoid the need for the .count element by passing the length into this function. From WriteColors it is simply `len(cs)` and in Write it is `len(buf)/4`.

deadprogram: Another good change. Updated.

aykevl:
Thank you! Merged in 773b01e911bc8883459b4ec0afb113a5a3748910.

aykevl:
I'm afraid I missed one thing: WriteColors returns `(int, error)` in this PR while the ws2812 WriteColors only returns `error`. This makes it impossible to use the same interface.

I think returning the number of written bytes doesn't make a lot of sense, it either succeeds or it fails. Also, in many cases there is not much you can do when it fails anyway: how would you recover from a partially-written LED strip? The only sane thing would be to just send it again, possibly resetting it first if needed.

What do you think about changing the method signature slightly?

deadprogram:
That is probably also true with the `Write()` method but we still comply with the interface. I do not have a strong opinion either way on this one. Do you want another PR that changes that signature?

PR #8, opened 2018-12-22
WS2812 interface changes

aykevl:
A few changes to make WS2812 compatible with APA102 (see #7).
@deadprogram what do you think of this interface?
In my code, code size is slightly increased. I'm not entirely sure why, but I think it's because some extra bit shifts and masks are necessary to insert values into and extract values from `image/color.RGBA`. I still think the change is worth it, though, for an easier-to-read interface.

### DIFF
--- a/ws2812/ws2812.go
+++ b/ws2812/ws2812.go
@@ -2,6 +2,7 @@
 package ws2812
 
 import (
+	"image/color"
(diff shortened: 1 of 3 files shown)


deadprogram:
Seems like good changes to me. Merge on.

PR #10, opened 2018-12-22
README: add README with project scope and contributing guidelines

aykevl:
@deadprogram what do you think, especially of the project scope? I think it's useful to decide such things from the beginning so everyone knows what is and is not in scope.

Also, I just realize there is no license yet, that should be the same one as the TinyGo license (3-clause BSD).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,30 @@
+# TinyGo drivers
+
+[![GoDoc](https://godoc.org/github.com/aykevl/tinygo-drivers?status.svg)](https://godoc.org/github.com/aykevl/tinygo-drivers)
+
+This package provides a collection of hardware drivers that can be used together
+with [TinyGo](https://github.com/aykevl/tinygo).
+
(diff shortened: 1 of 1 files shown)


deadprogram:
I think there is simple way to state "These drivers provide low-level access to the device functionality." 

I do not agree with the statement that "many of these drivers are not meant to
be used directly". They are in fact meant to do exactly that IMO, and nothing else.

deadprogram:
I used the basic idea but created an entirely different README. Closing this PR, please open another with any edits.

aykevl:
Which PR? This is for the drivers repo, which doesn't have a README yet.

deadprogram:
I added a README already, is what I am saying in https://github.com/tinygo-org/drivers/pull/10#issuecomment-461326090

aykevl:
Oops, didn't look good enough... Sorry about that.

PR #11, opened 2019-01-24
Initial support for BMP180 digital pressure sensor

conejoninja:
Based on https://github.com/hybridgroup/gobot/blob/master/drivers/i2c/bmp180_driver.go

Added a examples/bmp180 folder with a tiny example

### DIFF
--- a/bmp180/bmp180.go
+++ b/bmp180/bmp180.go
@@ -12,17 +12,6 @@ import (
 	"github.com/aykevl/tinygo/src/machine"
 )
 
-const (
-	// ULTRALOWPOWER is the lowest oversampling mode of the pressure measurement.
-	ULTRALOWPOWER OversamplingMode = iota
-	// BSTANDARD is the standard oversampling mode of the pressure measurement.
-	STANDARD
-	// HIGHRESOLUTION is a high oversampling mode of the pressure measurement.
-	HIGHRESOLUTION
-	// ULTRAHIGHRESOLUTION is the highest oversampling mode of the pressure measurement.
-	ULTRAHIGHRESOLUTION
-)
-
 // BMP180OversamplingMode is the oversampling ratio of the pressure measurement.
 type OversamplingMode uint
 
@@ -88,19 +77,19 @@ func (d *Device) Configure() {
 	d.calibrationCoefficients.md = readInt(data[20], data[21])
 }
 
-// Temperature returns the temperature in celsius degrees (ºC)
-func (d *Device) Temperature() (temperature float32, err error) {
+// Temperature returns the temperature in celsius milli degrees (ºC/1000)
+func (d *Device) Temperature() (temperature int32, err error) {
 	rawTemp, err := d.rawTemp()
 	if err != nil {
 		return
 	}
 	b5 := d.calculateB5(rawTemp)
 	t := (b5 + 8) >> 4
-	return float32(t) / 10, nil
+	return 100*t, nil
 }
 
-// Pressure returns the pressure in pascals (Pa)
-func (d *Device) Pressure() (pressure float32, err error) {
+// Pressure returns the pressure in milli pascals (mPa)
+func (d *Device) Pressure() (pressure int32, err error) {
 	rawTemp, err := d.rawTemp()
 	if err != nil {
 		return
@@ -129,7 +118,7 @@ func (d *Device) Pressure() (pressure float32, err error) {
 	x1 = (p >> 8) * (p >> 8)
 	x1 = (x1 * 3038) >> 16
 	x2 = (-7357 * p) >> 16
-	return float32(p + ((x1 + x2 + 3791) >> 4)), nil
+	return int32(1000*float32(p + ((x1 + x2 + 3791) >> 4))), nil
 }
 
 // rawTemp returns the sensor's raw values of the temperature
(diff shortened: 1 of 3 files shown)


bmp180/bmp180.go
```diff
@@ -0,0 +1,191 @@
...
+//
+package bmp180
+
+import (
+	"time"
+
+	"github.com/aykevl/tinygo/src/machine"
+)
+
+const (
```
deadprogram: Should these be in the other file with the other `const` values?

bmp180/bmp180.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return
+	}
+	b5 := d.calculateB5(rawTemp)
+	t := (b5 + 8) >> 4
+	return float32(t) / 10, nil
+}
+
+// Pressure returns the pressure in pascals (Pa)
+func (d *Device) Pressure() (pressure float32, err error) {
```
deadprogram: I know this is an ugly calculation, perhaps a link back to the datasheet showing why that is the case would be helpful for anyone trying to interpret these "magic numbers".

What do you think?

examples/bmp180/main.go
```diff
@@ -0,0 +1,31 @@
+package main
+
+import (
+	"time"
+
+	"github.com/aykevl/tinygo-drivers/bmp180"
```
deadprogram: Please update to point to new driver repo location.

bmp180/bmp180.go
conejoninja: The datasheet (already linked on top) did not have deeplinks, I had no way to link directly to the algo/calculations. But you are right, I should add a note stating where those numbers come from at least.

The official implementation (recommended on the datasheet) is ... hard to look at: https://github.com/BoschSensortec/BMP180_driver/blob/master/bmp180.c#L252

bmp180/bmp180.go
```diff
@@ -0,0 +1,191 @@
...
+	bus                     machine.I2C
+	mode                    OversamplingMode
+	calibrationCoefficients calibrationCoefficients
+}
+
+// New creates a new BMP180 connection. The I2C bus must already be
+// configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func New(bus machine.I2C) Device {
```
jkaflik: AFAIK it's good practise to include type name in "constructor"

jkaflik:
I don't know is it time for this, but it might be worth to expose common devices interfaces, like we have there:
```golang
type Thermometer interface {
  Temperature() (temperature float32, err error)
}

type Barometer interface {
  Pressure() (pressure float32, err error)
}
```

bmp180/bmp180.go
aykevl: Not always, consider `errors.New()`. The main thing about naming things in Go is not repeating anything, and in this case every package is one device. You don't have to repeat what it is. But of course, that's subjective :)
In any case, I care most about consistency and all other drivers use a bare `New` function.

jkaflik: Yes, do not repeat especially the package name. That's subjective of course and it's what I can live with 👍

jkaflik:
@aykevl 👍 for integers here

deadprogram:
Regarding the common interface thing, seems we're still discussing that, so once we decide on a change, we can update all the existing drivers at that time.

Regarding floats and embedded, that is very true. It is only slightly more confusing for devs, and it for sure more precise and efficient for calculation.

This would probably be a good time to change this, as it probably requires less discussion and consensus. I'm sure that @conejoninja can change the function signatures to return an int32 with the milli-unit of the thing being measured (millidegrees Celsius, millibars of pressure are already standard, so that one is easy.

Should we do this now? If so, let's make a quick decision and get it done. :)

conejoninja:
No problem changing it to int32, it makes sense for precision.

The SI unit for temperature is Kelvin, but Celsius is widely used

bmp180/bmp180.go
```diff
@@ -0,0 +1,180 @@
...
+	var p int32
+	if b7 < 0x80000000 {
+		p = int32((b7 << 1) / b4)
+	} else {
+		p = int32((b7 / b4) << 1)
+	}
+	x1 = (p >> 8) * (p >> 8)
+	x1 = (x1 * 3038) >> 16
+	x2 = (-7357 * p) >> 16
+	return int32(1000*float32(p + ((x1 + x2 + 3791) >> 4))), nil
```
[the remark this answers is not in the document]
conejoninja: Yes, will change that. I was thinking to return Pa isntead of mPa. The common unit used is hPa, and the sensor measures in Pa, which gives us 2 (instead of 3) digit precision. I think it's enough.

```go
  return 1000*(p + ((x1 + x2 + 3791) >> 4)), nil
```

or 



```go
  return p + ((x1 + x2 + 3791) >> 4), nil
```

I'm ok with both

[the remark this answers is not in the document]
aykevl: I would go with mPa. This particular sensor may be measuring with Pa precision, but there are probably other sensors that measure in a higher precision. By returning mPa, those other sensors can return measurements with a higher precision using the same interface.

Of course, you may change the example to show Pa instead of mPa, that's totally up to you.

deadprogram:
Thanks for making all the requested changes, @conejoninja now merging.

PR #12, opened 2019-01-26
Add esp8266 wifi interface supporting UDP

deadprogram:
This PR adds support for an attached separate ESP8266 chip for WiFi communication.

In other words:
Your computer <--> UART0 <--> MCU <--> UART1 <--> ESP8266 <--> INTERNET
or 
MCU <--> UART0 <--> ESP8266 <--> INTERNET
depending on your needs.

There are three examples included in this PR:

- a simple console app that lets you send AT commands to the ESP8266 from a separate microcontroller that is connected to the ESP8266.
- a "station" app that sends data to a UDP "server"
- a "hub" app that receives data from a connect UDP client, and sends responses.

This driver supports compatibility with the `net` Go standard package, but cannot use `net` directly yet. Once TinyGo supports reflection, the duplicated interface is likely no longer be needed in order to compile the code.


esp8266/tcp.go
```diff
@@ -0,0 +1,97 @@
...
+	TCPTransferModeUnvarnished = 1
+)
+
+// ConnectTCPSocket creates a new TCP socket connection for the ESP8266.
+// Currently only supports single connection mode.
+func (d *Device) ConnectTCPSocket(addr, port string) error {
+	protocol := "TCP"
+	val := "\"" + protocol + "\",\"" + addr + "\"," + port
+	d.Set(TCPConnect, val)
+	time.Sleep(100 * time.Millisecond)
```
jkaflik: Is it walkaround to make sure there is a data waiting in buffer?

deadprogram: There is a short delay from when the AT command is sent to the ESP8266 and when it responds/is ready for the next command or data.

I picked this value just at random based on experimenting at 115200 baud serial rate. Probably somewhere in the datasheet is a more specific way to calculate this.

jkaflik: That’s why asked, because didn’t found it at all in specs. Will dig into it

deadprogram:
Any feedback from you @aykevl  ?
I am eager to make any changes or merge this soon.

esp8266/esp8266.go
```diff
@@ -0,0 +1,245 @@
...
+				d.response[i] = header[0]
+				i++
+				d.response[i] = header[1]
+				i++
+			}
+
+			// read the rest of normal command response
+			for d.bus.Buffered() > 0 {
+				data, _ := d.bus.ReadByte()
+				d.response[i] = data
```
aykevl: This doesn't seem to handle errors. Perhaps it would be useful to detect when an overflow occurs?

esp8266/net.go
```diff
@@ -0,0 +1,143 @@
...
+// I/O, not just the immediately following call to Read or
+// Write. After a deadline has been exceeded, the connection
+// can be refreshed by setting a deadline in the future.
+//
+// An idle timeout can be implemented by repeatedly extending
+// the deadline after successful Read or Write calls.
+//
+// A zero value for t means I/O operations will not time out.
+func (c *SerialConn) SetDeadline(t time.Time) error {
+	return nil
```
aykevl: I wonder whether it would be better to either not implement these methods or always return an error? That makes clear these functions are not (yet) implemented.

esp8266/net.go
```diff
@@ -0,0 +1,143 @@
...
+// UDPAddr here to serve as compatible type. until TinyGo can compile the net package.
+type UDPAddr struct {
+	IP   IP
+	Port int
+	Zone string // IPv6 scoped addressing zone; added in Go 1.1
+}
+
+// ParseIP parses s as an IP address, returning the result.
+func ParseIP(s string) IP {
+	return IP([]byte(s))
```
aykevl: This looks odd, but I assume the esp8266 interface uses textual IP addresses?

esp8266/wifi.go
```diff
@@ -0,0 +1,154 @@
...
+// DisconnectFromAP disconnects the ESP8266 from the current access point.
+func (d *Device) DisconnectFromAP() error {
+	d.Execute(Disconnect)
+	time.Sleep(1000 * time.Millisecond)
+	d.Response()
+	return nil
+}
+
+// GetClientIP returns the ESP8266's current client IP addess when connected to an Access Point.
+func (d *Device) GetClientIP() []byte {
```
aykevl: `GetClientIP` returns `[]byte` while `SetClientIP` takes a string? That seems inconsistent.

esp8266/wifi.go
```diff
@@ -0,0 +1,154 @@
...
+// GetAPConfig returns the ESP8266 current configuration when acting as an Access Point.
+func (d *Device) GetAccessPointConfig() []byte {
+	d.Query(SoftAPConfigCurrent)
+	return d.Response()
+}
+
+// SetAPConfig sets the ESP8266 current configuration when acting as an Access Point.
+// ch indicates which radiochannel to use. security should be one of the const values
+// such as WifiAPSecurityOpen etc.
+func (d *Device) SetAPConfig(ssid, pwd string, ch, security int) error {
```
aykevl: These function names are inconsistent: `SetAPConfig` vs `GetAccessPointConfig`.

esp8266/wifi.go
```diff
@@ -0,0 +1,154 @@
...
+	d.Set(ConnectAP, val)
+	time.Sleep(500 * time.Millisecond)
+	d.Response()
+	return nil
+}
+
+// Access Point
+
+// GetAPConfig returns the ESP8266 current configuration when acting as an Access Point.
+func (d *Device) GetAccessPointConfig() []byte {
```
aykevl: What format is this in?
It may be best to return the configuration as structured data, but it should at least say what format the data is in.

esp8266/wifi.go
```diff
@@ -0,0 +1,154 @@
...
+	ecnval := strconv.Itoa(security)
+	val := "\"" + ssid + "\",\"" + pwd + "\"," + chval + "," + ecnval
+	d.Set(SoftAPConfigCurrent, val)
+	time.Sleep(1000 * time.Millisecond)
+	d.Response()
+	return nil
+}
+
+// GetAPClients returns the ESP8266's current clients when acting as an Access Point.
+func (d *Device) GetAPClients() []byte {
```
aykevl: What format is the return value in?

esp8266/wifi.go
```diff
@@ -0,0 +1,154 @@
...
+	d.Query(ListConnectedIP)
+	return d.Response()
+}
+
+// GetAPIP returns the ESP8266's current IP addess when configured as an Access Point.
+func (d *Device) GetAPIP() []byte {
```
aykevl: I guess this is a byte string?
Why not return an `IP` struct anyway?

esp8266/esp8266.go
```diff
@@ -0,0 +1,245 @@
...
+	// command responses that come back from the ESP8266
+	response []byte
+
+	// data received from a TCP/UDP connection forwarded by the ESP8266
+	socketdata    []byte
+	socketdataLen int
+}
+
+// New returns a new esp8266-wifi driver. Pass in a fully configured UART bus.
+func New(b machine.UART) Device {
```
aykevl: This struct is relatively big (~8 words), why not return a `*Device`?

examples/esp8266/esphub/main.go
```diff
@@ -0,0 +1,103 @@
...
+	laddr := &esp8266.UDPAddr{Port: 2222}
+	console.Write([]byte("Loading UDP listener...\r\n"))
+	conn, _ := adaptor.ListenUDP("UDP", laddr)
+
+	console.Write([]byte("Waiting for data...\r\n"))
+	data := make([]byte, 50)
+	for {
+		n, _ := conn.Read(data)
+		if n > 0 {
+			println(string(data))
```
aykevl: This prints all data, even the null bytes (which your console probably hides). You can print `string(data[:n])` to avoid that, and to avoid having to zero the slice afterwards.
Also, why `println` here when you use `console.Write` elsewhere?

esp8266/esp8266.go
```diff
@@ -0,0 +1,245 @@
...
+	count := len(b)
+	if len(b) > d.socketdataLen {
+		count = d.socketdataLen
+	}
+
+	for i := 0; i < count; i++ {
+		b[i] = d.socketdata[i]
+	}
+
+	d.socketdataLen = 0
```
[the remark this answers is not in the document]
deadprogram: Actually I answered my own question based on your next comment.

esp8266/esp8266.go
```diff
@@ -0,0 +1,245 @@
...
+// Device wraps UART connection to the ESP8266.
+type Device struct {
+	bus machine.UART
+
+	// command responses that come back from the ESP8266
+	response []byte
+
+	// data received from a TCP/UDP connection forwarded by the ESP8266
+	socketdata    []byte
+	socketdataLen int
```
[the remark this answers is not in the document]
deadprogram: You are so right, I did not think about using that. Making change that better accommodates the proper syntax.

esp8266/net.go
deadprogram: I left them like this, so that it would be easier to test if it was conforming with the real interface more easily. Also so that we could possibly implement them correctly in context in a future iteration.

esp8266/net.go
deadprogram: Yes, that is correct.

esp8266/wifi.go
deadprogram: Fixing.

esp8266/wifi.go
deadprogram: I want to return as structured data but do not have time to work on fully parsing that right now. Seems like at least for the moment best to return all these as `string`.

esp8266/wifi.go
deadprogram: Will just return a `string` for now.

esp8266/wifi.go
deadprogram: I will return a `string` from all these function for now, I think.

esp8266/esp8266.go
deadprogram: All of the other `New()` constructors just return a `struct`. They should probably all return a `*Device`.

Should we leave this one the same and then change them all at once later?

deadprogram:
OK many updates made based on feedback, and pushed.

esp8266/esp8266.go
aykevl: It really depends on the size of the Device type. Many simple drivers just embed a single never-changing pointer and it is more efficient to pass by value. But this one has a pretty large struct that is modified itself, so I think passing by value is wasteful (and may be incorrect in some cases, for example if you pass the `Device` object to a separate function as a parameter and it gets modified there).
All in all, I think it depends on the driver and we shouldn't try to make this uniform.

esp8266/esp8266.go
```diff
@@ -0,0 +1,245 @@
...
+	count := len(b)
+	if len(b) >= len(d.socketdata) {
+		// copy it all, then clear socket data
+		count = len(d.socketdata)
+		copy(b, d.socketdata[:count])
+		d.socketdata = d.socketdata[:0]
+	} else {
+		// copy all we can, then keep the remaining socket data around
+		copy(b, d.socketdata[:count])
+		d.socketdata = d.socketdata[count:]
```
aykevl: Unfortunately, this is a memory leak. The buffer pointer moves forward every time this happens, and `append` just appends to the end.
It may not be perfect, but I think copying the data at the end to the start is the best thing to do here, unless you want to implement a circular buffer.

```go
// copy all we can, then keep the remaining socket data around
copy(b, d.socketdata[:count])
copy(d.socketdata, d.socketdata[count:])
d.socketdata = d.socketdata[:len(d.socketdata)-count]
```

examples/esp8266/esphub/main.go
```diff
@@ -39,6 +39,14 @@ func main() {
 	adaptor = &dev
 	adaptor.Configure()
 
+	readyled := machine.GPIO{machine.LED}
+	readyled.Configure(machine.GPIOConfig{Mode: machine.GPIO_OUTPUT})
+	readyled.High()
+
+	msgled := machine.GPIO{machine.D7}
```
aykevl: Unfortunately, this is device-specific. Was that intentional?

esp8266/esp8266.go
```diff
@@ -0,0 +1,245 @@
...
+func (d *Device) Read(b []byte) (n int, err error) {
+	return d.bus.Read(b)
+}
+
+// how long in milliseconds to pause after sending AT commands
+const pause = 100
+
+// Execute sends an AT command to the ESP8266.
+func (d Device) Execute(cmd string) error {
+	_, err := d.Write([]byte("AT" + cmd + "\r\n"))
```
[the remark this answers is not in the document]
aykevl: Ok, I can see the point.

examples/esp8266/esphub/main.go
deadprogram: I removed the second LED from this example.

esp8266/esp8266.go
deadprogram: Change made, thanks.

deadprogram:
OK the most recent round of feedback had been reflected in the new changes.

deadprogram:
This PR is still awaiting final review/merge.

esp8266/esp8266.go
```diff
@@ -0,0 +1,249 @@
...
+
+// Echo sets the ESP8266 echo setting.
+func (d Device) Echo(set bool) {
+	if set {
+		d.Execute(EchoConfigOn)
+	} else {
+		d.Execute(EchoConfigOff)
+	}
+	// TODO: check for success
+	d.Response()
```
aykevl: There are many places like this, that seem to ignore any potential errors.
Is this intentional or just not yet handled? I can imagine that it may be difficult to handle errors and even harder to recover from them.

deadprogram: The error handling is still to be added in a future iteration. We have some improvements to the parsing and string handing now in TinyGo itself that should make this a bit easier. 

However, this method should probably return a `nil` even if the error handling is incomplete just to  avoid changing the API.

kamilion:
Looks like this would have no problems working on an ESP32; could this possibly be renamed to something like 'esp-libat-wifi' to make it more explicit that it works with either the ESP32 and ESP8266 with espressif's libat UART bridge firmware? (Or alternatives like zimodem/micropython's AT mode)

I just watched the FOSDEM19 video of the demos and tried stuff with a black pill.

https://www.aliexpress.com/item/10pcs-STM32F103C8T6-ARM-STM32-Minimum-System-Development-Board-Module-For-arduino/32794795688.html

Worked fine with `tinygo flash -target=bluepill` and a lilygo T8 for wifi.

https://www.aliexpress.com/store/product/TTGO-T8-V1-7-ESP32-4MB-PSRAM-TF-CARD-3D-ANTENNA-WiFi-bluetooth-ESP32-WROVER-Micropython/2090076_32851423415.

Eager to see direct support for the ESP32 and ESP8266 in tinygo; gimme a reason to leave micropython! :)

deadprogram:
Hi @kamilion thank you for the update. That is very good news on the AT implementation, and even better news that it works. I will certainly take your suggestion to modify the name of the driver somehow to reflect that it has more compatibility than originally expected.

aykevl:
I think the comments I had about error handling can perhaps be ignored for now. It's a very complex driver but I'd like to get this merged. So I'll wait for a rename, if that happens.

deadprogram:
Just added PR https://github.com/tinygo-org/drivers/pull/23 t replace this one.

Now closing, thank you everyone for the feedback.

PR #13, opened 2019-01-27
Preliminary support for 9-axis accelerometer sensor MPU9250

conejoninja:
This adds basic support for MPU9250, the sensor has some more complex features that I plan to add at a later time, but I'd like to resolve some doubts before continuing the work (not sure if this is the correct place for such questions).

* Acceleration is returned in mG (Gravity/1000), should it be m/s² ?
* Rotation is returned in millidegrees/second, should it be rads/second?
* Magnetometer values should be returned in µT, but I'm getting very low values, oddly enough, the horizontal directions seems to be the correct one (in milliDegrees, should be in rads too?)

### DIFF
--- a/examples/mpu9250/main.go
+++ b/examples/mpu9250/main.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"time"
+
+	"github.com/aykevl/tinygo-drivers/mpu9250"
+	"github.com/aykevl/tinygo/src/machine"
+)
+
+func main() {
+	machine.I2C0.Configure(machine.I2CConfig{})
+	sensor := mpu9250.New(machine.I2C0)
+	sensor.Configure()
+
+	connected := sensor.Connected()
+	if !connected {
+		println("MPU9250 not detected")
+		return
+	}
+	println("MPU9250 detected")
+
+	for {
+		sensor.AccelUpdate()
+		sensor.MagUpdate()
+		sensor.GyroUpdate()
+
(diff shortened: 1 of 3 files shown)


aykevl:
How does this relate to the mpu6050 driver that's already there? As far as I know, both chips are very similar (with the mpu9250 only adding a sensor).

conejoninja:
They are very similar, not sure what are the exact differences besides the added magnetometer (AK8963). Maybe there should be a common "mpu-xxxx" driver

deadprogram:
I am just wondering if perhaps instead of adding a whole new driver for the MPU9250, instead add a new driver for the AK8963 (in a `package ak8963`). 

Optionally then we can have a `package mpu9250` declare `type Driver mpu6050.Driver` or something like that.

Anyone have thoughts on that?

conejoninja:
That could work, and will make things easier to maintain. I couldn't find a AK8963-only board to test the driver, only single smd-chip or mpu9250. I'm closing this PR, will open the proper ones when done.

conejoninja:
I think you are required to configure the mpu9250 to allow bypass ( https://github.com/conejoninja/tinygo-drivers/blob/mpu9250/mpu9250/mpu9250.go#L90 maybe it's not needed), but once done, you could access the sensor directly. (In some mode/configuration you could access the AK8963 via MPU6050's EXT_SENS_DATA_XX).

PR #15, opened 2019-02-06
Added driver for BH1750 digital ambient light sensor

conejoninja:


bh1750/bh1750.go
```diff
@@ -0,0 +1,73 @@
...
+	var coef uint32
+	if d.mode == CONTINUOUS_HIGH_RES_MODE || d.mode == ONE_TIME_HIGH_RES_MODE {
+		coef = HIGH_RES
+	} else if d.mode == CONTINUOUS_HIGH_RES_MODE_2 || d.mode == ONE_TIME_HIGH_RES_MODE_2 {
+		coef = HIGH_RES2
+	} else {
+		coef = LOW_RES
+	}
+	// 1.2 = measurement accuracy as per the datasheet
+	return int32(float32(100*coef*lux) / 1.2)
```
aykevl: You can avoid the floating point operation here by doing this calculation `*5/6`, which has the same effect:

    return coef*lux * 5 / 6 * 100

bh1750/bh1750.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// SetMode changes the reading mode for the sensor
+func (d *Device) SetMode(mode byte) {
+	if mode == CONTINUOUS_HIGH_RES_MODE ||
+		mode == CONTINUOUS_HIGH_RES_MODE_2 ||
+		mode == CONTINUOUS_LOW_RES_MODE ||
+		mode == ONE_TIME_HIGH_RES_MODE ||
+		mode == ONE_TIME_HIGH_RES_MODE_2 ||
+		mode == ONE_TIME_LOW_RES_MODE {
```
aykevl: Why check for all these values?
You could add a little type safety by making these constants of type `Mode` (for example). Doesn't prevent you from casting any other values or passing a number directly, but avoids silly mistakes like passing an unrelated constant by accident.

bh1750/bh1750.go
conejoninja: Thanks, I will be more careful next time with floats

conejoninja:
ops, it was suppossed to be a uint8 instead of uint,  will change it later to byte

deadprogram:
Looks good ow, thanks for the updates @conejoninja now merging.

PR #19, opened 2019-02-13
ws2812: add nrf52 support, update timings for m0 and avr

aykevl:
This commit adds support for the nrf52, which is a Cortex-M4 running at 64MHz.
While implementing this, I discovered that the timings for AVR/16MHz and Cortex-M0 at 16MHz weren't quite right. They do work, but were slightly out of spec. So I fixed this as well.


aykevl:
I can split this PR up in multiple pieces if that makes it easier to review.

deadprogram:
Since the change is basically the same, although for 2 targets, it does not seem worth splitting it.

Now merging, thanks @aykevl for more :bulb:

PR #20, opened 2019-02-16
 Basic support for VL53L1X time-of-flight distance sensor 

conejoninja:
Basic support for VL53L1X time-of-flight distance sensor.

This driver was based on the library https://github.com/pololu/vl53l1x-arduino 

There's an official user manual and API implementation a little hard to read, and in some parts I blindly trust Pololu's implementation

### DIFF
--- a/vl53v1x/vl53l1x.go
+++ b/vl53v1x/vl53l1x.go
@@ -315,12 +315,12 @@ func (d *Device) Status() RangeStatus {
 
 // SignalRate returns the peak signal rate in count per second (cps)
 func (d *Device) SignalRate() int32 {
-	return int32(1000000*d.rangingData.signalRateMCPS)
+	return int32(1000000 * d.rangingData.signalRateMCPS)
 }
 
 // AmbientRate returns the ambient rate in count per second (cps)
 func (d *Device) AmbientRate() int32 {
-	return int32(1000000*d.rangingData.ambientRateMCPS)
+	return int32(1000000 * d.rangingData.ambientRateMCPS)
 }
 
 // getRangingData stores in the buffer the ranging data
--- a/vl53v1x/vl53l1x.go
+++ b/vl53v1x/vl53l1x.go
@@ -1,0 +1,541 @@
+// Package vl53l1x provides a driver for the VL53L1X time-of-flight
+// distance sensor
+//
+// Datasheet:
+// https://www.st.com/resource/en/datasheet/vl53l1x.pdf
+// This driver was based on the library https://github.com/pololu/vl53l1x-arduino
+// and ST's VL53L1X API (STSW-IMG007)
+// https://www.st.com/content/st_com/en/products/embedded-software/proximity-sensors-software/stsw-img007.html
+package vl53l1x
+
+import (
+	"machine"
+	"time"
+)
+
+type DistanceMode uint8
+type RangeStatus uint8
+
+type RangingData struct {
+	mm              uint16
+	status          RangeStatus
+	signalRateMCPS  float32 //MCPS : Mega Count Per Second
+	ambientRateMCPS float32
+}
+
+type ResultBuffer struct {
+	status                     uint8
+	streamCount                uint8
+	effectiveSPADCount         uint16
+	ambientRateMCPSSD0         uint16
+	mmCrosstalkSD0             uint16
+	signalRateCrosstalkMCPSSD0 uint16
+}
+
+// Device wraps an I2C connection to a VL53L1X device.
+type Device struct {
+	bus                machine.I2C
+	mode               DistanceMode
+	timeout            uint32
+	fastOscillatorFreq uint16
+	oscillatorOffset   uint16
+	calibrated         bool
+	VHVInit            uint8
+	VHVTimeout         uint8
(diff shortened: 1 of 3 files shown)


vl53v1x/vl53l1x.go
```diff
@@ -0,0 +1,541 @@
...
+	if d.VHVTimeout != 0 {
+		d.writeReg(VHV_CONFIG_TIMEOUT_MACROP_LOOP_BOUND, d.VHVTimeout)
+	}
+
+	// remove phasecal override
+	d.writeReg(PHASECAL_CONFIG_OVERRIDE, 0x00)
+}
+
+// writeReg sends a single byte to the specified register address
+func (d *Device) writeReg(reg uint16, value uint8) {
```
deadprogram: Just wondering if this function and the other similar ones would be better added to the main repo like https://github.com/tinygo-org/tinygo/blob/master/src/machine/i2c.go#L17 and https://github.com/tinygo-org/tinygo/blob/master/src/machine/i2c.go#L30 since they are not really driver specific.

What do you think?

conejoninja: I think it makes sense to have Write16BitRegister/Read16BitRegister, it's the first device/sensor I see that uses 16 bit registers, I'm not sure how common it is.

A bit off-topic, I don't know if there should be functions to convert (u)int16/32 <=> []byte{} (big/little endian) too. *binary* package is not (yet?) supported.

vl53v1x/vl53l1x.go
```diff
@@ -0,0 +1,541 @@
...
+
+import (
+	"machine"
+	"time"
+)
+
+type DistanceMode uint8
+type RangeStatus uint8
+
+type RangingData struct {
```
aykevl: This struct seems to be used internally only, so I don't think it should be exported.

vl53v1x/vl53l1x.go
deadprogram: Actually `binary` is currently working.

vl53v1x/vl53l1x.go
```diff
@@ -0,0 +1,541 @@
...
+	mm              uint16
+	status          RangeStatus
+	signalRateMCPS  float32 //MCPS : Mega Count Per Second
+	ambientRateMCPS float32
+}
+
+type ResultBuffer struct {
```
aykevl: same here

vl53v1x/vl53l1x.go
```diff
@@ -0,0 +1,541 @@
...
+			d.rangingData.status = RangeValid
+		}
+		break
+
+	default:
+		d.rangingData.status = None
+	}
+
+	d.rangingData.signalRateMCPS = float32(d.results.signalRateCrosstalkMCPSSD0) / (1 << 7)
+	d.rangingData.ambientRateMCPS = float32(d.results.ambientRateMCPSSD0) / (1 << 7)
```
aykevl: It looks like `signalRateMCPS` and `ambientRateMCPS` are converted to float here but only used as integers, so I think this int -> float -> int conversion can be removed.

vl53v1x/vl53l1x.go
```diff
@@ -0,0 +1,541 @@
...
+
+	d.SetDistanceMode(d.mode)
+	d.SetMeasurementTimingBudget(50000)
+
+	d.writeReg16Bit(ALGO_PART_TO_PART_RANGE_OFFSET_MM, d.readReg16Bit(MM_CONFIG_OUTER_OFFSET_MM)*4)
+
+	return true
+}
+
+// SetTimetou configures the timeout
```
aykevl: typo

vl53v1x/vl53l1x.go
```diff
@@ -0,0 +1,541 @@
...
+func (d *Device) SetTimeout(timeout uint32) {
+	d.timeout = timeout
+}
+
+// SetDistanceMode sets the mode for calculating the distance.
+// Distance mode vs. max. distance
+// SHORT: 136cm (dark) - 135cm (strong ambient light)
+// MEDIUM: 290cm (dark) - 76cm (strong ambient light)
+// LONG: 360cm (dark) - 73cm (strong ambient light)
+func (d *Device) SetDistanceMode(mode DistanceMode) bool {
```
aykevl: Can you add a description of the return value to the comment here?
In this case, it seems to be that it returns whether the `mode` is a valid DistanceMode.

vl53v1x/vl53l1x.go
conejoninja: oh, good to know, I need to make a few updates

aykevl: I think that would be useful. For now, we can leave them here and in parallel add these new methods to the `I2C` object, updating this code with the next release if necessary.

deadprogram:
Hi @conejoninja it would be great to merge this, any change of getting the requested changes made soon?

conejoninja:
Yes, probably this week

conejoninja:
Changes made as requested. I will remove the write/readReg methods when they are added  to I2C package

deadprogram:
Looks good now @conejoninja thanks for the update. Now merging.

PR #21, opened 2019-02-16
Very simple 4-wire stepper driver

conejoninja:
This is a very simple 4-wire stepper driver. Maybe an experimental *"x"* package makes more sense for the moment as I feel its interface is not yet defined completely.

### DIFF
--- a/examples/easystepper/main.go
+++ b/examples/easystepper/main.go
@@ -8,7 +8,6 @@ import (
 )
 
 func main() {
-	//motor := easystepper.New(8, 9, 10, 11)
 	motor := easystepper.New(machine.P13, machine.P15, machine.P14, machine.P16, 200, 75)
 
 	for {
--- a/examples/easystepper/main.go
+++ b/examples/easystepper/main.go
@@ -9,15 +9,15 @@ import (
 
 func main() {
 	//motor := easystepper.New(8, 9, 10, 11)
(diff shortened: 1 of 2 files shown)


aykevl:
Out of curiosity, what is the unit of 'steps'? Is that something device-specific?

conejoninja:
A stepper motor (aka simply as "stepper") divides a revolution (360º) in steps, usually 200 steps (each one of 1.8º). "Steps" are the unit. In a "200steps motor", if you rotate 100 steps, you have done half a revolution.

As said, most common number of steps is [200](https://www.sparkfun.com/products/9238), but there are motors with [400 steps](https://www.sparkfun.com/products/10846) or even [only 48](https://www.sparkfun.com/products/10551). And you should take into account microstepping or gears factor (if using any).

examples/easystepper/main.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"machine"
+	"time"
+
+	"github.com/tinygo-org/drivers/easystepper"
+)
+
+func main() {
+	//motor := easystepper.New(8, 9, 10, 11)
```
deadprogram: This looks like it was left in by mistake.

conejoninja:
removed that line

aykevl:
Merged in ae53ea7d814b053a4df07921c5cc0687d9e2204c.

PR #24, opened 2019-02-23
examples: add example for mag3110 magnetometer

deadprogram:
This PR adds an example for mag3110 magnetometer. Not very much to it.

### DIFF
--- a/examples/mag3110/main.go
+++ b/examples/mag3110/main.go
@@ -1,0 +1,22 @@
+// Connects to an MAG3110 I2C magnetometer.
+package main
+
+import (
+	"machine"
+	"time"
+
+	"github.com/tinygo-org/drivers/mag3110"
+)
+
+func main() {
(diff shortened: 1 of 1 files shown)


aykevl:
While useful, I think it will be much more useful when you get some usable data out of it. By default, the magnetometer needs to be calibrated (this is an inherent limitation of all magnetometers).

Also, but that's a comment on the driver itself (which I wrote), the units should be much better defined. Currently they simply return the raw values while they should return values in a standard ISO unit. That would make calibration much easier too as you know the range.

deadprogram:
You can add to the example when you update the driver. Right now there is no example at all, which is what this PR tries to address. I intend to submit a basic example for each driver that currently lacks one in various PRs.

deadprogram:
Any further comments, or can this get merged? We can extend the sample once the driver functionality is extended.

aykevl:
Fair enough.

PR #28, opened 2019-03-10
mma8653: return acceleration in a standard unit: micro-G.

aykevl:
This PR improves the mma8653 driver by returning values in a standard unit, and makes this unit configurable (2G, 4G, 8G). I have chose G and not m/s² because that's what is returned by the 2 sensors I looked at (mma8653 and mpu6050).

### DIFF
--- a/examples/mma8653/main.go
+++ b/examples/mma8653/main.go
@@ -12,10 +12,10 @@ func main() {
 	machine.I2C0.Configure(machine.I2CConfig{})
 
 	accel := mma8653.New(machine.I2C0)
-	accel.Configure(mma8653.DataRate200Hz)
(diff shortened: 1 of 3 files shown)


deadprogram:
Looks like this change will also require the a change in the example here:
https://github.com/tinygo-org/drivers/blob/master/examples/mma8653/main.go#L15

aykevl:
> Looks like this change will also require the a change in the example here:
> https://github.com/tinygo-org/drivers/blob/master/examples/mma8653/main.go#L15

That's already part of this PR.

deadprogram:
Somehow missed that first time. Anyhow now merging.

Issue #14, opened 2019-01-29
Need a way to override the default I2C address


deadprogram:
There are several cases where an i2c device can have an alternative i2c address, due to manufacturer settings or variation, or my user jumper settings.

Perhaps this could be changing:

```go
type Device struct {
	bus machine.I2C
        Address uint16
}

func New(bus machine.I2C) Device {
	return Device{bus: bus, Address: Address}
}
```

Which would allow user to do something like this:

```go
sensor := mag3110.New(machine.I2C0)
sensor.Address = 0xC2 // some alternative if needed
sensor.Configure()
```

We have to change the usage of `Address` in the methods to `d.Address` as well.

aykevl:
This sounds like a good idea!
Of course, this depends on device support but as far as I know most devices have some way to adjust the I2C address.

deadprogram:
Also, sometime "clones" use a different address. I recall that being the case with some MPU6050s and BMP180's that I have gotten from Alibaba.

deadprogram:
This is now in the `master` branch. Closing.

PR #30, opened 2019-03-11
i2c: refactor all i2c drivers to use Address property of Driver.

deadprogram:
This PR refactors all current i2c drivers to use Address property of Driver. It allows variations of a driver to override the default address as discussed in #14

### DIFF
--- a/blinkm/blinkm.go
+++ b/blinkm/blinkm.go
@@ -9,7 +9,8 @@ import (
 
 // Device wraps an I2C connection to a BlinkM device.
 type Device struct {
-	bus machine.I2C
+	bus     machine.I2C
+	Address uint16
 }
 
 // New creates a new BlinkM connection. The I2C bus must already be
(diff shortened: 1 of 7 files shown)


deadprogram:
Conflicts resolved, can now be squashed and merged.

deadprogram:
This branch should be ready for review and then squash and merge.

deadprogram:
Please review me then squash/merge.

blinkm/blinkm.go
```diff
@@ -9,46 +9,52 @@ import (
 
 // Device wraps an I2C connection to a BlinkM device.
 type Device struct {
-	bus machine.I2C
+	bus     machine.I2C
+	Address uint16
```
aykevl: Why are you using `uint16` here?
I haven't yet seen an I2C device in the wild that uses 10 bit addressing. This `Device` struct is also driver specific so there is no shared interface to support here. I think it can be a `uint8`. This will also avoid having to replace `ReadRegister` with `Tx`.

(This comment also applies to all other drivers).

deadprogram: The reason I replaced `ReadRegister` is actually because it creates a memory leak on AVR, since the allocated slice is never freed.

aykevl: I have looked into this a bit, but switching to `Tx` does not seem to improve the situation. I can still see a `runtime.alloc` call in the bh1750 driver after this change, for example.

After some investigation, I think it's because the built-in escape analysis of LLVM is too conservative for our purposes:
https://llvm.org/doxygen/CaptureTracking_8cpp_source.html#L332
In safe Go, there is no crazy pointer arithmetic possible like in C. Therefore, a comparison against nil can never be used to let a pointer escape. I'm not sure what the best way is to fix this.

deadprogram: The reason for `Address uint16` is because the i2c interface signature is `Tx(addr uint16, w, r []byte) error`.

deadprogram:
Squashing and merging.

Issue #27, opened 2019-03-09
Display interface discussion


conejoninja:
This is not an issue, but a discussion on how the _Display_ interface, common interface for managing screens, should look. I really like to heard opinions

I've been looking at [Adafruit's GFX library](https://github.com/adafruit/Adafruit-GFX-Library), the only required function for a display driver is _drawPixel(x,y,color)_ , everything else could be done by GFX itself or overridden by the driver. There are two sets of functions _drawXXX_ and _writeXXX_ but I'm not sure what is the difference between them (I guess one writes in the buffer and the other send it to the screen), or what _startWrite_ and _endWrite_ do.  Also, the 3 drivers I saw (2dom/PxMatrix, adafruit/PCD8544, or adafruit/SSD1306) all use the _display_ method (which send the buffer to the screen), but it's not in GFX specification.

_Display_ interface should be as small as possible, and while related, should not be mistaken by a GFX-like library.

This is part of the Adafruit_GFX.h
```C
// This MUST be defined by the subclass:
  virtual void drawPixel(int16_t x, int16_t y, uint16_t color) = 0;

  // TRANSACTION API / CORE DRAW API
  // These MAY be overridden by the subclass to provide device-specific
  // optimized code.  Otherwise 'generic' versions are used.
  virtual void startWrite(void);
  virtual void writePixel(int16_t x, int16_t y, uint16_t color);
  virtual void writeFillRect(int16_t x, int16_t y, int16_t w, int16_t h, uint16_t color);
  virtual void writeFastVLine(int16_t x, int16_t y, int16_t h, uint16_t color);
  virtual void writeFastHLine(int16_t x, int16_t y, int16_t w, uint16_t color);
  virtual void writeLine(int16_t x0, int16_t y0, int16_t x1, int16_t y1, uint16_t color);
  virtual void endWrite(void);

  // CONTROL API
  // These MAY be overridden by the subclass to provide device-specific
  // optimized code.  Otherwise 'generic' versions are used.
  virtual void setRotation(uint8_t r);
  virtual void invertDisplay(boolean i);

  // BASIC DRAW API
  // These MAY be overridden by the subclass to provide device-specific
  // optimized code.  Otherwise 'generic' versions are used.
  virtual void
    // It's good to implement those, even if using transaction API
    drawFastVLine(int16_t x, int16_t y, int16_t h, uint16_t color),
    drawFastHLine(int16_t x, int16_t y, int16_t w, uint16_t color),
[...]

PR #31, opened 2019-03-12
Define common display interface

conejoninja:
definition of Display interface

### DIFF
--- a/display.go
+++ b/display.go
@@ -1,0 +1,21 @@
+package drivers
+
+import "image/color"
+
+type Display interface {
+	// Size returns the current size of the display.
+	Size() (x, y int16)
+
+	// SetPizel modifies the internal buffer.
+	SetPixel(x, y int16, c color.RGBA)
+
+	// Display sends the buffer (if any) to the screen.
+	Display() error
+
+	// DisplayPixel sends a single pixel to the screen.
+	DisplayPixel(x, y int16, c color.RGBA)
+
+	// DisplayRect sends this part of the buffer to the screen for incremental updates
+	// (e.g. button press animations, blinking cursor, etc.).
+	DisplayRect(x1, y1, x2, y2 int16)
+}
--- a/displayer.go
+++ b/displayer.go
@@ -11,11 +11,4 @@ type Displayer interface {
 
 	// Display sends the buffer (if any) to the screen.
 	Display() error
-
-	// DisplayPixel sends a single pixel to the screen.
-	DisplayPixel(x, y int16, c color.RGBA)
-
-	// DisplayRect sends this part of the buffer to the screen for incremental updates
-	// (e.g. button press animations, blinking cursor, etc.).
-	DisplayRect(x1, y1, x2, y2 int16)
 }
--- a/displayer.go
+++ b/displayer.go
@@ -2,7 +2,7 @@ package drivers
 
 import "image/color"
 
-type Display interface {
+type Displayer interface {
 	// Size returns the current size of the display.
 	Size() (x, y int16)
 


aykevl:
To me this looks good, but perhaps it's better to wait with merging this until an actual implementation has been written.

conejoninja:
I was waiting for this to make the implementation :sweat_smile: I'll send the driver first.

In the current implementation I'm working (hub75) it's not possible to implement _DisplayRect()_ (or I'm not sure how to, need to look in more detail). But I consider it useful as some other devices may support it and left it in the interface.

aykevl:
Yeah maybe `DisplayRect` should be left out of the base interface. My idea was that it updates only the relevant part of the screen if possible, but a driver implementation can fall back to doing a full `Display` call if it does not support partial updates. For now, perhaps it's easier/simpler to leave it out. It can always be added in the future if needed.
(You may notice I have a strong tendency towards interfaces that are as simple as they could possibly be :slightly_smiling_face:)

conejoninja:
On hold until https://github.com/tinygo-org/tinygo/issues/270 is fixed, but let's keep the discussion in #27

display.go
```diff
@@ -0,0 +1,21 @@
+package drivers
+
+import "image/color"
+
+type Display interface {
```
deadprogram: I suggest this be named `Displayer`

display.go
```diff
@@ -0,0 +1,21 @@
...
+
+	// Display sends the buffer (if any) to the screen.
+	Display() error
+
+	// DisplayPixel sends a single pixel to the screen.
+	DisplayPixel(x, y int16, c color.RGBA)
+
+	// DisplayRect sends this part of the buffer to the screen for incremental updates
+	// (e.g. button press animations, blinking cursor, etc.).
+	DisplayRect(x1, y1, x2, y2 int16)
```
deadprogram: Was @aykevl suggesting to remove this?

conejoninja: Here https://github.com/tinygo-org/drivers/pull/31#issuecomment-473262223

In some devices it might not be possible to update just a part of the display, I'm thinking about the SSD1306 or the PCD8544, both use a byte to represent a 8 pixel column and it's the minimum you can send to the display, you will end up updating a bigger area (you need to calculate that new area). In other devices, such a function might not exists. I think it might be useful, it's easy to fallback and update the whole screen (or a bigger area) but the results might not be what you expected.

displayer.go
```diff
@@ -0,0 +1,21 @@
...
+	Size() (x, y int16)
+
+	// SetPizel modifies the internal buffer.
+	SetPixel(x, y int16, c color.RGBA)
+
+	// Display sends the buffer (if any) to the screen.
+	Display() error
+
+	// DisplayPixel sends a single pixel to the screen.
+	DisplayPixel(x, y int16, c color.RGBA)
```
deadprogram: This should also return `error`.

displayer.go
```diff
@@ -0,0 +1,21 @@
...
+
+	// DisplayRect sends this part of the buffer to the screen for incremental updates
+	// (e.g. button press animations, blinking cursor, etc.).
+	DisplayRect(x1, y1, x2, y2 int16)
```
deadprogram: This should also return `error`.

displayer.go
```diff
@@ -0,0 +1,21 @@
+package drivers
+
+import "image/color"
+
+type Displayer interface {
+	// Size returns the current size of the display.
+	Size() (x, y int16)
+
+	// SetPizel modifies the internal buffer.
+	SetPixel(x, y int16, c color.RGBA)
```
deadprogram: This should also return `error`.

deadprogram:
OK made another comment, and I think that is my last one, about methods returning `error`. Seems like a more idiomatic and future-proof signature.

displayer.go
aykevl: I would not return an error here: the only possible error is an index out of bounds which is a programming error. In my opinion, things that are clearly a bug in the program should just panic (just like an index out of bounds). If you want to know more, read about "abandonment" in http://joeduffyblog.com/2016/02/07/the-error-model/.

deadprogram:
Hard to argue against the benefits of less. Let's do that.

deadprogram:
OK then. Time to merge.

PR #32, opened 2019-03-16
Initial support for RGB led matrix *hub75*

conejoninja:
The driver is far from perfect and there are some improvements to make, but it's usable at its current status (you will not be able to display a movie, but could display some slow animations)


hub75/hub75.go
```diff
@@ -0,0 +1,292 @@
...
+	d.oe.High()
+
+	var i uint32
+	for i = 0; i < uint32(d.height); i++ {
+		d.rowOffset[i] = (i%uint32(d.rowPattern))*uint32(d.sendBufferSize) + uint32(d.sendBufferSize) - 1
+	}
+}
+
+// SetPixelColor565 modifies the internal buffer in a single pixel.
+func (d *Device) SetPixelColor565(x int16, y int16, color uint16) {
```
deadprogram: Should the signature for this be `func (d *Device) SetPixelColor565(pt image.Point, color uint16)`?

hub75/hub75.go
```diff
@@ -0,0 +1,292 @@
...
+	r := uint8(((((color >> 11) & 0x1F) * 527) + 23) >> 6)
+	g := uint8(((((color >> 5) & 0x3F) * 259) + 33) >> 6)
+	b := uint8((((color & 0x1F) * 527) + 23) >> 6)
+	d.fillMatrixBuffer(x, y, r, g, b)
+}
+
+// SetPixel modifies the internal buffer in a single pixel.
+func (d *Device) SetPixel(x int16, y int16, c color.RGBA) {
```
deadprogram: Should the signature for this be `func (d *Device) SetPixel(pt image.Point, c color.RGBA)`?

hub75/hub75.go
```diff
@@ -0,0 +1,292 @@
...
+	d.fillMatrixBuffer(x, y, c.R, c.G, c.B)
+}
+
+// DisplayPixel sends a single pixel on the buffer and send it to the screen.
+func (d *Device) DisplayPixel(x int16, y int16, c color.RGBA) {
```
deadprogram: Should the signature for this be `func (d *Device) DisplayPixel(pt image.Point, c color.RGBA)`?

hub75/hub75.go
```diff
@@ -0,0 +1,292 @@
...
+	bufferSize := (d.width * d.height * 3) / 8
+	for c := uint16(0); c < d.colorDepth; c++ {
+		for j := int16(0); j < bufferSize; j++ {
+			d.buffer[c][j] = 0
+		}
+	}
+}
+
+// Size returns the current size of the display.
+func (d *Device) Size() (w, h int16) {
```
deadprogram: Should this be `func (d *Device) Size() (image.Point)`?

conejoninja:
I think it's common to use x,y coords instead of single Point, at least most (if not all) graphics libraries I had worked with use x,y  (maybe because _Point_ doesn't exist in them).  There are some example (in go) _image_ package : https://golang.org/pkg/image/ (eg. Set(x, y int, c color.Color) ) or https://github.com/fogleman/gg

There are some other examples :
- GD PHP : https://www.php.net/manual/en/function.imagesetpixel.php
- openGL? (not sure) : https://www.khronos.org/registry/OpenGL-Refpages/es3.1/html/glReadPixels.xhtml
- Adafruit GFX : https://github.com/adafruit/Adafruit-GFX-Library/blob/master/Adafruit_GFX.h

When it comes to geometry operations, then I find it more common to use Point

Size() should return a Point to be more similar to _std image_ package, I prefer using x,y instead of point for the rest of the functions, but there's no particular reason behind that.



These might be changed too: https://github.com/tinygo-org/drivers/issues/27#issuecomment-471187971  &  https://github.com/tinygo-org/drivers/pull/31

aykevl:
@deadprogram I think we should discuss the interface to #27 and #31.

hub75/hub75.go
aykevl: @conejoninja is there a reason why you put this function in the driver? I think the color encoding is a property of the image data, not of the driver itself.

conejoninja: To make it compatible with examples from 2dom/pxMatrix , but I will delete it.

deadprogram:
hi @conejoninja seems like there are a bunch of commits in the PR that do not pertain to it from other PRs. It needs some cleanup before it can be merged. Perhaps you would be better of starting a new branch from the current master, and then git cherrypick your commits? Not sure on the specific solution.

PR #33, opened 2019-03-26
 Driver for ADXL345 3-axis digital accelerometer 

conejoninja:


deadprogram:
Hi @conejoninja can you please make this PR more like the current `master` pattern for handle the i2c address? e.g. https://github.com/tinygo-org/drivers/blob/master/bmp180/bmp180.go#L36 

Thanks!

adxl345/adxl345.go
```diff
@@ -0,0 +1,191 @@
...
+	bwRate           bwRate
+	x, y, z          int32
+	rawX, rawY, rawZ int32
+}
+
+// New creates a new BMP180 connection. The I2C bus must already be
+// configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func New(bus machine.I2C, address uint16) Device {
```
deadprogram: Shouldn't this function be `func New(bus machine.I2C) Device` to match the other signatures?

adxl345/adxl345.go
```diff
@@ -0,0 +1,191 @@
...
+			measure: 1,
+		},
+		dataFormat: dataFormat{
+			sensorRange: RANGE_2G,
+		},
+		bwRate: bwRate{
+			lowPower: 1,
+			rate:     RATE_100HZ,
+		},
+		Address: address,
```
deadprogram: This should use the default address here, and then the user can set it to some other value if they so desire.

conejoninja:
Changed. Added it too to the readme.md file

deadprogram:
Thanks for the updates. Now merging.

PR #34, opened 2019-04-06
Basic support for DS3231 RTC device

conejoninja:
**IMPORTANT**: This needs  https://github.com/tinygo-org/tinygo/pull/253 (already merged on dev branch but not master)

Basic support for DS3231 RTC

It's missing some advanced features like alarms, square wave or 32kHz outputs, but never used them and will need more time to properly implement them.

It's common for the DS3231 (and DS1307) to ship in the same board as a AT24C32 (EEPROM) but I plan to add a separated driver for it.

### DIFF
--- a/ds3231/ds3231.go
+++ b/ds3231/ds3231.go
@@ -1,0 +1,164 @@
+// Package ds3231 provides a driver for the DS3231 RTC
+//
+// Datasheet:
+// https://datasheets.maximintegrated.com/en/ds/DS3231.pdf
+package ds3231
+
+import (
+	"machine"
+	"time"
+)
+
+type Mode uint8
+
+// Device wraps an I2C connection to a DS3231 device.
+type Device struct {
+	bus machine.I2C
+}
+
+// New creates a new DS3231 connection. The I2C bus must already be
+// configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func New(bus machine.I2C) Device {
+	return Device{
+		bus: bus,
+	}
+}
+
+// Configure sets up the device for communication
+func (d *Device) Configure() bool {
+	return true
+}
+
+// IsDateTimeValid return true/false is the datetime in the device is valid
+func (d *Device) IsDateTimeValid() bool {
+	data := []byte{0}
+	err := d.bus.ReadRegister(Address, REG_STATUS, data)
+	if err != nil {
+		return false
+	}
+	return (data[0] & (1 << OSF)) == 0x00
+}
+
+// IsRunning returns if the oscillator is running
+func (d *Device) IsRunning() bool {
+	data := []uint8{0}
+	err := d.bus.ReadRegister(Address, REG_CONTROL, data)
+	if err != nil {
+		return false
+	}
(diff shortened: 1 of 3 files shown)


ds3231/registers.go
```diff
@@ -0,0 +1,48 @@
...
+	AlarmFlag_Alarm1    = 0x01
+	AlarmFlag_Alarm2    = 0x02
+	AlarmFlag_AlarmBoth = 0x03
+
+	None Mode = 0
+	BatteryBackup Mode = 1
+	Clock Mode = 2
+	AlarmOne Mode = 3
+	AlarmTwo Mode = 4
+	ModeAlarmBoth Mode = 5
```
aykevl: Did you run `go fmt` over this package?

ds3231/ds3231.go
```diff
@@ -0,0 +1,164 @@
...
+		year += 100
+	}
+	month := time.Month(bcdToInt(monthRaw & 0x7F))
+
+	dt = time.Date(year, month, day, hour, minute, second, 0, time.UTC)
+	return
+}
+
+// GetTemperature returns the temperatur
+func (d *Device) GetTemperature() int32 {
```
aykevl: Can you follow the API I proposed [here](https://github.com/tinygo-org/drivers/wiki/Driver-design---API#simple-sensors) (or propose a different API)?
I don't think any other driver has it implemented yet, however.

ds3231/ds3231.go
```diff
@@ -0,0 +1,164 @@
...
+	err = d.bus.WriteRegister(Address, REG_TIMEDATE, data)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GetDateTime returns the date and time
+func (d *Device) GetDateTime() (dt time.Time, err error) {
```
aykevl: The convention in Go is to not provide `Get*` getters, but instead simply the value name.
However, this is not just a getter. In other drivers I've used `Read*` for that instead, to make it clear that it actually does I/O and thus is not a cheap operation. So I would propose `ReadDateTime` instead (or `ReadTime`, see the other comment).

conejoninja:
Sorry, I got excited to see it working and get onto the next thing, I need to be more careful next time

deadprogram:
Hi @conejoninja can you also please make this PR more like the current master pattern to the i2c address? e.g. https://github.com/tinygo-org/drivers/blob/master/bmp180/bmp180.go#L36

This is same feedback as for ADXL345.

Thank you very much.

ds3231/ds3231.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// New creates a new DS3231 connection. The I2C bus must already be
+// configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func New(bus machine.I2C) Device {
+	return Device{
+		bus: bus,
+		Address: Address,
```
aykevl: Can you re-run `go fmt`?

aykevl:
Merged in f6bdc9734f33a133f12c745830e70b7cf7f07e74. Thank you!

PR #36, opened 2019-04-13
lis3dh: initial implementation

deadprogram:
This PR adds an initial implementation for the lis3dh that can retrieve both raw and scaled accelerometer data.

### DIFF
--- a/lis3dh/lis3dh.go
+++ b/lis3dh/lis3dh.go
@@ -1,0 +1,126 @@
+// Package lis3dh provides a driver for the LIS3DH digital accelerometer.
+//
+// Datasheet: https://www.st.com/resource/en/datasheet/lis3dh.pdf
+package lis3dh
+
+import (
+	"machine"
+)
+
+// Device wraps an I2C connection to a LIS3DH device.
+type Device struct {
+	bus     machine.I2C
+	Address uint16
+	r       Range
+}
+
+// New creates a new LIS3DH connection. The I2C bus must already be configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func New(bus machine.I2C) Device {
+	return Device{bus: bus, Address: Address0}
+}
+
+// Configure sets up the device for communication
+func (d *Device) Configure() {
+	// enable all axes, normal mode
+	d.bus.WriteRegister(uint8(d.Address), REG_CTRL1, []byte{0x07})
+
+	// 400Hz rate
+	d.SetDataRate(DATARATE_400_HZ)
+
+	// High res & BDU enabled
+	d.bus.WriteRegister(uint8(d.Address), REG_CTRL4, []byte{0x88})
+
+	// get current range
+	d.r = d.ReadRange()
+}
+
+// Connected returns whether a LIS3DH has been found.
+// It does a "who am I" request and checks the response.
+func (d *Device) Connected() bool {
+	data := []byte{0}
+	err := d.bus.ReadRegister(uint8(d.Address), WHO_AM_I, data)
+	if err != nil {
+		return false
+	}
+	return data[0] == 0x33
+}
+
+// SetDataRate sets the speed of data collected by the LIS3DH.
+func (d *Device) SetDataRate(rate DataRate) {
+	ctl1 := []byte{0}
+	err := d.bus.ReadRegister(uint8(d.Address), REG_CTRL1, ctl1)
+	if err != nil {
+		println(err.Error())
+	}
+	// mask off bits
+	ctl1[0] &^= 0xf0
+	ctl1[0] |= (byte(rate) << 4)
+	d.bus.WriteRegister(uint8(d.Address), REG_CTRL1, ctl1)
+}
+
+// SetRange sets the G range for LIS3DH.
+func (d *Device) SetRange(r Range) {
+	ctl := []byte{0}
+	err := d.bus.ReadRegister(uint8(d.Address), REG_CTRL4, ctl)
+	if err != nil {
+		println(err.Error())
+	}
+	// mask off bits
+	ctl[0] &^= 0x30
+	ctl[0] |= (byte(r) << 4)
+	d.bus.WriteRegister(uint8(d.Address), REG_CTRL4, ctl)
+
+	// store the new range
+	d.r = r
+}
+
+// ReadRange returns the current G range for LIS3DH.
+func (d *Device) ReadRange() (r Range) {
+	ctl := []byte{0}
+	err := d.bus.ReadRegister(uint8(d.Address), REG_CTRL4, ctl)
+	if err != nil {
+		println(err.Error())
+	}
+	// mask off bits
+	r = Range(ctl[0] >> 4)
+	r &= 0x03
+
+	return r
+}
+
+// ReadAcceleration reads the current acceleration from the device and returns
+// it in µg (micro-gravity). When one of the axes is pointing straight to Earth
+// and the sensor is not moving the returned value will be around 1000000 or
+// -1000000.
+func (d *Device) ReadAcceleration() (int32, int32, int32, error) {
+	x, y, z := d.ReadRawAcceleration()
+	divider := float32(1)
+	switch d.r {
+	case RANGE_16_G:
+		divider = 1365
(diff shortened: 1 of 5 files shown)


lis3dh/lis3dh.go
```diff
@@ -0,0 +1,133 @@
...
+	d.bus.WriteRegister(uint8(d.Address), LIS3DH_REG_CTRL1, []byte{0x07})
+
+	// 400Hz rate
+	d.SetDataRate(LIS3DH_DATARATE_400_HZ)
+
+	// High res & BDU enabled
+	d.bus.WriteRegister(uint8(d.Address), LIS3DH_REG_CTRL4, []byte{0x88})
+
+	// DRDY on INT1
+	d.bus.WriteRegister(uint8(d.Address), LIS3DH_REG_CTRL3, []byte{0x10})
```
aykevl: It seems like there is no support for interrupts yet? Why do you enable them?

lis3dh/lis3dh.go
```diff
@@ -0,0 +1,133 @@
...
+	d.bus.WriteRegister(uint8(d.Address), LIS3DH_REG_TEMPCFG, []byte{0x80})
+}
+
+// Connected returns whether a LIS3DH has been found.
+// It does a "who am I" request and checks the response.
+func (d *Device) Connected() bool {
+	data := []byte{0}
+	err := d.bus.ReadRegister(uint8(d.Address), WHO_AM_I, data)
+	if err != nil {
+		println(err.Error())
```
aykevl: This looks like debugging information?
I would just return false on error, because apparently something is wrong with the connection.

lis3dh/lis3dh.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	// mask off bits
+	r = Range(ctl[0] >> 4)
+	r &= 0x03
+
+	return r
+}
+
+// Update reads and stores the sensor values.
+func (d *Device) Update() {
```
aykevl: I'm personally not a fan of keeping state inside the driver. Why not `ReadAcceleration` that directly reads the acceleration from the chip? Just like https://godoc.org/tinygo.org/x/drivers/mma8653#Device.ReadAcceleration.

lis3dh/lis3dh.go
```diff
@@ -0,0 +1,133 @@
...
+	d.bus.WriteRegister(uint8(d.Address), LIS3DH_REG_OUT_X_L|0x80, nil)
+
+	data := []byte{0, 0, 0, 0, 0, 0}
+	d.bus.Tx(d.Address, nil, data)
```
aykevl: I _think_ you can also merge the WriteRegister call above into this `.Tx` call, something like:
```go
d.bus.Tx(d.Address, []uint8{LIS3DH_REG_OUT_X_L}, data)
```

lis3dh/lis3dh.go
```diff
@@ -0,0 +1,133 @@
...
+
+	buf := bytes.NewReader(data)
+	binary.Read(buf, binary.LittleEndian, &(d.x))
+	binary.Read(buf, binary.LittleEndian, &(d.y))
+	binary.Read(buf, binary.LittleEndian, &(d.z))
```
aykevl: I think it's much more efficient to skip the binary package and do the conversion manually:
```go
	x = int16((uint16(data[0])<<8)|uint16(data[1]))
	y = int16((uint16(data[2])<<8)|uint16(data[3]))
	z = int16((uint16(data[4])<<8)|uint16(data[5]))
```

This works on big endian systems as well, if anyone still uses them.

lis3dh/lis3dh.go
```diff
@@ -0,0 +1,133 @@
...
+		divider = 8190
+	case LIS3DH_RANGE_2_G:
+		divider = 16380
+	}
+
+	return int16(float32(d.x) / divider * 1000), int16(float32(d.y) / divider * 1000), int16(float32(d.z) / divider * 1000)
+}
+
+// RawXYZ returns the raw x, y and z axis from the LIS3DH
+func (d *Device) RawXYZ() (x int16, y int16, z int16) {
```
aykevl: Why `RawXYZ`? To me, `RawAcceleration` seems more appropriate (or if you're going for `ReadAcceleration`, it would be `ReadRawAcceleration`).

lis3dh/registers.go
```diff
@@ -0,0 +1,77 @@
...
+// The I2C addresses which this device listens to.
+const (
+	Address0 = 0x18 // SA0 is low
+	Address1 = 0x19 // SA0 is high
+)
+
+// Registers. Names, addresses and comments copied from the datasheet.
+const (
+	WHO_AM_I             = 0x0F
+	LIS3DH_REG_STATUS1   = 0x07
```
aykevl: Is there any reason you used the `LIS3DH_` prefix? This is the lis3dh package, so the prefix is unnecessary (see "effective Go").

lis3dh/lis3dh.go
deadprogram: Removed

lis3dh/lis3dh.go
deadprogram: Removed.

lis3dh/lis3dh.go
deadprogram: I tried that but did not work. Probably explains why the various code out there on the internets does this.

lis3dh/lis3dh.go
deadprogram: Removed. :heavy_minus_sign:

lis3dh/registers.go
deadprogram: Removed.

deadprogram:
All comments have been addressed.

lis3dh/lis3dh.go
aykevl: Ok, then so be it. It's an implementation detail after all and can easily be changed in the future.

deadprogram:
This PR is ready for another review, please.

deadprogram:
This PR seems ready.

deadprogram:
Also added CI smoke test.

lis3dh/lis3dh.go
```diff
@@ -0,0 +1,124 @@
...
+		println(err.Error())
+	}
+	// mask off bits
+	ctl[0] &^= 0x30
+	ctl[0] |= (byte(r) << 4)
+	d.bus.WriteRegister(uint8(d.Address), REG_CTRL4, ctl)
+}
+
+// GetRange returns the current G range for LIS3DH.
+func (d *Device) GetRange() (r Range) {
```
aykevl: All other methods have the Read* prefix when reading from a sensor.

lis3dh/lis3dh.go
```diff
@@ -0,0 +1,124 @@
...
+		divider = 1365
+	case RANGE_8_G:
+		divider = 4096
+	case RANGE_4_G:
+		divider = 8190
+	case RANGE_2_G:
+		divider = 16380
+	}
+
+	return int16(float32(x) / divider * 1000), int16(float32(y) / divider * 1000), int16(float32(z) / divider * 1000)
```
aykevl: I'm not too happy with the floating point calculations here. While this is internal to the driver (and thus easy to change), it would be great if they can be replaced with integer math. However, this is not a big issue, so if it's hard to fix just leave it.

lis3dh/lis3dh.go
deadprogram: Change made.

lis3dh/lis3dh.go
```diff
@@ -0,0 +1,124 @@
...
+	r = Range(ctl[0] >> 4)
+	r &= 0x03
+
+	return r
+}
+
+// ReadAcceleration returns the adjusted x, y and z axis in milli-Gs.
+func (d *Device) ReadAcceleration() (x int16, y int16, z int16) {
+	x, y, z = d.ReadRawAcceleration()
+	r := d.GetRange()
```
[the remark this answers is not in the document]
deadprogram: Good idea. Changes made.

deadprogram:
OK, corrections made.

Issue #37, opened 2019-04-18
BME280 implementation


Munsio:
Hi, 

I'm currently trying to implement a bme280 driver.
This is the first time working on something like that and i have a few problems which i like to discuss on.

I started the whole process by "copying" the bmp180 code and tinker around with the registers i found at: https://github.com/hybridgroup/gobot/blob/master/drivers/i2c/bmp280_driver.go

And here is the Datasheet:
https://www.embeddedadventures.com/datasheets/BME280.pdf

There are 2 things i found:
1) every time I'm using the `sensor.Connected` function the whole program stops working. Reading out the temperature in an for loop works without problems

2) i copied the configuration set and temperature calculation code from the gobot project but the resulting value is 10° to low.

I will link an WIP Pull Request to this issue

aykevl:
On which chip are you running this?
Most ARM Cortex-M chips can be used with GDB, which allows single stepping through the program or breaking it to find where it hangs.

Munsio:
I use an Arduino Uno - currently i am building the hex file with the docker container cause of my panic during installation. As soon as i can fix that i will try it with a debugger - thanks

aykevl:
I see. Unfortunately, the AVR backend has many bugs and I'm not aware of usable AVR debuggers (without expensive hardware). These bugs may result in taking the wrong side of an if/else, for example. I think you're better off using an ARM based chip (any Cortex-M, like the ones branded "M0" or "Zero"): hopefully the AVR backend will mature to be usable but we're not there yet.

ssorc:
> I started the whole process by "copying" the bmp180 code

There is a BME280 driver in the gobot repository that may be a better reference than the BME180 driver that you have used:

https://github.com/hybridgroup/gobot/blob/master/drivers/i2c/bme280_driver.go

PR #38, opened 2019-04-18
Implementation of the BME280 Sensor

Munsio:
Corresponding Issue for discussions: #37


bme280/bme280.go
```diff
@@ -0,0 +1,121 @@
...
+	d.bus.WriteRegister(uint8(d.Address), CTRL_CONFIG, []byte{0xA8})
+
+}
+
+// Connected returns whether a BME280 has been found.
+// It does a "who am I" request and checks the response.
+func (d *Device) Connected() bool {
+	data := []byte{0}
+	d.bus.ReadRegister(uint8(d.Address), WHO_AM_I, data)
+	time.Sleep(1 * time.Second)
```
deadprogram: Is this delay needed here?

bme280/bme280.go
```diff
@@ -0,0 +1,121 @@
...
+	return data[0] == CHIP_ID
+}
+
+// Temperature returns the temperature in celsius milli degrees (ºC/1000)
+func (d *Device) Temperature() (temperature float32, err error) {
```
deadprogram: I suggest changing the signature to this:

```go
func (d *Device) Temperature() (temperature int32, err error)
```

This lets you match the other `Temperature()` methods on other sensors. There is not a formal interface, but this is the custom we have been trying to follow.

Please let me know if this makes sense.

aykevl: At the moment we have both `Temperature() (temperature int32, err error)` and `ReadTemperature() (int32, error)` in the code base (and `ReadTemperature() (temp int8)` but that should be updated). My preference goes to `ReadTemperature() (int32, error)` because the `Read` prefix makes it clear the operation does I/O (and isn't just a getter). Also see https://github.com/tinygo-org/drivers/wiki/Driver-design---API#simple-sensors

aykevl: I think we should pick a single interface and stick with that. We can bikeshed about which it is, but it should be the same across all drivers (where reasonable).

deadprogram: OK, let's go with `ReadTemperature() (int32, error)` across all drivers. Change made.

bme280/bme280.go
```diff
@@ -0,0 +1,145 @@
...
+func readUint(msb byte, lsb byte) uint16 {
+	return (uint16(msb) << 8) | uint16(lsb)
+}
+
+// readData does a burst read from 0xF7 to 0xF0 according to the datasheet
+// resulting in an slice with 8 bytes 0-2 = pressure / 3-5 = temperature / 6-7 = humidity
+func (d *Device) readData() ([]byte, error) {
+	// time.Sleep(5 * time.Millisecond)
+	data := make([]byte, 8)
+	err := d.bus.ReadRegister(uint8(d.Address), REG_PRESSURE, data)
```
deadprogram: This line is always getting the value of the pressure register, not the temperature or other registers.

aykevl: That's actually not what happens. The _first_ register is the pressure register, but it continues reading there and also reads other registers (see the "burst reading" comment). So I think it's fine.

bme280/bme280.go
```diff
@@ -0,0 +1,134 @@
...
+// readUint converts two bytes to uint16
+func readUint(msb byte, lsb byte) uint16 {
+	temp := (uint16(msb) << 8) | uint16(lsb)
+	return (temp >> 8) | (temp << 8)
+}
+
+// readData does a burst read from 0xF7 to 0xF0 according to the datasheet
+// resulting in an slice with 8 bytes 0-2 = pressure / 3-5 = temperature / 6-7 = humidity
+func (d *Device) readData() ([]byte, error) {
+	data := make([]byte, 8)
```
aykevl: You can avoid a memory allocation by moving this `make` call to the caller. For example:

```go
data = make([]byte, 8)
err := d.readData(data)
```

Or, you could return an `[8]byte` as the data is of fixed size anway, that's not too big:

```go
func (d *Device) readData() ([8]byte, error) {
    var data [8]byte
    err := d.bus.ReadRegister(uint8(d.Address), REG_PRESSURE, data[:])
    // etc.
}

bme280/bme280.go
deadprogram: OK, then. Closing.

deadprogram:
Hi @Munsio anything I can help with to try to get this PR ready to merge? Even if it does not have the full sensor functionality it would be great, and other people could then possible help add those missing features. Thanks!

Munsio:
> Hi @Munsio anything I can help with to try to get this PR ready to merge? Even if it does not have the full sensor functionality it would be great, and other people could then possible help add those missing features. Thanks!

Not really - I traveled a lot in the last 2 weeks, but that's over now and i scheduled a new "hacking" night for tomorrow to complete the implementation ;)

deadprogram:
Seems like something odd going on with this branch? The CI tests are failing in a weird way...

Munsio:
> Seems like something odd going on with this branch? The CI tests are failing in a weird way...

Didn't rebase in a while and i do have currently issues with the humidity calculation but i wanted to push it cause my laptop behaves not as he should.

Munsio:
@deadprogram, @aykevl  - i think you can review the PR now

deadprogram:
Now merging, great work on it getting thru so many reviews/updates. Thank you!

PR #39, opened 2019-04-22
thermistor: add support for thermistors such as the NTC 3950

deadprogram:
This PR add support for thermistors such as the NTC 3950.

### DIFF
--- a/thermistor/thermistor.go
+++ b/thermistor/thermistor.go
@@ -1,0 +1,86 @@
+// Package thermistor is for temperature sensing using a thermistor
+// such as the NTC 3950.
+//
+// Datasheet: https://www.farnell.com/datasheets/33552.pdf
+//
+// This code is an interpretation of Adafruit Thermistor module in Python:
+// https://github.com/adafruit/Adafruit_CircuitPython_Thermistor
+//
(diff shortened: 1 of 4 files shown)


thermistor/thermistor.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+// Configure configures the ADC pin used for the thermistor.
+func (d *Device) Configure() {
+	d.adc.Configure()
+}
+
+// Temperature returns the temperature in celsius milli degrees (ºC/1000)
+func (d *Device) Temperature() (temperature int32, err error) {
```
aykevl: See https://github.com/tinygo-org/drivers/pull/38#discussion_r277324240

deadprogram: OK, let's go with ReadTemperature() (int32, error) across all drivers. Change made.

deadprogram:
This PR should now be ready for review.

deadprogram:
Added smoke test to PR.

aykevl:
The only thing I'm not really happy about is the use of floats. But I guess converting this to integer math with the same amount of precision is difficult, so maybe we should keep it this way.
Feel free to update or merge.

PR #42, opened 2019-04-24
bmp180: change method names/signature for obtaining temperature/pressure

deadprogram:
This PR changes the method names/signature for obtaining temperature/pressure on the BMP180 sensor to be uniform with other similar drivers.


aykevl:
Looks good to me.

PR #43, opened 2019-04-24
ci: perform go fmt test on all files

deadprogram:
I think this performs the test against all project files using CircleCI.


aykevl:
This is a good idea.

PR #44, opened 2019-04-25
Initial support for RGB led matrix, usually called hub75

conejoninja:
In replacement of PR #32 this should be cleaner


deadprogram:
Thank for making so many requested changes. Now merging.

PR #48, opened 2019-05-01
Driver for BBC micro:bit on-board LED matrix

conejoninja:
There's some code in the board definition that should be removed ( https://github.com/tinygo-org/tinygo/pull/319 )

This driver allows to enable 2 or more leds at a time (vs only 1 led or all leds with the current code at microbit_board.go file)


deadprogram:
I think this can be safely merged without removing the other code, correct? We can then figure out the disposition of the main repo code in https://github.com/tinygo-org/tinygo/pull/319

conejoninja:
Do not merge yet.

https://github.com/tinygo-org/tinygo/pull/319 should be closed (not merged) and leave the code as is, it's already pretty minimum. 

 I will modify this to use the functions from microbit_board.go code instead when possible.

deadprogram:
Now that the other PR has landed, I an going to merge this one. Good work, @conejoninja once again!

PR #50, opened 2019-05-06
microphone: PDM MEMS microphone support using I2S interface

deadprogram:
This PR implements support for PDM MEMS microphones using I2S interface, such as the microphone built-in to the Adafruit Circuit Playground Express.

The PR in the TinyGo repo https://github.com/tinygo-org/tinygo/pull/331 must be merged before this PR can be merged, since it requires the I2S implementation..

### DIFF
--- a/examples/microphone/main.go
+++ b/examples/microphone/main.go
@@ -1,0 +1,37 @@
+// Example using the i2s hardware interface on the Adafruit Circuit Playground Express
+// to read data from the onboard MEMS microphone.
+//
+// Uses ideas from the https://github.com/adafruit/Adafruit_CircuitPlayground repo.
+//
+package main
+
+import (
+	"machine"
+
+	"tinygo.org/x/drivers/microphone"
+)
+
+const (
+	defaultSampleRate        = 22000
+	quantizeSteps            = 64
+	msForSPLSample           = 50
+	defaultSampleCountForSPL = (defaultSampleRate / 1000) * msForSPLSample
+)
+
+func main() {
+	machine.I2S0.Configure(machine.I2SConfig{
+		Mode:           machine.I2SModePDM,
+		AudioFrequency: defaultSampleRate * quantizeSteps / 16,
+		ClockSource:    machine.I2SClockSourceExternal,
+		Stereo:         true,
+	})
+
+	mic := microphone.New(machine.I2S0)
+	mic.SampleCountForSPL = defaultSampleCountForSPL
+	mic.Configure()
+
+	for {
+		spl, maxval := mic.GetSoundPressure()
+		println("C", spl, "max", maxval)
+	}
+}
--- a/microphone/microphone.go
+++ b/microphone/microphone.go
@@ -1,0 +1,173 @@
+// Package microphone implements a driver for a PDM microphone.
+// For example, the Adafruit PDM MEMS breakout board (https://www.adafruit.com/product/3492)
+//
+// Datasheet: https://cdn-learn.adafruit.com/assets/assets/000/049/977/original/MP34DT01-M.pdf
+//
+package microphone // import "tinygo.org/x/drivers/microphone"
+
+import (
+	"machine"
+	"math"
+)
+
+const (
+	defaultSampleRate        = 22000
+	quantizeSteps            = 64
+	msForSPLSample           = 50
+	defaultSampleCountForSPL = (defaultSampleRate / 1000) * msForSPLSample
+	defaultGain              = 9.0
+	defaultRefLevel          = 0.00002
+)
+
+// Device wraps an I2S connection to a PDM microphone device.
+type Device struct {
+	bus machine.I2S
+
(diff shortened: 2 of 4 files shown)


deadprogram:
Any implementation feedback for this PR?

examples/microphone/main.go
```diff
@@ -0,0 +1,37 @@
...
+	msForSPLSample           = 50
+	defaultSampleCountForSPL = (defaultSampleRate / 1000) * msForSPLSample
+)
+
+func main() {
+	machine.I2S0.Configure(machine.I2SConfig{
+		Mode:           machine.I2SModePDM,
+		AudioFrequency: defaultSampleRate * quantizeSteps / 16,
+		ClockSource:    machine.I2SClockSourceExternal,
+		Stereo:         true,
```
aykevl: I see this only now after the I2S PR was merged but a channel count might have been more flexible.
On the other hand, AFAIK I2S has really been built for mono/stereo and there are very few DAC/ADCs with more than two channels. I could be wrong, though.

microphone/microphone.go
```diff
@@ -0,0 +1,158 @@
...
+		ReferenceLevel:    defaultRefLevel,
+	}
+}
+
+// Configure the microphone.
+func (d *Device) Configure() {
+	d.data = make([]int32, d.SampleCountForSPL)
+	d.buf = make([]uint32, (quantizeSteps / 16))
+
+	return
```
aykevl: Nit: this return is unnecessary.

microphone/microphone.go
```diff
@@ -0,0 +1,158 @@
...
+	for i := 0; i < len(r); i++ {
+
+		// get the next group of samples
+		machine.I2S0.Read(d.buf)
+
+		// filter
+		sum = applySincFilter(d.buf)
+
+		// adjust to 10 bit value
+		s := int32(sum >> 6)
```
aykevl: Is there a reason to cap at 10bit?
While cheap microphones probably don't have anything but noise beyond that point, it seems better to me to not lose bits unnecessarily.

microphone/microphone.go
```diff
@@ -0,0 +1,158 @@
...
+	for i := 0; i < len(d.data); i++ {
+		d.data[i] -= avg
+	}
+
+	// get max value
+	var maxval int32
+	for i := 0; i < len(d.data); i++ {
+		v := d.data[i]
+		if v < 0 {
+			v *= -1
```
aykevl: Why not just `v = -v`?
The effect should be the same, but `-v` seems easier to read for me.

microphone/microphone.go
```diff
@@ -0,0 +1,158 @@
...
+	94, 63, 39, 21, 9, 2, 0, 0,
+}
+
+// applySincFilter uses the sinc filter to process a single set of sample values.
+func applySincFilter(samples []uint32) (result uint16) {
+	var sample uint16
+	pos := 0
+	for j := 0; j < len(samples); j++ {
+		// takes only the low order 16-bits
+		sample = uint16(samples[j] & 0xffff)
```
aykevl: Technically the `& 0xffff` us unnecessary here but I agree it may be easier to read (to make the capping explicit).

examples/microphone/main.go
deadprogram: A bit of research shows a number of multi-channel chips, but not too many MCUs. I suggest we defer that until a future iteration.

microphone/microphone.go
deadprogram: This might be somewhat specific to the mic in the Circuit Playground Express. Might be good to make that a setting.

microphone/microphone.go
deadprogram: Did not realize you could do that.

microphone/microphone.go
aykevl: Why a setting? No common devices are more efficient at 10bit than at 16bit (and most even support 32bit just as efficiently). Using more bits always avoids all this and is most likely both simpler to implement and to use, apart from being more efficient.
Or am I missing something?

microphone/microphone.go
deadprogram: Done.

microphone/microphone.go
deadprogram: I added a separate method `ReadWithFilter()` to provide this capability, and put in a simpler raw `Read()` implementation instead for the more general cases.

microphone/microphone.go
deadprogram: Done.

microphone/microphone.go
deadprogram: Good to be explicit here, probably.

deadprogram:
This PR is ready for review now.

@conejoninja you might be interested in this, as you have a CPE board. You could make a very cool VU meter with it, I bet... ;)

deadprogram:
Squashed everything here into single commit.

PR #53, opened 2019-05-08
SHT3x driver 

torntrousers:
Start of a driver for Sensirion's SHT3x digital temperature and humidity sensor

### DIFF
--- a/sht3x/sht3x.go
+++ b/sht3x/sht3x.go
@@ -29,28 +29,32 @@ func New(bus machine.I2C) Device {
 	}
 }
 
-// Read returns the temperature in celsius.
-func (d *Device) ReadTemperature() (tempCelsius float32) {
-	tempCelsius, _ = d.Read()
-	return tempCelsius
+// Read returns the temperature in celsius milli degrees (ºC/1000).
+func (d *Device) ReadTemperature() (tempMilliCelsius int32, err error) {
+	tempMilliCelsius, _, err = d.ReadTemperatureHumidity()
+	return tempMilliCelsius, err
 }
 
-// Read returns the relative humidity.
-func (d *Device) ReadHumidity() (relativeHumidity float32) {
-	_, relativeHumidity = d.Read()
-	return relativeHumidity
+// Read returns the relative humidity in thousandth of a percent.
+func (d *Device) ReadHumidity() (relativeHumidity int16, err error) {
+	_, relativeHumidity, err = d.ReadTemperatureHumidity()
+	return relativeHumidity, err
 }
 
-// Read returns both the temperature in celsius and relative humidity.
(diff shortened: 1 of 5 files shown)


torntrousers:
This works now on both Arduino and Microbit. The example only runs as it is on the Microbit. I've updated .circleci/config.yml  but I don't know what its for or how to test it.

conejoninja:
Thanks Anthony, great work!


I think it's ok, and could be merged

NOTE: ReadHumidity returns a int16 of the (0-100)% with 2 decimal precision (ie 54.32%) I think it's enough, but no device has this interface yet

conejoninja:
Thanks Anthony, great work!

I think it's ok, and could be merged

NOTE: ReadHumidity returns a int16 of the (0-100)% with 2 decimal precision (ie 54.32%) I think it's enough, but no device has this interface yet

deadprogram:
@conejoninja what made you say "I think that if in your example (the commented code) you use 17 instead of 2 milliseconds - time.Sleep(17*time.Milliseconds) - it should work in arduino devices" and how did you know this? Please share your secrets.

conejoninja:
Oh yeah, I should have added some reference. I got it from here : https://gophers.slack.com/archives/CDJD3SUP6/p1556388326142500

If I understood it correctly, there's some sort of issue with the time.Sleep because a problem on AVR backend of LLVM,  Ayke proposed a workaround that is to sleep the same amount every time. The driver is already sleeping 17 ms, that's why I suggested the same amount

sht3x/sht3x.go
```diff
@@ -0,0 +1,72 @@
...
+	return tempMilliCelsius, relativeHumidity, err
+}
+
+// rawReadings returns the sensor's raw values of the temperature and humidity
+func (d *Device) rawReadings() (uint16, uint16, error) {
+	d.bus.Tx(d.Address, []byte{MEASUREMENT_COMMAND_MSB, MEASUREMENT_COMMAND_LSB}, nil)
+
+	time.Sleep(17 * time.Millisecond)
+
+	data := make([]byte, 5)
```
deadprogram: It would be a good idea to allocate this only one time instead of every time that the method is called.

One way to do this is to make `data` a member of the `Device` struct, and initialize it in the `New()` function.

deadprogram: Sorry I did not notice this before, but all of the conversation about AVR and limited memory made me think of it. Thanks!

aykevl:
I have just verified that the last commit introduces a heap allocation. The previous `make` was so trivial that the compiler optimized it to a stack allocation, so actually the code was both easier to read and more efficient before this change.

deadprogram:
Thank you very much for all the changes @torntrousers it is appreciated. Now merging.

Issue #18, opened 2019-02-07
Tests


deadprogram:
We need to have a way to run tests against these drivers...

deadprogram:
If we were to run "smoke" tests by building the example for each driver, what would be the best way to do this? Download the latest released pre-built binary version of TinyGo from https://github.com/tinygo-org/tinygo/releases or build a dev version from source, or ?

aykevl:
I think downloading the release is both faster and more reliable.

deadprogram:
I have implemented the build of one example per driver as a smoke test. Build is located here: https://circleci.com/gh/tinygo-org/drivers

I will leave this issue open, as we still need more automated tests, this is just a starting point.

PR #54, opened 2019-05-09
Rough in example testing

directionless:
Not sure if this is the right approach, so this is a draft for now.

The examples don't seem to build on all platforms. So need some way to enumerate those disrepencies.

Probably also needs some kind target selection analogous to https://github.com/tinygo-org/tinygo/blob/master/main_test.go#L46-L94

Fixes: https://github.com/tinygo-org/drivers/issues/18

### DIFF
--- a/examples_test.go
+++ b/examples_test.go
@@ -8,18 +8,28 @@ import (
 	"testing"
 )
 
-const TINYGO = "/usr/local/bin/tinygo" // FIXME
-
 func TestExamples(t *testing.T) {
 	examples, err := filepath.Glob("examples/**/main.go")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


aykevl:
Not very Good like but what if we just moved all tests that are now in .circleci/config.yml to a Makefile? Then it's much easier to just run `make test` locally to quickly run the smoke tests.

directionless:
I think of make as an execution engine. I've moved many of my for loops out of make and into go. I'd so something like this PR, and stick a make target in.

deadprogram:
Closing this, since we have moved in a somewhat different direction regarding testing. Thanks!

PR #55, opened 2019-05-09
GPS driver

torntrousers:
Currently this runs and the example prints out fixes every second:
```
GPS Example
675
$GNGGA,092037.00,5135.20819,N,00006.13578,W,1,06,4.36,0.5,M,45.6,M,,*54
09:20:37, lat=51.586803, long=-0.102263, altitude:=0, satelites=6
675
$GNGGA,092038.00,5135.20784,N,00006.13592,W,1,06,4.36,0.1,M,45.6,M,,*50
09:20:38, lat=51.586797, long=-0.102265, altitude:=0, satelites=6
...
```
and runs for about a minute or so and then hangs or gets a panic OOM


ubloxGPS/ubloxGPS.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (d *Device) Read() (result string, err error) {
+	var available = d.Available()
+	if available < 1 {
+		return "", err
+	}
+	println(available)
+	// data := [1024]byte{}
+	data := make([]byte, available)
```
deadprogram: I'd add the `data` to the `Device` struct to avoid this allocation here.

ubloxGPS/ubloxGPS.go
```diff
@@ -0,0 +1,197 @@
...
+	// data := d.buff[0:available]
+	d.bus.ReadRegister(uint8(d.Address), FF, data)
+	buff := bytes.NewBuffer(data)
```
conejoninja: I think all this could be replaced by string(data)

torntrousers:
This seems to work fine as it is on a Microbit, but fails right away on an Arduino with out of memory. I think it looks like the whole approach of reading the entire buffer of available GPS data into an array of strings isn't going to work on the Arduino so need to do something like stream read the sentences byte by byte ...

deadprogram:
Just was looking at this PR, and was thinking it would be better to name the package and files in all lower case.

torntrousers:
OK. Any other comments as it is so far?

torntrousers:
This seems to work ok and does what I need now. Looking for feedback on the API and impl.

ubloxgps/registers.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+const (
+	FD = 0xfd
+	FE = 0xfe
+	FF = 0xff
+)
+
+const (
+	buffer_size = 32
```
aykevl: The convention in Go is camelCase.

ubloxgps/gpsparser.go
```diff
@@ -0,0 +1,137 @@
...
+	Satelites int16
+}
+
+func Parser(gpsDevice GPSDevice) GPSParser {
+	return GPSParser{
+		gpsDevice: gpsDevice,
+	}
+}
+
+func (parser *GPSParser) NextFix(fix *Fix) {
```
aykevl: Can you add some docs here? What it does and how it returns values.

Also, perhaps the API would be cleaner if you just returned the struct. That's a big struct, but I think there are no performance penalties when you return it directly. It would certainly make the API a bit more straightforward: you're returning it after all.

ubloxgps/gpsparser.go
```diff
@@ -0,0 +1,137 @@
...
+type GPSParser struct {
+	gpsDevice GPSDevice
+}
+
+// fix is a GPS location fix
+type Fix struct {
+	Valid     bool
+	Time      string
+	Latitude  string
+	Longitude string
```
aykevl: Would it be possible to convert these 3 strings to integers? That makes it a bit easier to work with them. Or maybe floats are more appropriate for lat/long, they seem to be used a lot to represent coordinates. Time could be represented as seconds since last midnight (so hours * 3600 + minutes * 60 + seconds).

ubloxgps/gpsparser.go
```diff
@@ -0,0 +1,137 @@
...
+	return GPSParser{
+		gpsDevice: gpsDevice,
+	}
+}
+
+func (parser *GPSParser) NextFix(fix *Fix) {
+	// println("ReadNextFix")
+	var ggaSentence = nextGGA(parser.gpsDevice)
+	var ggaFields = strings.Split(ggaSentence, ",")
+	fix.Valid = true
```
aykevl: `.Valid` has not yet been implemented?

ubloxgps/gpsparser.go
```diff
@@ -0,0 +1,137 @@
...
+	"strconv"
+	"strings"
+)
+
+type GPSParser struct {
+	gpsDevice GPSDevice
+}
+
+// fix is a GPS location fix
+type Fix struct {
```
aykevl: @deadprogram should this struct be included in the drivers package itself? Meaning "tinygo.org/x/drivers". If it is, it is possible to define a common interface because all "fix" structs are of the same type.

deadprogram: Well, I think that it is more likely that we would want to add the UART interface to this package than that we will implement an entirely different driver, just because UBLOX is the main manufacturer out there.

ubloxgps/gpsparser.go
```diff
@@ -0,0 +1,137 @@
...
+	fix.Altitude = findAltitude(ggaFields)
+	fix.Satelites = findSatelites(ggaFields)
+	fix.Longitude = findLongitude(ggaFields)
+	fix.Latitude = findLatitude(ggaFields)
+	fix.Time = findTime(ggaFields)
+}
+
+func nextGGA(gpsDevice GPSDevice) (sentence string) {
+	// println("ReadNextGGA")
+	// $--GGA,,,,,,,,,,,,,,*hh
```
deadprogram: This line

```
// $--GGA,,,,,,,,,,,,,,*hh
```

probably should be part of the comment for the `nextGGA()` function.

ubloxgps/gpsparser.go
torntrousers: Maybe there should be a drivers/gps that implements the basic gps over UART function and then another drivers/ubloxgps that supports both I2C and UART. The only difference btw the I2C and UART impl is the [fillBuffer()](https://github.com/HarringayMakerSpace/drivers/blob/ubloxgps/ubloxgps/ubloxgps.go#L68) and [available()](https://github.com/HarringayMakerSpace/drivers/blob/ubloxgps/ubloxgps/ubloxgps.go#L84). The main reason I haven't done this is I don't have an MCU which Tinygo supports that has multiple UARTs so it was hard to test, guess I could go get a Circuit Express to sort that out. Then another thing I'd like to add to the u-blox impl is some ublox specific functions to send commands to the gps device to do things like disable the COCOM limits, similar to this: https://github.com/HarringayMakerSpace/HAB/blob/master/HABTracker/gps.ino#L84

torntrousers:
How about more like this latest commit? 
So a GPS driver which supports both generic UART GPS devices plus u-blox I2C, and some u-blox specific commands in a separate file.generic

gps/gps.go
```diff
@@ -0,0 +1,103 @@
...
+	buffer   []byte
+	bufIdx   int
+	sentence strings.Builder
+	uart     machine.UART
+	bus      machine.I2C
+	address  uint16
+}
+
+// New creates a new UART GPS connection. The UART must already be configured.
+func New(uart machine.UART) GPSDevice {
```
deadprogram: Should probably name this `NewUART()` just to be completely obvious.

gps/gps.go
```diff
@@ -0,0 +1,103 @@
...
+	return gps.buffer[gps.bufIdx]
+}
+
+func (gps *GPSDevice) fillBuffer() {
+	if (machine.I2C{}) == gps.bus {
+		gps.uartFillBuffer()
+	} else {
+		gps.i2cFillBuffer()
+	}
+	// print("[[[")
```
deadprogram: Please remove commented out lines.

gps/gpsparser.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func Parser(gpsDevice GPSDevice) GPSParser {
+	return GPSParser{
+		gpsDevice: gpsDevice,
+	}
+}
+
+func (parser *GPSParser) NextFix() (fix Fix) {
+	// println("ReadNextFix")
```
deadprogram: Please remove.

gps/gpsparser.go
```diff
@@ -0,0 +1,131 @@
...
+	fix.Altitude = findAltitude(ggaFields)
+	fix.Satellites = findSatellites(ggaFields)
+	fix.Longitude = findLongitude(ggaFields)
+	fix.Latitude = findLatitude(ggaFields)
+	fix.Time = findTime(ggaFields)
+	return fix
+}
+
+func nextGGA(gpsDevice GPSDevice) (sentence string) {
+	// println("ReadNextGGA")
```
deadprogram: Please remove this line.

gps/gpsparser.go
```diff
@@ -0,0 +1,131 @@
...
+	// $--GGA,,,,,,,,,,,,,,*hh
```
deadprogram: Please move this line to comment on the `nextGGA()` function.

gps/gpsparser.go
```diff
@@ -0,0 +1,131 @@
...
+	for {
+		sentence = gpsDevice.ReadNextSentence()
+		// print("---")
```
deadprogram: Please remove.

gps/ublox.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func SetCfgGNSS(gpsDevice GPSDevice) (err error) {
+	err = sendCommand(gpsDevice, cfg_gnss_cmd[:])
+	return err
+}
+
+func sendCommand(gpsDevice GPSDevice, command []byte) (err error) {
+	for i := 0; i < len(command); i++ {
+		// gpsDevice.write(command[i])
```
deadprogram: Is this line supposed to send the command?

deadprogram:
I think this is on the right track for sure!

torntrousers:
Great! It doesn't seem perfect that the I2C cod within gps.go is u-blox specific, but I couldn't find a way to structure it to have the u-blox specific code in a separate .go file and still share the generic functions in gps.go and have the parser code work with either UART or I2C. Any idea on a better approach for that?

gps/gps.go
```diff
@@ -0,0 +1,103 @@
...
+func (gps *GPSDevice) readNextByte() (b byte) {
+	gps.bufIdx += 1
+	if gps.bufIdx >= bufferSize {
+		gps.fillBuffer()
+	}
+	return gps.buffer[gps.bufIdx]
+}
+
+func (gps *GPSDevice) fillBuffer() {
+	if (machine.I2C{}) == gps.bus {
```
deadprogram: Seems like you could just make the much simpler test `if gps.uart != nil` here?

torntrousers: I couldn't get that to work - it gives a compile error: ```cannot convert nil (untyped nil value) to machine.UART```

deadprogram: You should probably be doing this anyhow:

```
type GPSDevice struct {
	buffer   []byte
	bufIdx   int
	sentence strings.Builder
	uart     *machine.UART
	bus      *machine.I2C
	address  uint16
}
```

deadprogram: If you really want to make this nice, do like the `Buser` interface in https://github.com/tinygo-org/drivers/blob/master/ssd1306/ssd1306.go does.

torntrousers: Thanks, thats what I was looking for. Will do that.

ubloxgps/gpsparser.go
aykevl: I think we can do such cleanups later when there really is another device, without breaking user code too badly.

gps/gps.go
aykevl: Sounds like a good solution.
Also note that the zero value of `machine.I2C` does not necessarily mean "this is not an I2C bus". Some device may have just one I2C peripheral and for those, `machine.I2C` is zero sized (and thus always equal to the zero value).

ubloxgps/gpsparser.go
deadprogram: Why not represent the Time as a `Time` (https://golang.org/pkg/time/#Time), and the Latitude/Longitude as `float32`? That seems more like the expected types.

deadprogram:
I think the example `examples/ubloxgps/main-uart.go` should be moved to `examples/gps/uart/main.go` and that `examples/ubloxgps/main.go` should be in `examples/gps/i2c/main.go`.

deadprogram:
Hi @torntrousers I think there are only 2 more things to do before this can be merged:

- move the examples to `examples/gps/i2c/main.go` and `examples/gps/uart/main.go` since the package is named `gps` the folder should also be named the same way.
- add the examples to the `Makefile` smoke tests here: https://github.com/tinygo-org/drivers/blob/master/Makefile#L10-L38 please add them where they would go in alphabetical order.

Sounds good?

torntrousers:
Ok, I've done that

gps/ublox.go
```diff
@@ -0,0 +1,53 @@
...
+	return err
+}
+
+func SetCfgGNSS(gpsDevice GPSDevice) (err error) {
+	err = sendCommand(gpsDevice, cfg_gnss_cmd[:])
+	return err
+}
+
+func sendCommand(gpsDevice GPSDevice, command []byte) (err error) {
+	gpsDevice.WriteBytes(command)
```
deadprogram: This should be `gpsDevice.writeBytes(command)`

deadprogram:
In order to compile the `uart` example, this issue in the main TinyGo https://github.com/tinygo-org/tinygo/issues/351 needs to be addressed.

gps/ublox.go
torntrousers: Shucks. It would be great if I knew how to add some automated tests for this.

deadprogram:
I had to go back to the current release of TinyGo 0.5.0 to test on my little UART GPS. Also, my unit is a slightly newer one, so I had to change the example to:

```go
machine.UART1.Configure(machine.UARTConfig{BaudRate: 115200})
```

But then I was able to test using my ItsyBitsy-M0:

```screen
Terminal ready
13:23:21, lat=43.362911, long=-5.849006, altitude:=226, satellites=4
13:23:21, lat=43.362915, long=-5.849011, altitude:=226, satellites=4
```

Success!

I was able to compile but not test the I2C example, as I do not have that HW. However I know you tested it, so I will now squash/merge this PR. Thank you very much for working on it @torntrousers thru all the requested changes.

PR #56, opened 2019-05-09
Support for Waveshare 2.13" (B & C versions) e-paper display

conejoninja:
I put it inside the _waveshare-epd_ folder because there're several devices and unfortunately it can not be created a common driver (different registers, power up process, different resolution, colors supported...)

*NOTE:* This driver is for both B (red) and C(yellow) versions, the official implementation/docs refer to the name epd2in13*b*, there's a D version (black/white only) that might be compatible with a few tweaks. A version is a bit more different and I don't think it could be included in the driver. I can't decided which name suits better epd2in13b (as the official name for B & C) or epd2in13x


deadprogram:
Any comments from you @aykevl before I squash and merge this PR?

deadprogram:
Merging now, great work once again @conejoninja

Issue #51, opened 2019-05-07
RGBW leds


directionless:
I was playing with rgbw leds, using the existing `ws2812` driver. It turned out to be quite easy, instead of writing the GRB bytes,  I needed to write a 4th W byte. So, what might support for this look like?

I imagine there's some benefit to using the existing `color.RGBA` type. Does adding an optional parameter for the white array make sense?

deadprogram:
I just wonder if in this case, we should use the alpha channel for that? Your typical WS2812 does not support which is why `WriteColors()` currently ignores that data. Seems like the easiest thing to do, probably, would be to add an additional member to the struct type like `HasWhite bool` which would control if 3 or 4 bytes are sent.

What do you think?

aykevl:
Perhaps even better: an enum to describe the color model (or however this should be called):

- GRB (currently hardcoded)
- GRBW (your typical RGBW LED strip)

Some (older) strips do not follow the GRB order and it would be useful to add support for those without breaking the API.

deadprogram:
That's a much better idea.

aykevl:
Still, I wonder how we can best pass the white channel in this API. We could (ab)use the alpha channel for that, but I think that's wrong. Maybe we should define a new color type for that, for RGBW?

Come to think of it, I suspect the APA102 uses the wrong color format. I think it should use NRGBA instead of RGBA because it essentially does the alpha calculation inside the LED.

syoder:
I have a 5x8 Neopixel RGBW panel that I was trying to get working and ran into this issue. And my initial thought was maybe it's not completely crazy to use the alpha channel for white.  But if we want to avoid that, we could keep it simple and do something like this:
```go
type RGBW struct {
	color.RGBA
	W uint8
}

func (d Device) WriteColorsRGBW(buf []RGBW) error {
	for _, color := range buf {
		d.WriteByte(color.G) // green
		d.WriteByte(color.R) // red
		d.WriteByte(color.B) // blue
		d.WriteByte(color.W) // white
	}
	return nil
}
```

That would leave the existing `WriteColors` as-is so that we wouldn't break it.

Thoughts? I'm happy to throw together a PR if there is some consensus on how to support RGBW.

nealey:
Please don't use the alpha channel for white. The alpha channel is meant to represent the alpha channel value. There is other existing code that will blend colors based on the alpha channel value.

[...]

PR #57, opened 2019-05-14
Initial tests around rgbw

directionless:
Some initial testing. This code doesn’t seem to work on my Arduino board. Still debugging.

Due to https://github.com/tinygo-org/tinygo/issues/349 the interface feels cumbersome.

Until we understand how to work with white LEDs, there are a couple of interfaces. It’s all an experiment around https://github.com/tinygo-org/drivers/issues/51

### DIFF
--- a/ws2812/ws2812.go
+++ b/ws2812/ws2812.go
@@ -6,15 +6,60 @@ import (
 	"machine"
 )
 
+type ColorOrder string
+
+const (
+	RGB  ColorOrder = "RGB"
+	GRB             = "GRB"
+	GRBW            = "GRBW"
+)
+
 // Device wraps a pin object for an easy driver interface.
 type Device struct {
-	Pin machine.GPIO
+	Pin   machine.GPIO
+	order ColorOrder
+}
+
+// Opt is a functional parameter option
+type DeviceOpt func(*Device)
+
+func WithOrder(o ColorOrder) DeviceOpt {
+	return func(d *Device) {
+		d.order = o
+	}
+}
+
+type ColorOptions struct {
+	Whites []uint8
+}
+
+type ColorOpt func(*ColorOptions)
+
+func WithWhite(whites []uint8) ColorOpt {
+	return func(co *ColorOptions) {
+		co.Whites = whites
+	}
+}
+
+type ColorRGBWA struct {
+	R, G, B, W, A uint8
 }
 
 // New returns a new WS2812 driver. It does not touch the pin object: you have
 // to configure it as an output pin before calling New.
-func New(pin machine.GPIO) Device {
-	return Device{pin}
+func New(pin machine.GPIO, opts ...DeviceOpt) Device {
+	d := Device{
+		Pin: pin,
+	}
+
+	// set default color order
+	WithOrder(GRB)(&d)
+
+	for _, opt := range opts {
+		opt(&d)
+	}
+
+	return d
 }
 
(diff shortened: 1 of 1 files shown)


aykevl:
I see that this PR heavily relies on function pointers. While that results in some nice high-level code, it is also rather bad for optimizations. In general, compilers cannot reason about them and usually can't optimize them. Therefore, I have avoided them everywhere so far.
(Also, I am not entirely convinced that [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) really improve APIs. To me they look like a stop-gap for the lack of optional parameters. But that's my opinion).

Instead, when I wanted to configure something, I have passed a `Config` struct with sane zero values. For example:

```go
type Config struct {
    Order ColorOrdering
}
```

You can see the pattern in all the peripherals (I2C, SPI, etc.) in the machine package.
Also, I see you have made color ordering be a string. It is much more efficient to make it an integer. And with that, I would put `GRB` first because that's what you want in the vast majority of cases, making it possible to just use the zero value of Config: `ws2812.New(machine.NEOPIXEL, ws2812.Config{})`.

```go
type ColorOrdering uint8

const (
    GRB ColorOrdering = iota
    RGB
    RGBW
)
```

Also, wouldn't it be much simpler to just say "RGBW reinterprets the A channel as W"? That's a bit dirty, but keeps almost the same API:

```go
func (d Device) WriteColors(buf []color.RGBA) (n int, err error) {
	for _, c := range buf {
		switch d.order {
		case GRB:
			d.WriteByte(c.G)
			d.WriteByte(c.R)
			d.WriteByte(c.B)
		case GRBW:
			// reinterpret the alpha channel as the white channel
			d.WriteByte(c.G)
			d.WriteByte(c.R)
			d.WriteByte(c.B)
			d.WriteByte(c.A)
		}
	}
}
```

If that is too dirty, I would propose the addition of a new color slice type `RGBW` instead of reusing `RGBA`.

My goal here is 1) to make the API less complex, and 2) to make the API less heavy for embedded systems.

directionless:
I personally prefer functional arguments to config structs, but I think consistency within the project is important. I will convert to that. 

I forget why I used string and not iota there. It's mostly a spike around what interfaces might look like, no big deal to change it.

> RGBW reinterprets the A channel as W

This feels wrong to me. It precludes using the alpha channel as intended. One thing I'd like to support, is being able to send RGBA values to a RGBW led, and to send RGBW values to an RGB one. Though that might be too ambitious in practice. 

This PR shows two potential interfaces. One defines a new type, ColorsRGBW and creates an interface to consume it. The other adds an option to `WriteColors` to accept an array of white values. 

I'll rework this with some of the above feedback, and we can see how those feel.

aykevl:
Are there any LEDs with both an alpha channel and a white channel?
Not saying that we shouldn't support that, but limiting to one of them keeps each RGB value to 4 bytes which may be convenient in some cases.

Also some background: when I initially wrote this, the color slice was `[]uint32`. With some shifting and masking you can get the color out, and in fact this is a bit more efficient than something like `[]struct{byte, byte, byte, byte}` because you only need one load instead of four. But it isn't very clean and somewhat error prone so we switched to `color.RGBA` (especially with the APA102 support that was added around the same time).

deadprogram:
Closing due to having changed/refined our testing strategy a bit. Thank you for working on ideas @directionless

PR #58, opened 2019-05-15
Support for DS1307

k-brk:
This PR implements support for DS1307. It is cheaper version of DS3231, it does not offer alarms nor temperature measurements. Instead it has 56 bytes battery backed SRAM with unlimited writing and square wave output up to 32768 Hz. Tested under bluepill(stm32f103c8t6).


ds1307/ds1307.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"time"
+
+	"machine"
+)
+
+// Device wraps an I2C connection to a DS1307 device.
+type Device struct {
+	bus        machine.I2C
+	address    uint8
```
deadprogram: We have been making the `Address` member of the struct public, so that if someone runs into a weird device with a non-standard I2C address, they can override it.

ds1307/ds1307.go
```diff
@@ -0,0 +1,145 @@
...
+	addresSRAM uint8
```
deadprogram: This should also probably be public. Also, should probably be spelled `AddressSRAM` as weird as that looks.

deadprogram:
Thank you very much for the PR @k-brk I made a couple of small comments. Please let me know if you have any questions.

ds1307/ds1307.go
```diff
@@ -0,0 +1,144 @@
...
+	month := time.Month(bcdToDec(d.data[5]))
+	year := bcdToDec(d.data[6])
+	year += 2000
+
+	t := time.Date(year, month, day, hour, minute, seconds, 0, time.UTC)
+	return t, nil
+}
+
+// SetSRAMAddress sets SRAM register address. Range (SRAMBeginAddres, SRAMEndAddress)
+func (d *Device) SetSRAMAddress(address uint8) {
```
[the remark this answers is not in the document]
k-brk: Thanks for review. Made suggested changes. Although I have few questions

- Is there any difference  in using `d.bus.Tx` over `d.bus.WriteRegister` in `Write` method in terms of on heap allocation? We still have to make new slice with length of data + 1, insert register and copy data to new slice. Wouldn’t escape analysis put slice on stack it this case either way?
- How can I confirm that particular variable is not allocated on heap?

Additionaly I modified `SetOscillatorRunning` to use `d.bus.Tx`, wanted to allocate one data buffer with length of two, read two bytes then swap, insert address at 0 index and push through `d.bus.Tx` to device, this way we could avoid double alocation(one for read, one for write) but I encountered weird behavior. When we read from device two bytes, all registers related to time except for seconds are set to FF. Any data that was in SRAM is fine. This happens only when we read two bytes. When read buffer has length different from two then everything works fine.

Tested in following way:
1 ) Set time to 2019-12-05 20:34:12:00 time.UTC
2 ) Write to 0x8 two bytes ‘A’, ‘B’
3 ) Seek at position 0x0 and read time registers
4 ) Seek at position 0x8 and read two bytes
5 ) Seek at position 0x0 and read time registers
6 ) Seek at position 0x8 and read two bytes

Dumped values of registers at given step(Hex):
3 ) 12 34 20 05 05 12 19   
4 ) 41 42 
5 ) 12 ff ff ff ff ff ff   
6 ) 41 42 

3 ) Dumped time registers from 0x0 to 0x6. They match first step, everything here is OK(Under hex is bcd value, bcd to dec gives step one).
4 ) Two bytes from SRAM, they are fine.
5 ) Read time registers again and they are gone.
6 ) SRAM, fine again.

I have not found the cause yet. I will try it tommorrow with different DS1307 and other uC.

deadprogram:
Hello @k-brk there have been some changes to this repo related to testing that need the merge conflicts resolved. The smoke tests are now being run from the `Makefile` instead of directly in the circleci config file.

Do you have time to look at this? Also, is the rest of the PR ready for further review now?

Thank you.

k-brk:
Hi @deadprogram Merged to master, resolved conflicts. It is ready for further review, Although the bug with registers when we read two bytes still exists. Probably is in stm32f103 I2C read method implementation, I will check and provide findings later today/tomorrow.

deadprogram:
Thanks, @k-brk I look forward to hearing more. I do not have this device to try it out myself, anyone else have one to see if the issue is with the STM32 implementation.

k-brk:
Quick update. Managed to narrow bug down using logic analyzer to case where we read two bytes and next operation is read 4 or more bytes. Only then we are receiving invalid values. Reading 2 bytes and  then in next operation less than 4 works just fine, same thing for 2 -> 3 -> 8. Below you can find output of logic analyzer. In step 1) few values were written to address 0x00, then at 2) and 3) there is 2 byte read operation, then at 4) and 5) 5 bytes read which fails, NACK appears too early.I took a quick look at I2C implementation but it seems OK, I will check it more in depth tomorrow. Also I will try I2C EEPROM to confirm it is related to I2C implementation. Attached minimal working example where it occurs(To test it without logic analyzer you must print value of `data` and `data2`).
 
<details><summary>Logic analyzer results</summary>

| Operation |  Decoded Protocol Result    |
|-----------|-----------------------------|
| 1         | Setup Write to [0xD0] + ACK |
|           | 0x00 + ACK                  |
|           | 0x01 + ACK                  |
|           | 0x02 + ACK                  |
|           | 0x03 + ACK                  |
|           | 0x04 + ACK                  |
| 2         | Setup Write to [0xD0] + ACK |
|           | 0x00 + ACK                  |
| 3         | Setup Read to [0xD1] + ACK  |
|           | 0x01 + ACK                  |
|           | 0x02 + NAK                  |
| 4         | Setup Write to [0xD0] + ACK |
|           | 0x00 + ACK                  |
| 5         | Setup Read to [0xD1] + ACK  |
|           | 0x01 + NAK                  |
|           | 0xFF + ACK                  |
|           | 0xFF + ACK                  |
|           | 0xFF + ACK                  |
|           | 0xFF + ACK                  |
</details>

<details><summary>Minimal working example</summary>

```go
package main

import (
	"machine"
)

var (
	tx uint8 = machine.UART_TX_PIN
	rx uint8 = machine.UART_RX_PIN
)

func main() {
	z := machine.I2C0
	z.Configure(machine.I2CConfig{Frequency: machine.TWI_FREQ_100KHZ})

	icAddr := uint8(0x68)
	reg := uint8(0x00)
	data := make([]byte, 2)
	data2 := make([]byte, 5)

	z.WriteRegister(icAddr, reg, []byte{1, 2, 3, 4})
	for {
		err := z.ReadRegister(icAddr, reg, data)
		if err != nil {
			println(err)
		}
		err = z.ReadRegister(icAddr, reg, data2)
		if err != nil {
			println(err)
		}
	}
}
```

</details>

deadprogram:
OK, great work on both the driver, and the issue reported with I2C on the STM32.

I will squash/merge this PR now, thank you very much for your contributions!

PR #61, opened 2019-05-21
support for Waveshare EPD 2.13" display (black&white only)

conejoninja:
With small modifications (new look up tables) it _might_ be compatible with displays 1.54" (200x200) and 2.9" (296x128), but I have no way to test it.


deadprogram:
Another display, this is awesome! :tv: 

Thank you @conejoninja now merging.

PR #63, opened 2019-05-24
microbitmatrix: fix for dev branch / volatile package/API

conejoninja:
This commit ( https://github.com/tinygo-org/tinygo/commit/9f8340a97065c4868dae0d434967183c7a8b1bc6 ) broke the current microbit matrix driver and the CI for the rest

### DIFF
--- a/microbitmatrix/microbitmatrix.go
+++ b/microbitmatrix/microbitmatrix.go
@@ -4,7 +4,6 @@
 package microbitmatrix
 
 import (
-	"device/nrf"
 	"image/color"
 	"machine"
 	"time"
@@ -46,6 +45,7 @@ type Config struct {
 }
 
 type Device struct {
+	pin      [12]*machine.GPIO
 	buffer   [3][9]bool
 	rotation uint8
 }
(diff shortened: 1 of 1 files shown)


deadprogram:
The volatile package API now requires that you use `Set()`/`Get()` functions.

However, I think this is what the `PortMaskSet()` and `PortMaskClear()` functions are intended for, instead of directly changing register values. 

It would appear that you will also need a `PortMaskDir()` to handle the needs of this driver:
https://github.com/tinygo-org/tinygo/blob/dev/src/machine/machine_nrf.go#L37-L49

What do you think?

conejoninja:
I'll take a deeper look at the changes, I didn't fully understand them, just saw the regValue => Set change and follow with it :laughing: I see now  I could have used PortMaskSet/Clear. Thanks!

conejoninja:
I'm a bit confused on how to use the PortMaskSet/Clear as they only return the register (nrf.GPIO.OUTSET / nrf.GPIO.OUTCLR ¿?) and the mask. For changing the value I still need to use Set (or SetBits but that will change the register multiple times instead of one) on the register and with the mas returned from portMaskSet/Clear, did I understand it correctly? And to use PortMaskSet/Clear I need to previously configure the 12 pins individually (so I have a GPIO struct from the pin number). I see a lot of extra steps, I'm probably missing something, why not use the GPIO.High()/Low() directly?

deadprogram:
The functions return a `*uint32` so you can just change it directly, ala

```go
portClear, _ := d.Pin.PortMaskClear()
set := 0
for col := 0; col < 9; col++ {
	if d.buffer[row][col] {
		set |= 1 << uint8(machine.LED_COL_1+col)
	}
}
*portClear |= set // changes all the bits at same time
```

That said, if you can just change each pin individually, better to use the GPIO interface, as you mention.

deadprogram:
Thanks for the fix @conejoninja now merging.

aykevl:
Late comment from me: yes using just `High()` and `Low()` is better as it avoids a dependency on the device/nrf package.

The example in https://github.com/tinygo-org/drivers/pull/63#issuecomment-495914625 is actually subtly wrong as it does not use volatile and thus may not do what you expect. `PortMaskSet` and `PortMaskClear` are really intended for low-level bitbanging (in assembly), like in the ws2812 driver. See that driver for an example.

PR #66, opened 2019-05-29
ws2812: better support the nrf52832

aykevl:
This has cost me _hours_ to find. There are many clones of the ws2812
and they have slightly different timing characteristics. Some don't work
so well when you get close to the minimum T1H time: a 1-bit may be
interpreted as a 0-bit.


deadprogram:
:traffic_light:

deadprogram:
BTW I changed the branch for this PR to `dev`. Now merging.

PR #67, opened 2019-05-29
Support for AT24C32/64 2-wire serial EEPROM

conejoninja:
sometimes used in some DS1307/DS3231 modules


deadprogram:
:1234: :1234: 

Now merging, @conejoninja thanks for another useful contribution!

PR #68, opened 2019-06-26
espat: update commands

deadprogram:
This PR adds several important features to the ESP-AT driver:

- ensure compatibility with both ESP8266 and ESP32
- adds support for TLS connections using the ESP-AT command set implementation
- changes the `Reponse()` method to use a timeout value instead of fixed time pauses for better performance and stability
- adds an example program that uses MQTT/MQTTS to publish messages to an MQTT broker.
- adds several more methods that are interface compatible with the `net` package such as `Dial()`.

### DIFF
--- a/espat/README.md
+++ b/espat/README.md
@@ -1,0 +1,69 @@
+# ESP-AT Driver
+
+This package provides a driver to use a separate connected WiFi processor either the ESP8266 or the ESP32 from Espressif. 
+
+The way this driver works is by using the UART interface to communicate with the WiFi chip using the Espressif AT command set.
+
+## ESP-AT Firmware Installation
+
+In order to use this driver, you must have the ESP-AT firmware installed on the ESP8266/ESP32 chip.
+
+### Installing on ESP32
+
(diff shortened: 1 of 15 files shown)


deadprogram:
I've added to the PR to provide an `mqtt` package that uses the packets from the Paho mqtt package's packets, along with a modified version of the `Client` that consumes the espat API. It currently only supports `Connect()` and `Publish()`.

This PR is now ready for review.

deadprogram:
OK after adding yet another refactor, this PR has become a lot of changes, sorry about that.

I am not going to add anything further to this PR for now. Once it lands, I will submit any additional PRs as a more properly atomic set of changes.

deadprogram:
I need this PR for next week's release, please! :cat:

conejoninja:
Thanks for the reminder, this is now on top on my list and I'll review as soon as possible

conejoninja:
I agree with Ayke, It looks good and I assume it does work (will be able test on real hardware by the end of the week, beginning of the next one)

deadprogram:
OK, merging. Thank you both for review, and I look forward to iterating more on this.

PR #69, opened 2019-06-30
added rotation to epd2in13 (waveshare)

conejoninja:
This simplifies rotating the screen without need to care about new coordinates in your program

### DIFF
--- a/waveshare-epd/epd2in13/epd2in13.go
+++ b/waveshare-epd/epd2in13/epd2in13.go
@@ -12,10 +12,10 @@ import (
 )
 
 type Config struct {
-	Width        int16
+	Width        int16 // Width is the display resolution
 	Height       int16
-	Rotation     uint8
-	DisplayWidth int16
+	LogicalWidth int16    // LogicalWidth must be a multiple of 8 and same size or bigger than Width
+	Rotation     Rotation // Rotation is clock-wise
 }
 
 type Device struct {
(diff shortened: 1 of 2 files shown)


waveshare-epd/epd2in13/epd2in13.go
```diff
@@ -274,5 +298,28 @@ func (d *Device) ClearBuffer() {
 
 // Size returns the current size of the display.
 func (d *Device) Size() (w, h int16) {
+	if d.rotation == 1 || d.rotation == 3 {
+		return d.height, d.width
+	}
 	return d.width, d.height
 }
+
+// SetRotation changes the rotation of the device
```
aykevl: I think it would be useful to add what the rotation number means. For example: "pass rotation as a number 0-3 that is interpreted as 0, 90, 180, and 270 degrees respectively".

waveshare-epd/epd2in13/epd2in13.go
```diff
@@ -12,8 +12,10 @@ import (
 )
 
 type Config struct {
-	Width  int16
-	Height int16
+	Width        int16
+	Height       int16
+	Rotation     uint8
+	DisplayWidth int16
```
aykevl: What does DisplayWidth mean? Can you maybe add a comment? It's not obvious what the difference is between Width/Height and DisplayWidth, or what DisplayWidth is for.

deadprogram: That is an excellent question!

conejoninja:
Hope it's better explained now, replaced uint8 by a new type "Rotation" and rename width/displayWidth with better names

deadprogram:
@aykevl any feedback on these changes?

aykevl:
Thank you! Merged in d8675515bc519d5b784a24210e1a70a40c594778.

PR #71, opened 2019-07-06
Error strings should not be capitalized

conejoninja:
According to https://github.com/golang/go/wiki/CodeReviewComments#error-strings

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.


deadprogram:
:abc: merging.

PR #72, opened 2019-07-08
Driver ST7735 for TFT color displays

conejoninja:


st7735/st7735.go
```diff
@@ -0,0 +1,442 @@
...
+// Display does nothing, there's no buffer as it might be too big for some boards
+func (d *Device) Display() error {
+	return nil
+}
+
+// SetPixel sets a pixel in the screen
+func (d *Device) SetPixel(x int16, y int16, c color.RGBA) {
+	if x < 0 || y < 0 ||
+		(((d.rotation == NO_ROTATION || d.rotation == ROTATION_180) && (x >= d.width || y >= d.height)) ||
+			((d.rotation == ROTATION_90 || d.rotation == ROTATION_270) && (x >= d.height || y >= d.width))) {
```
aykevl: Wouldn't it be simpler to just compare against the result of `.Size()`? It seems to do the same thing.

st7735/st7735.go
```diff
@@ -0,0 +1,442 @@
...
+	d.Command(RAMWR)
+}
+
+// FillRectangle fills a rectangle at a given coordinates with a color
+func (d *Device) FillRectangle(x, y, width, height int16, c color.RGBA) error {
+	if x < 0 || y < 0 || width <= 0 || height <= 0 ||
+		(((d.rotation == NO_ROTATION || d.rotation == ROTATION_180) &&
+			(x >= d.width || y >= d.height || (x+width) > d.width || (y+height) > d.height)) ||
+			((d.rotation == ROTATION_90 || d.rotation == ROTATION_270) &&
+				(x >= d.height || y >= d.width || (x+width) > d.height || (y+height) > d.width))) {
```
aykevl: Same here - the code here could be much cleaner by sharing the logic with `.Size()`.

st7735/st7735.go
```diff
@@ -0,0 +1,442 @@
...
+func (d *Device) IsBGR(bgr bool) {
+	d.isBGR = bgr
+}
+
+// RGBATo565 converts a color.RGBA to uint16 used in the display
+func RGBATo565(c color.RGBA) uint16 {
+	r, g, b, _ := c.RGBA()
+	/*r = 65535-r
+	g = 65535-g
+	b = 65535-b*/
```
aykevl: this looks like debug code that was never removed?

st7735/st7735.go
```diff
@@ -0,0 +1,442 @@
...
+			((d.rotation == ROTATION_90 || d.rotation == ROTATION_270) &&
+				(x >= d.height || y >= d.width || (x+width) > d.height || (y+height) > d.width))) {
+		return errors.New("rectangle coordinates outside display area")
+	}
+	d.setWindow(x, y, width, height)
+	c565 := RGBATo565(c)
+	c1 := uint8(c565 >> 8)
+	c2 := uint8(c565)
+
+	data := make([]uint8, d.batchLength*2)
```
aykevl: Because `d.batchLength` is not a constant I think it would be better to allocate this slice in the `Configure` call to avoid unnecessary heap allocations.

The same goes for the `make` call in `FillRectangleWithBuffer` below.

deadprogram:
Seems like this PR is ready, so I will squash/merge now. Thanks for the work on this, @conejoninja

PR #73, opened 2019-07-10
mpu6050: properly scale the outputs of the accel/gyro

aykevl:
Units were unspecified before but were in practice the raw output from
the sensor. They have now been changed to sensible outputs
understandable to humans, like the other accelerometers.


deadprogram:
Looks good, thanks for the PR. Now merging.

PR #76, opened 2019-07-29
added driver for ST7789 TFT color display

conejoninja:


conejoninja:
This driver is similar to https://github.com/tinygo-org/drivers/pull/72  probably suggestions for one could be made for the other

deadprogram:
This PR now has merge conflict @conejoninja can you please resolve?

aykevl:
(treat this as a drive-by comment)
Looking at it briefly, there seems to be a lot of duplicate code. That's unfortunate. But perhaps it's better to leave it right now and refactor when a third driver comes along?

conejoninja:
I asked the same somewhere (maybe on slack), I think that a single driver (st77xx) will add more complexity for the moment. In FillRectangle, st7789 uses int32 while st7735 use only int16, SetRotation is also different, the st7789doesn't have the CS pin so Tx() is different too (at least the device I have),... but I'm open to change it

deadprogram:
Let's merge in, and then refactor to shared code as a separate series of PRs.

aykevl:
If it's hard to unify the two, just keep them separate.

PR #79, opened 2019-08-04
Ledstrip display

conejoninja:
A few notes:
- This PR also add a "ledstriper" interface to control a led strips.
- I'm open to change any name, not good naming things, or add more documentation/explanation
- This is more of a _crazy idea_ and I'm not sure it should be on the official _drivers_ repository

It's based on the idea that you could create a display/screen with a ledstrip, like [this example](https://www.makeuseof.com/tag/weekend-project-build-giant-led-pixel-display/). There are a few possible layouts that make sense (parallel, Z, S). There's a limitation on the screen size (as it uses int16) but I think it's enough, a bigger display will probably need a more powerful board (and a ton of money on ledstrips)


peterhellberg:
The `LedStriper` type should most likely be renamed to something like `LEDStriper` since [**LED**](https://en.wikipedia.org/wiki/Light-emitting_diode) is an initializm.

PR #80, opened 2019-08-10
ws2812: fix "invalid symbol redefinition" error

aykevl:
Named local labels shouldn't be used in LLVM because they might get
duplicated resulting in multiple symbol definitions and a compiler
error. Instead, use numeric labels.

---

This fixes just the case for the samd21, other targets also need fixing.


aykevl:
Ran into this issue again today, so I've updated the PR and tested it on all 4 devices.

deadprogram:
I have never actually run into this issue myself, but seems legit. Merging...

PR #81, opened 2019-08-14
ws2812: Add build tag for Arduino Nano33 IoT

13rac1:
Uses `atsamd21` to cover anything using the same chip. Should I be more specific?

### DIFF
--- a/ws2812/ws2812_m0_48m.go
+++ b/ws2812/ws2812_m0_48m.go
@@ -1,4 +1,4 @@
-// +build circuitplay_express itsybitsy_m0
+// +build circuitplay_express itsybitsy_m0 arduino_nano33
 
 package ws2812
 


aykevl:
Yes, I think it must be more specific. The chip used here can also run at lower frequencies and the build tag here really selects the CPU core (Cortex-M0) and the frequency (48MHz). It is possible there are boards where this is not the case but still use a chip from the same family: atsamd21.

13rac1:
The Arduino Nano33 IoT also can run "up to 48Mhz." Changing the build tag to the more specific `arduino_nano33` won't guarantee 48Mhz.

Seems this build tag process may need to be changed for timing sensitive code like this.

aykevl:
I can't easily find a way to change the frequency, is there one? Many chips can't change it on the fly but depend on the used crystal oscillator for that.

In any case, the changes look good to me.

13rac1:
Everything says "Up to", but it seems as though it's not supported by even the Arduino IDE.

deadprogram:
This PR is at least consistent with the other listed boards.

Merging into `dev` branch.

PR #82, opened 2019-08-18
ws2812: fix timings for the nrf51

aykevl:
The timings needed to be changed slightly for ws2811 chips which are slightly slower.

Before this change, my string of ws2811 LEDs would flicker slightly (LEDs randomly turning off a very short time). With this extra nop, it's super solid.

ws2811


deadprogram:
Thanks for the fix. Now merging.

PR #83, opened 2019-08-23
Initial support for LSM6DS3 IMU

conejoninja:
Initial support for LSM6DS3 IMU (the one in the Arduino Nano33).
It support the pedometer feature, but I didn't fully tested it (my USB cable is not that long), but it works. It also support the temperature readings. Missing FIFO support and SPI, but the one in the nano33 is connected via I2C.

### DIFF
--- a/lsm6ds3/lsm6ds3.go
+++ b/lsm6ds3/lsm6ds3.go
@@ -96,10 +96,6 @@ func (d *Device) Configure(cfg Configuration) {
 
 		// Enable pedometer
 		d.bus.WriteRegister(uint8(d.Address), TAP_CFG, []byte{0x40})
-
-		// Configure step detector interrupt
-		//d.bus.WriteRegister(uint8(d.Address), INT1_CTRL, []byte{0x10}) // FIFO
-
 	} else { // NORMAL USE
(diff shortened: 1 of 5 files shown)


lsm6ds3/lsm6ds3.go
```diff
@@ -0,0 +1,187 @@
...
+			d.bus.WriteRegister(uint8(d.Address), CTRL10_C, []byte{0x3E})
+		} else {
+			d.bus.WriteRegister(uint8(d.Address), CTRL10_C, []byte{0x3C})
+		}
+
+		// Enable pedometer
+		d.bus.WriteRegister(uint8(d.Address), TAP_CFG, []byte{0x40})
+
+		// Configure step detector interrupt
+		//d.bus.WriteRegister(uint8(d.Address), INT1_CTRL, []byte{0x10}) // FIFO
```
deadprogram: Did you mean to leave this line in?

conejoninja: I should have removed it, it'll used when/if FIFO is supported

deadprogram:
Thanks for yet another very useful contribution, @conejoninja now squash/merging.

PR #86, opened 2019-09-08
Support software SPI for APA102 (Itsy Bitsy M0 on-board "Dotstar" LED as example)

bgould:
Added implementation and example to support software-based SPI for APA102, for use with boards like Adafruit Itsy Bitsy M0 for instance.

Note: I didn't notice any software SPI implementations already in TinyGo so I implemented the minimum necessary to get the APA102 working.  Since it is incomplete for general-purpose use I left that implementation unexported, to make it easier to swap out in case it would make sense to have software SPI be provided by the TinyGo standard library instead at some point in the future.

### DIFF
--- a/apa102/softspi.go
+++ b/apa102/softspi.go
@@ -1,0 +1,68 @@
+package apa102
+
+import "github.com/tinygo-org/tinygo/src/machine"
+
+// bbSPI is a dumb bit-bang implementation of SPI protocol that is hardcoded
+// to mode 0 and ignores trying to receive data. Just enough for the APA102.
+// Note: making this unexported for now because it is probable not suitable
+// most purposes other than the APA102 package. It might be desirable to make
+// this more generic and include it in the TinyGo "machine" package instead.
+type bbSPI struct {
+	SCK   machine.Pin
+	MOSI  machine.Pin
+	Delay uint32
+}
+
+// Configure sets up the SCK and MOSI pins as outputs and sets them low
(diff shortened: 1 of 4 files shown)


apa102/softspi.go
```diff
@@ -0,0 +1,68 @@
+package apa102
+
+import "github.com/tinygo-org/tinygo/src/machine"
```
conejoninja: shouldn't this just be "machine" ?

bgould: yeah it should.  I thought I had fixed that up before I committed but I think maybe goimports snuck it back in on me.  Good catch, thanks

deadprogram:
Hi @bgould thanks for the contribution. Agreed that it would be great to add the `SoftwareSPI` feature to the TinyGo `machine` package. Perhaps you would like to submit a PR for that?

In the meantime, I tested this PR with my ItsyBitsy M0 and it did work as expected, so now squash/merging.

Thank you!

deadprogram:
OK I should not have merged into `master` but too late. Oops!

Looks like I will need to add another tag for new release.

conejoninja:
I think it could be reverted, but I didn't have a great experience using that option :/

deadprogram:
Likewise. I will make the needed updates.

deadprogram:
Done: https://github.com/tinygo-org/drivers/releases/tag/v0.6.0

PR #87, opened 2019-09-18
veml6070: add Vishay UV light sensor

cmur2:
Product: https://www.vishay.com/docs/84277/veml6070.pdf

I would like to contribute support for this UV light sensor to use it easily with tinygo, I tested the submitted code on the sensor I own and it works well. I used this sensor before which is why I created https://github.com/cmur2/python-veml6070

Please give me feedback for this PR, should I change something or is there more documentation to add?

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ func main() {
 | [ST7735 TFT color display](https://www.crystalfontz.com/controllers/Sitronix/ST7735R/319/) | SPI |
 | [ST7789 TFT color display](https://cdn-shop.adafruit.com/product-files/3787/3787_tft_QT154H2201__________20190228182902.pdf) | SPI |
 | [Thermistor](https://www.farnell.com/datasheets/33552.pdf) | ADC |
+| [VEML6070 UV light sensor](https://www.vishay.com/docs/84277/veml6070.pdf) | I2C |
 | [VL53L1X time-of-flight distance sensor](https://www.st.com/resource/en/datasheet/vl53l1x.pdf) | I2C |
 | [Waveshare 2.13" e-paper display](https://www.waveshare.com/w/upload/e/e6/2.13inch_e-Paper_Datasheet.pdf) | SPI |
 | [Waveshare 2.13" (B & C) e-paper display](https://www.waveshare.com/w/upload/d/d3/2.13inch-e-paper-b-Specification.pdf) | SPI |
--- a/veml6070/veml6070.go
+++ b/veml6070/veml6070.go
@@ -1,0 +1,140 @@
+// Package veml6070 provides a driver for the VEML6070 digital UV light sensor
+// by Vishay.
+//
+// Datasheet:
+// https://www.vishay.com/docs/84277/veml6070.pdf
+// Application Notes:
+// https://www.vishay.com/docs/84310/designingveml6070.pdf
+//
+package veml6070 // import "tinygo.org/x/drivers/veml6070"
+
+import (
+	"time"
+
+	"machine"
+)
+
+// Device wraps an I2C connection to a VEML6070 device.
+type Device struct {
+	bus         machine.I2C
+	AddressLow  uint16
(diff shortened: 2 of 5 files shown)


veml6070/veml6070.go
```diff
@@ -0,0 +1,138 @@
...
+package veml6070 // import "tinygo.org/x/drivers/veml6070"
+
+import (
+	"time"
+
+	"machine"
+)
+
+// Device wraps an I2C connection to a VEML6070 device.
+type Device struct {
```
conejoninja: We usually allow the I²C address to be configurable (for example https://github.com/tinygo-org/drivers/blob/master/bh1750/bh1750.go#L20 ), but this case is a bit special, as it uses two address for reading. Maybe it's a good idea to allow both to be configurable, or maybe configure one and let the other be +1. What do @aykevl & @deadprogram  think?

veml6070/veml6070.go
```diff
@@ -0,0 +1,138 @@
...
+	return d.bus.Tx(uint16(ADDR_L), []byte{CONFIG_DISABLE}, nil)
+}
+
+func (d *Device) enable() error {
+	return d.bus.Tx(uint16(ADDR_L), []byte{CONFIG_ENABLE | d.IT}, nil)
+}
+
+func (d *Device) readData(address uint16) (byte, error) {
+	data := []byte{0}
+
```
conejoninja: for my taste, there's no need for these empty lines

veml6070/veml6070.go
```diff
@@ -0,0 +1,138 @@
+// Package bmp180 provides a driver for the BMP180 digital pressure sensor
```
conejoninja: References to BMP180 digital pressure sensor

veml6070/veml6070.go
cmur2: I also thought about this but then this sensor only allows those two fixed addresses to be used - should it still be configurable?

veml6070/veml6070.go
cmur2: Uh thanks for spotting my copy&paste error :+1:

veml6070/veml6070.go
cmur2: Removed

veml6070/veml6070.go
deadprogram: I think we need to expose the same interface, but what the driver actually does can be specific.

One reason we have the `Address` able to be changed, is to handle "clone" or "off-market" sensors that use different addresses. Another is the case that someone uses an I2C multiplexer like https://www.adafruit.com/product/2717

Basically, it seems best to allow it to be changed, even if the device itself does not.

cmur2: That's a good point, didn't thinkg about that... I now made both (the lower and the higher) addresses individually configurable so the user has maximum freedom - what do you think @deadprogram ?

deadprogram: The change looks good to me.

README.md
```diff
@@ -82,6 +82,7 @@ func main() {
 | [ST7735 TFT color display](https://www.crystalfontz.com/controllers/Sitronix/ST7735R/319/) | SPI |
 | [ST7789 TFT color display](https://cdn-shop.adafruit.com/product-files/3787/3787_tft_QT154H2201__________20190228182902.pdf) | SPI |
 | [Thermistor](https://www.farnell.com/datasheets/33552.pdf) | ADC |
+| [VEML6070](https://www.vishay.com/docs/84277/veml6070.pdf) | I2C |
```
deadprogram: This should probably say "VEML6070 UV light sensor" not just "VEML6070".

cmur2: You're right, done! Anything else?

cmur2:
Is this PR mergeble now?

deadprogram:
I've switched the branch for the PR to `dev`, and will now merge it. Thank you @cmur2 for the contribution!

PR #93, opened 2019-10-29
mcp3008: add implementation for MCP3008 ADC with SPI interface

deadprogram:
This PR adds implementation for MCP3008 ADC with SPI interface.

### DIFF
--- a/mcp3008/mcp3008.go
+++ b/mcp3008/mcp3008.go
@@ -1,0 +1,92 @@
+// Package mcp3008 implements a driver for the MCP3008 Analog to Digital Converter.
+//
+// Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/21295d.pdf
+//
+package mcp3008 // import "tinygo.org/x/drivers/mcp3008"
+
+import (
+	"errors"
+	"machine"
+)
+
+// Device wraps MCP3008 SPI ADC.
+type Device struct {
+	bus machine.SPI
+	cs  machine.Pin
+	tx  []byte
+	rx  []byte
+	CH0 ADCPin
+	CH1 ADCPin
+	CH2 ADCPin
+	CH3 ADCPin
+	CH4 ADCPin
+	CH5 ADCPin
+	CH6 ADCPin
+	CH7 ADCPin
+}
+
+// ADCPin is the implementation of the ADConverter interface.
+type ADCPin struct {
+	machine.Pin
+	d *Device
+}
+
+// New returns a new MCP3008 driver. Pass in a fully configured SPI bus.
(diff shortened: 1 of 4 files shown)


mcp3008/mcp3008.go
```diff
@@ -0,0 +1,92 @@
...
+	CH3 *ADCPin
+	CH4 *ADCPin
+	CH5 *ADCPin
+	CH6 *ADCPin
+	CH7 *ADCPin
+}
+
+// ADCPin is the implementation of the ADConverter interface.
+type ADCPin struct {
+	machine.ADC
```
aykevl: Why do you need to embed this type? Why not just `machine.Pin`?

mcp3008/mcp3008.go
```diff
@@ -0,0 +1,92 @@
...
+	tx  []byte
+	rx  []byte
+	CH0 *ADCPin
+	CH1 *ADCPin
+	CH2 *ADCPin
+	CH3 *ADCPin
+	CH4 *ADCPin
+	CH5 *ADCPin
+	CH6 *ADCPin
+	CH7 *ADCPin
```
aykevl: Why not use an array or slice here?

Also, you could make these non-pointers, which is somewhat lighter on the heap/GC. For example:

```go
type Device struct {
    // ...
    [8]CH ADCPin
}
```

You can always get the pointer with `&device.CH[0]` for example.

mcp3008/mcp3008.go
deadprogram: A future PR will add an interface type `ADConverter` so we can use the same as a normal `machine.ADC` for other drivers such as the `thermistor`.

mcp3008/mcp3008.go
deadprogram: I was trying to make this act as much as possible like a "regular" `machine.ADC`, hence the specific members. The pointer is so we can in the future use as a param of type `machine.ADConverter`.

mcp3008/mcp3008.go
aykevl: But you wouldn't need to embed this type to implement an interface?

For example, the interface could look like this:

```go
type ADCer interface {
    Get() uint16
}
```

In that case, the `ADCPin` type already implements the interface.
(Also, perhaps `Get() uint16` is a bit too generic if ADC is used as an interface).

deadprogram: Changed to simply `machine.Pin`

mcp3008/mcp3008.go
deadprogram: Got rid of pointers, but left as individual members.

deadprogram:
Changes made and force pushed.

PR #97, opened 2019-11-16
espat: implement MQTT subscribe functionality via channels

deadprogram:
This PR implements the `Subscribe()` method using channels. It also refactors response processing for greater speed and efficiency.

### DIFF
--- a/espat/espat.go
+++ b/espat/espat.go
@@ -19,6 +19,7 @@
 package espat // import "tinygo.org/x/drivers/espat"
 
 import (
+	"errors"
 	"machine"
 	"strconv"
 	"strings"
@@ -54,11 +55,11 @@ func (d *Device) Connected() bool {
 	d.Execute(Test)
 
 	// handle response here, should include "OK"
-	r := d.Response(100)
-	if strings.Contains(string(r), "OK") {
-		return true
+	_, err := d.Response(100)
+	if err != nil {
+		return false
 	}
-	return false
+	return true
 }
 
 // Write raw bytes to the UART.
@@ -72,7 +73,7 @@ func (d *Device) Read(b []byte) (n int, err error) {
 }
 
 // how long in milliseconds to pause after sending AT commands
-const pause = 100
+const pause = 300
 
 // Execute sends an AT command to the ESP8266/ESP32.
 func (d Device) Execute(cmd string) error {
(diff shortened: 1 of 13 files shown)


deadprogram:
Also, I have confirmed that the new MQTT example in this PR that has both publish and subscribe using blocking channels works as expected with the latest `dev` branch using LLVM 9.

conejoninja:
I'm inclined to greenlight this PR, because it's an improvement but "Subscribe" is a delicate piece that might fail.

I found the regular issues with wifi, looks like a crashed nano33 it's not able to re-join a wifi network if the client is still active, but not sure if it's because I used my phone as a hotspot, or if it's something we might be able to fix at driver level (turning off/on the hotspot worked fine).


Besides that, in a non-ssl but with user/password MQTT server, publishing works quite well, no issues detected.

Subscribe doesn't work that well:
- It misses some messages sent to the channel it's listening
- One time it stopped receiving messages at all but kept publishing fine
- Another time I got a "panic: runtime error: slice out of range" (no more information)
- If receiving messages too frequently (maybe 2 in a a couple of seconds?) it may crash resulting in 
```
Publishing MQTT message...
Publishing MQTT message...
[tinygo/rx] TEST
Publishing MQTT message...
StartSocketSend error:
[tinygo/rx] TEST
Publishing MQTT message...
response timeout error:
Publishing MQTT message...
response timeout error:
Publishing MQTT message...
response timeout error:
Recv 45 bytes

SEND OK

Publishing MQTT message...
response error:AT+CIPSEND=45

ERROR

Publishing MQTT message...
response error:AT+CIPSEND=45

ERROR

Publishing MQTT message...
response error:AT+CIPSEND=45

ERROR

..... 
```

This is a great improvement and a big step forward. I'm a bit worried that it could be crashed from _outside_, just by sending a burst of messages to a channel, but I guess this is still a work-in-progress (I see some TODOs in the code)

deadprogram:
We should probably merge this as "experimental" feature and then continue to work on it.

As you mentioned, the other changes of this PR are actually very good changes.

conejoninja:
Agreed on that

deadprogram:
Build is working again, and tests are passing. @conejoninja are you going to rebase/merge this PR?

conejoninja:
merged!

PR #99, opened 2019-11-27
all: add md5sum to smoke tests

aykevl:
This makes it easier to quickly compare all smoke tests for any changes.


deadprogram:
Good idea! Merging.

PR #101, opened 2019-12-04
Dual stepper

conejoninja:
_DualDevice_ allows to move two motors simultaneously , you could move each motor a different number of steps (in the same time, so one will be slower than usual).

Useful for two wheels robots


easystepper/easystepper.go
```diff
@@ -49,6 +79,68 @@ func (d *Device) Move(steps int32) {
...
+func (d *Device) Off() {
+	d.pins[0].Low()
+	d.pins[1].Low()
+	d.pins[2].Low()
+	d.pins[3].Low()
+}
+
+// Move rotates the motors the number of given steps
+// (negative steps will rotate it the opposite direction)
+func (d *DualDevice) Move(stepsA, stepsB int32) {
```
deadprogram: Have not gone thru the details, but seems like this method could be cleaned up further...

conejoninja:
I run out of ideas to clean it up further

deadprogram:
some version of this code is duplicated 3 times:

```go
if directions[max] {
    d.devices[max].stepMotor(uint8(s % 4))
} else {
    d.devices[max].stepMotor(uint8((s + 2*(s%2)) % 4))
}
```

This could be a method;

```go
func (d *Device) moveSteps(direction bool, steps int32) error {
	var s int32
	if direction {
		s = steps % 4
	} else {
		s = (steps + 2*(steps%2)) % 4
	}
	d.stepMotor(uint8(s))
}
```

Which you could then call from https://github.com/tinygo-org/drivers/pull/101/files#diff-1317a793b01cf4ab43b2db73e8d16199R70, https://github.com/tinygo-org/drivers/pull/101/files#diff-1317a793b01cf4ab43b2db73e8d16199R112 and https://github.com/tinygo-org/drivers/pull/101/files#diff-1317a793b01cf4ab43b2db73e8d16199R120

What do you think?

conejoninja:
Thanks, I think it's a little more complicate to understand the _"step"_ variable of the new function with this change, but it's an internal function so users will not be exposed to it.

easystepper/easystepper.go
```diff
@@ -26,31 +31,97 @@ func New(pin1, pin2, pin3, pin4 machine.Pin, steps int32, rpm int32) Device {
 	}
 }
 
+// NewDual returns a new dual easystepper driver given 8 pins numbers (not pin object),
+// number of steps and rpm
+func NewDual(pin1, pin2, pin3, pin4, pin5, pin6, pin7, pin8 machine.Pin, steps int32, rpm int32) DualDevice {
+	pin1.Configure(machine.PinConfig{Mode: machine.PinOutput})
```
deadprogram: i would remove these from the `New` and `NewDual()` functions, and instead add `Configure()` method like this:

```go
// Configure sets up the device.
func (d *Device) Configure() {
	for _, pin_ := range d.pins {
		pin.Configure(machine.PinConfig{Mode: machine.PinOutput})
	}
}
```

Then you can add one to `DualDevice` like this:

```go
// Configure sets up the device.
func (dd *DualDevice) Configure() {
	dd.devices[0].Configure()
	dd.devices[1].Configure()
}
```

easystepper/easystepper.go
```diff
@@ -26,31 +31,97 @@ func New(pin1, pin2, pin3, pin4 machine.Pin, steps int32, rpm int32) Device {
...
-			stepN = int8(s % 4)
-		} else {
-			stepN = int8((s + 2*(s%2)) % 4)
+		d.moveDirectionSteps(direction, s)
+	}
+}
+
+// Off turns off all motor pins
+func (d *Device) Off() {
+	d.pins[0].Low()
```
deadprogram: Suggest instead:

```go
for _, pin_ := range d.pins {
	pin.Low()
}
```

deadprogram:
I've added a few more.

conejoninja:
Thanks! changes made

deadprogram:
:car: now merging, thanks @conejoninja

PR #102, opened 2019-12-05
WS2812: Add support for Cortex-M4 120MHz

misterikkit:
This adjusts the timing for 120MHz as seen on atsamd51 boards. The
timing was kept as close as possible to the Cortex-M4 64MHz logic in
ws2812_m4_64m.go.

Tested: This works on ItsyBitsy-M4

### DIFF
--- a/ws2812/ws2812_m4_120m.go
+++ b/ws2812/ws2812_m4_120m.go
@@ -1,0 +1,173 @@
+// +build atsamd51
+
+package ws2812
+
+// This file implements the WS2812 protocol for 120MHz Cortex-M4
+// microcontrollers.
+// Note: This implementation does not work with tinygo 0.9.0 or older.
+
+import (
+	"device/arm"
+)
+
+// Send a single byte using the WS2812 protocol.
+func (d Device) WriteByte(c byte) error {
+	// For the Cortex-M4 at 120MHz
+	portSet, maskSet := d.Pin.PortMaskSet()
+	portClear, maskClear := d.Pin.PortMaskClear()
+
+	// See:
+	// https://wp.josh.com/2014/05/13/ws2812-neopixels-are-not-so-finicky-once-you-get-to-know-them/
+	// T0H: 32-34   cycles or  266.67ns -  283.33ns
+	// T0L: 101-103 cycles or  841.67ns -  858.33ns
+	//   +: 133-137 cycles or 1108.33ns - 1141.67ns
+	// T1H: 73-75   cycles or  608.33ns -  625.00ns
+	// T1L: 58-60   cycles or  483.33ns -  500.00ns
+	//   +: 131-135 cycles or 1091.67ns - 1125.00ns
+	// A branch is treated here as 1-3 cycles, because apparently it might get
+	// speculated. This is more of a guess than hard fact, because the only docs
+	// by ARM that state this are now considered superseded (by what?).
+	value := uint32(c) << 24
+	arm.AsmFull(`
+	send_bit:
+		str   {maskSet}, {portSet}     @ [2]   T0H and T0L start here
+		lsls  {value}, #1              @ [1]
+		nop                            @ [28]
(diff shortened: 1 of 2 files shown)


bgould:
I tested this using the on-board neopixel on Metro M4 Airlift, and it works pretty well.  I adapted this example https://github.com/tinygo-org/drivers/blob/master/examples/apa102/itsybitsy-m0/main.go to use ws2812, and when I run it I see the expected effect however there is an almost imperceptible "flicker" occassionally, even when I exclude the goroutine and remove timer.Sleep to make sure the scheduler isn't tripping things up.  It might be a problem on my end but just wanted to mention it.  In general this contribution is serviceable and appreciated by me.

misterikkit:
@bgould Thanks for the feedback! I have also seen occasional color flickers driving 24 neopixels from a GPIO pin. Looking into this some more, I observe no flicker when running off a 3.7V LiPo battery. When I connect to USB power on my PC, I can get somewhat consistent flickering, but only in the first few moments after resetting the board. **Question**: Do you observe flicker after the board has been on a couple seconds (since restart)? Or with a different power supply?

When I switch over to my laptop's charger (USB-C, I wired my own adapter to the ItsyBitsy board) I do not get any flickering. As far as USB power goes, I am not doing any power delivery protocols, so I should be getting 500mA or 900mA, depending on whether it's actually USB 3.0. Either way, it's reasonable to assume that a 65W power brick can provide cleaner power than my desktop PC. (Maybe one is actually supplying 400mA more than the other?)

Based on my reading, the WS2812 timing in this implementation is aggressively low, and we can afford to move closer to nominal values. **I modified the timing to the following, but it didn't eliminate flickering.** This may be more related to voltage & current than actual signalling, but that is beyond my level of understanding.

If I could compare signals with a logic analyzer, it might become apparent what the problem is. If this adjusted timing has no practical effect on the flicker problem, then I will not bother updating the PR.


x|min cycles|max cycles|min ns|max ns
--|-- | -- | -- | --
T0H | 42 | 44 | 350.00 | 366.67
T0L | 110 | 112 | 916.67 | 933.33
sum | 152 | 156 | 1266.67 | 1300.00
T1H | 108 | 110 | 900.00 | 916.67
T1L | 42 | 44 | 350.00 | 366.67
sum | 150 | 154 | 1250.00 | 1283.33

aykevl:
The flickering is caused by interrupts coming from the USB peripheral. That's why you don't see them when running from a LiPo battery or when running directly from USB without data line.

You can work around it by disabling/enabling interrupts while sending the colors, [like this](https://github.com/aykevl/things/blob/7df097748a0d7f438d025f579875cd25a5e95517/spirit-level/main.go#L76-L78):

```go
mask := arm.DisableInterrupts()
strip.WriteColors(leds)
arm.EnableInterrupts(mask)
```

bgould:
FWIW I disabled the interrupts as @aykevl indicated, and the flickering went away entirely.

aykevl:
Okay, thanks for confirming @bgould! Will merge now.

misterikkit:
> You can work around it by disabling/enabling interrupts while sending the colors, [like this](https://github.com/aykevl/things/blob/7df097748a0d7f438d025f579875cd25a5e95517/spirit-level/main.go#L76-L78):

Thanks for the tip! Although when I was digging through the source, I recall seeing that `arm.DisableInterrupts()` returns 0 unconditionally. Since I never got usb-initiated flashing working on Windows, I think I'll be fine though.

aykevl:
> Although when I was digging through the source, I recall seeing that `arm.DisableInterrupts()` returns 0 unconditionally.

True, it has not yet been fully implemented (due to incomplete inline assembler support). The limitation is that it doesn't currently allow nesting these calls, but when you do that you're probably making things too complicated anyway. However, it should definitely disable interrupts, so you could try and see whether it fixes the occasional flicker you've seen.

PR #105, opened 2019-12-14
ws2812: add support for simulation

aykevl:
Call a special `__tinygo_ws2812_write_byte` function to send a single byte. It is implemented in the playground with the following PR: https://github.com/tinygo-org/playground/pull/6


deadprogram:
:traffic_light: merging.

PR #106, opened 2019-12-15
Decouple net package from espat

bgould:
This pull request moves the `net`, `mqtt`, and `tls` packages out of `espat` and defines a `DeviceDriver` interface to make it possible to switch between network devices.

See: https://github.com/tinygo-org/drivers/pull/98#issuecomment-562983590

I've updated the `espat` examples to use the new `net` package but I don't have a good setup right now for testing for regressions in the `espat` package, so if someone could test some existing `espat` that would be very helpful.


deadprogram:
Tested `examples/espat/mqttclient` with success using both tcp and tls connection:

```
Terminal ready
Connecting to wifi adaptor...
Connecting to wifi adaptor...
Connected to wifi adaptor.
Connecting to wifi network 'rems1'
Connected.
AT+CIPSTA?
+CIPSTA:ip:"10.0.0.20"
+CIPSTA:gateway:"10.0.0.1"
+CIPSTA:netmask:"255.255.255.0"

OK

Connecting to MQTT broker at tcp://test.mosquitto.org:1883
Publishing MQTT message...
Publishing MQTT message...
...
```

Also tried `examples/espat/mqttpub` with success using tls connection:

From my point of view, that is a very successful refactoring.

We can consider changes to interfaces in later PRs, such as those suggested in comments you have left.

@bgould thank you so very much for this awesome work!

PR #107, opened 2019-12-21
added generic 8bit shift register

conejoninja:
Simple driver for generic 8/16/32 bit shift registers.

- ReadXXInput, but open to a better name
- One small issue could be the order to read the inputs, not sure if I need to make it configurable

- This will not pass the Ci until pybadge board is added to tinygo

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ The following 34 devices are supported.
 | [MMA8653 accelerometer](https://www.nxp.com/docs/en/data-sheet/MMA8653FC.pdf) | I2C |
 | [MPU6050 accelerometer/gyroscope](https://store.invensense.com/datasheets/invensense/MPU-6050_DataSheet_V3%204.pdf) | I2C |
 | [PCD8544 display](http://eia.udg.edu/~forest/PCD8544_1.pdf) | SPI |
+| [Shift register](https://en.wikipedia.org/wiki/Shift_register) | GPIO |
 | [SHT3x Digital Humidity Sensor](https://www.sensirion.com/fileadmin/user_upload/customers/sensirion/Dokumente/0_Datasheets/Humidity/Sensirion_Humidity_Sensors_SHT3x_Datasheet_digital.pdf) | I2C |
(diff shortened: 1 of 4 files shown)


deadprogram:
I would recommend that you implement the `Configure()` and `Get()` methods of `Pin` 

I did this for the ADCPin interface with this driver:
https://github.com/tinygo-org/drivers/blob/master/mcp3008/mcp3008.go#L74-L92

Basically, if you do this, then you can treat a channel on the shift register just like any other GPIO in.

Again, an example from the MCP3008 driver: 
https://github.com/tinygo-org/drivers/blob/master/examples/mcp3008/main.go#L24-L32

README.md
```diff
@@ -79,6 +79,7 @@ The following 34 devices are supported.
 | [MMA8653 accelerometer](https://www.nxp.com/docs/en/data-sheet/MMA8653FC.pdf) | I2C |
 | [MPU6050 accelerometer/gyroscope](https://store.invensense.com/datasheets/invensense/MPU-6050_DataSheet_V3%204.pdf) | I2C |
 | [PCD8544 display](http://eia.udg.edu/~forest/PCD8544_1.pdf) | SPI |
+| [Shift register](https://en.wikipedia.org/wiki/Shift_register) | GPIO |
```
deadprogram: Since your current implementation is parallel-in serial out, perhaps https://en.wikipedia.org/wiki/Shift_register#Parallel-in_serial-out_(PISO) is a more correct link.

Also might be good to mention `74HC165` and `74165` somewhere which are the most common numbers for this chip, even though I think they are all clones now.

conejoninja:
I like the idea of the ADCPin-like interface (ShiftPin), but in order to get a specific value (1), you need to read all the values (8), if you want to read 2 pins, you end up reading it twice (8x2) with a considerably performance loss. I would like to keep the ReadXXInput() interfaces too, what do you thing?

deadprogram:
Additional interfaces for optimization are fine seems to me. I just also like the standard interfaces being supported for the naive use case.

deadprogram:
Refactoring is done, and tests are passing. Good work @conejoninja now merging.

PR #108, opened 2019-12-24
remove stxx default offsets

conejoninja:
Remove the default offsets (row/column) it's not possible to set them to 0 (needed for pybadge).


deadprogram:
Makes sense. Merging.

aykevl:
I think it could also be determined from the width/height (zero offset if width/height are at the max) but making it explicit is probably better anyway.

PR #109, opened 2019-12-25
Support for ADT7410 temperature sensor

bgould:
This adds basic support (reading a temperature value) for the ADT7410 temperature sensor, which among other things is found on the PyPortal board from Adafruit.

### DIFF
--- a/adt7410/adt7410.go
+++ b/adt7410/adt7410.go
@@ -5,29 +5,6 @@ import (
 	"time"
 )
 
-const (
-	// Default I2C address
-	Address = 0x48
-
-	// Temperature Value MSB Register
-	RegTempValueMSB = 0x0
-
-	// Temperature Value LSB Register
-	RegTempValueLSB = 0x1
-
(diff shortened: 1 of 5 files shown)


adt7410/adt7410.go
```diff
@@ -0,0 +1,108 @@
+package adt7410
+
+import (
+	"machine"
+	"time"
+)
+
+const (
```
deadprogram: This all could be in a `registers.go` file in the same package, to match the other i2c devices?

deadprogram:
Hi @bgould thanks for another contribution.

All of the other temperature reading sensors implement this interface:

```go
// ReadTemperature returns the temperature in celsius milli degrees (ºC/1000)
func (d *Device) ReadTemperature() (temperature int32, err error)
```

I would suggest doing the same for this one.

deadprogram:
Thanks for making the requested changes @bgould now squash/merging.

PR #111, opened 2019-12-28
wifinina: update docs and add Dockerfile to build firmware

deadprogram:
This PR provides instructions and a Dockerfile to build the WifiNINA firmware needed to use the Arduino Nano33 IoT and other boards with the Arduino official firmware.

### DIFF
--- a/wifinina/Dockerfile
+++ b/wifinina/Dockerfile
@@ -1,0 +1,28 @@
+# docker build -t wifinina .
+# docker run wifinina -v "../build/wifinina:/src/build"
+
+FROM debian:stable-slim AS esp
+WORKDIR /src
+
+RUN apt-get clean && apt-get update && \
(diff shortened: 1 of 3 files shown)


deadprogram:
@bgould let me know what you think about this PR.

bgould:
The instructions look good, though I wonder if it might be useful to note that flashing the firmware should only be necessary on Arduino Nano IOT 33 in order to upgrade or if other firmware was installed previously, otherwise it should work with that board out of the box.  I had a Nano 33 IOT that was unopened and I just tested it and confirmed that is true (I hadn't specifically tested that previously).

The Dockerfile is a great idea, that should really help I think.  Will give that a spin shortly, at first glance looks good to me.

deadprogram:
I've updated the README to include the info you suggested @bgould thanks.

deadprogram:
I made one further update that reordered the documentation a little more, which I then force pushed. Now ready.

PR #112, opened 2020-01-02
l293x: added support for h-bridge motor driver

deadprogram:
This PR adds support for the L293/L293D/L293N family of H-bridge motor drivers.

### DIFF
--- a/l293x/l293x.go
+++ b/l293x/l293x.go
@@ -1,0 +1,104 @@
+// Package l293x provides a driver to the L293/L293D H-bridge chip
+// typically used to control DC motors.
+//
+// Datasheet: https://www.ti.com/lit/ds/symlink/l293d.pdf
+//
+package l293x // import "tinygo.org/x/drivers/l293x"
+
+import (
+	"machine"
+)
+
+// Device is a motor without speed control.
+// a1 and a2 are the directional pins.
+// en is the pin turns the motor on/off.
+type Device struct {
+	a1, a2 machine.Pin
+	en     machine.Pin
+}
+
+// New returns a new Motor driver for GPIO-only operation.
+func New(direction1, direction2, enablePin machine.Pin) Device {
+	return Device{
+		a1: direction1,
+		a2: direction2,
+		en: enablePin,
+	}
+}
+
+// Configure configures the Device.
+func (d *Device) Configure() {
+	d.a1.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	d.a2.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	d.en.Configure(machine.PinConfig{Mode: machine.PinOutput})
+
+	d.Stop()
+}
+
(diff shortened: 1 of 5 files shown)


l293x/l293x.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// Device is a motor without speed control.
+type Device struct {
+	d1, d2 machine.Pin
+	power  machine.Pin
+}
+
+// New returns a new Motor driver for GPIO-only operation.
+func New(d1, d2, p machine.Pin) Device {
```
conejoninja: I'd change p => power so it's more clear what the function expect

l293x/l293x.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// PWMDevice is a motor with speed control.
+type PWMDevice struct {
+	d1, d2 machine.Pin
+	speed  machine.PWM
+}
+
+// NewWithSpeed returns a new PWMMotor driver that uses PWM to control speed.
+func NewWithSpeed(d1, d2 machine.Pin, s machine.PWM) PWMDevice {
```
conejoninja: here it's a bit more confusing if we do s => speed, maybe speedPin? or pwn?

l293x/l293x.go
```diff
@@ -0,0 +1,100 @@
...
+func (d *Device) Configure() {
+	d.d1.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	d.d2.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	d.power.Configure(machine.PinConfig{Mode: machine.PinOutput})
+
+	d.Stop()
+}
+
+// Forward turns motor on in forward direction.
+func (d *Device) Forward() {
```
conejoninja: What do you think about ClockWise / CounterClockWise instead of Forward / Backward? In a two opposite wheels configuration (a car for example) one wheel will go "backward" to move forward. I'm not a fan of CW/CCW either, but forward/backward sounds strange to me.

deadprogram: I agree the directionality is probably relative to how to have motors mounted and the pins wired, so it is a bit arbitrary.

However, other motor control software tens to use the forward/backward nomenclature. 

For example https://github.com/adafruit/Adafruit-Motor-Shield-library/blob/master/examples/MotorTest/MotorTest.ino#L24

l293x/l293x.go
deadprogram: Did some renaming based on your suggestions, and also matching the datasheet more closely.

l293x/l293x.go
deadprogram: Did some renaming based on your suggestions, and also matching the datasheet more closely.

deadprogram:
I made some changes, then rebased and force pushed. Please take another look @conejoninja

conejoninja:
Thanks for your contribution, merged.

PR #113, opened 2020-01-03
l9110x: add support for L9110x h-bridge motor driver

deadprogram:
This PR adds support for the L9110/L9110S motor driver using the same interface as that used by the L293x motor driver.


conejoninja:
This sounds familiar :thinking:     :laughing: 

Thanks for another great contribution, merged.

PR #115, opened 2020-01-06
ILI9341 TFT driver

bgould:
This PR adds support for the ILI9341 TFT driver chip: https://cdn-shop.adafruit.com/datasheets/ILI9341.pdf

This chip is found in Adafruit PyPortal and other breakouts from Adafruit, as well as cheap displays from other sources.  Protocol for communicating with the chip can be SPI, 8-bit parallel, or 16-bit parallel. 
 Current implementation only supports 8-bit parallel mode and uses atsamd51 build tag, which is compatible with PyPortal in its default configuration.  A generic SPI implementation should be easy to add as well.


ili9341/registers.go
```diff
@@ -0,0 +1,137 @@
+package ili9341
+
+type Rotation uint8
+
+const (
+
+	// register definitions based on source:
+	// https://github.com/adafruit/Adafruit_ILI9341/blob/master/Adafruit_ILI9341.h
```
deadprogram: Since this is a reinterpretatation but not the same code, I think you can leave these first 2 lines lines but remove the rest of the comments, since this is not just copying that source code.

As far as which displays this code works with, that should probably be in GoDoc readable format somewhere like in the beginning of `ili9341.go`

conejoninja:
If the SPI implementation is not done yet, we could delete spi_generic.go to avoid confusion.

I think [color definitions](https://github.com/bgould/drivers/blob/feature/ili9341/ili9341/registers.go#L110) in the registers.go file should be moved to the example files instead. And get rid of _debug lines, or standardize its use in other drivers. What do you think?

bgould:
Updated as per feedback above, also added documentation and smoke test.  Removed WIP, this is ready for review

deadprogram:
Thanks for the great work on this @bgould now squash/merging.

:framed_picture:

PR #116, opened 2020-01-06
Added nrf52840 tag to ws2812

bgould:
This adds the nrf52840 tag to the WS2812 driver to support boards like Circuit Playground Bluefruit


aykevl:
Looks good, thank you!

PR #118, opened 2020-01-10
Adding driver for four-wire resistive touchscreen

bgould:
This is a driver for a simple resistive four-wire touchscreen like the one found on PyPortal and many inexpensive displays.

The code works but this PR is still WIP as I am doing some final testing I still need to:

Update README
Add example(s)
Validate the logic

Also I would like to get some feedback on the package layout and interface.  It might be premature for a "standard" interface but I added on called "Pointer" in the touch package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,8 @@ smoke-test:
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=pyportal ./examples/touch/resistive/fourwire/main.go
 	@md5sum ./build/test.hex
+	tinygo build -size short -o ./build/test.hex -target=pyportal ./examples/touch/resistive/pyportal_touchpaint/main.go
+	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/vl53l1x/main.go
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=microbit ./examples/waveshare-epd/epd2in13/main.go
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,8 @@ smoke-test:
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=circuitplay-express ./examples/thermistor/main.go
 	@md5sum ./build/test.hex
+	tinygo build -size short -o ./build/test.hex -target=pyportal ./examples/touch/resistive/fourwire/main.go
+	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/vl53l1x/main.go
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=microbit ./examples/waveshare-epd/epd2in13/main.go
--- a/examples/touch/resistive/pyportal_touchpaint/main.go
+++ b/examples/touch/resistive/pyportal_touchpaint/main.go
@@ -11,12 +11,7 @@ import (
 )
 
 var (
-	resistiveTouch = resistive.FourWireTouchscreen{
-		YP: machine.TOUCH_YD,
-		YM: machine.TOUCH_YU,
-		XP: machine.TOUCH_XR,
-		XM: machine.TOUCH_XL,
-	}
+	resistiveTouch = &resistive.FourWireTouchscreen{}
 
 	display = ili9341.NewParallel(
 		machine.LCD_DATA0,
@@ -57,7 +52,12 @@ func main() {
 
 	// configure touchscreen
 	machine.InitADC()
-	resistiveTouch.Configure()
+	resistiveTouch.Configure(&resistive.FourWireConfig{
+		YP: machine.TOUCH_YD,
+		YM: machine.TOUCH_YU,
+		XP: machine.TOUCH_XR,
+		XM: machine.TOUCH_XL,
+	})
 
 	// configure display
 	display.Configure(ili9341.Config{})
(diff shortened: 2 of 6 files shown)


bgould:
Added smoke test and removed WIP from title.  I think this is ready for review please.

Makefile
```diff
@@ -85,6 +85,8 @@ smoke-test:
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=circuitplay-express ./examples/thermistor/main.go
 	@md5sum ./build/test.hex
+	tinygo build -size short -o ./build/test.hex -target=pyportal ./examples/touch/resistive/fourwire/main.go
+	@md5sum ./build/test.hex
```
aykevl: The PyPortal example should also be included here in the smoke tests.

examples/touch/resistive/pyportal_touchpaint/main.go
```diff
@@ -0,0 +1,188 @@
...
+
+	}
+
+	if (int16(touch.Y) - penRadius) > boxSize {
+		display.FillRectangle(
+			int16(touch.X), int16(touch.Y), penRadius*2, penRadius*2, currentColor)
+	}
+}
+
+/**** End of Touchpaint demo ************************************************/
```
aykevl: this looks a bit out of place in Go source code

touch/resistive/fourwire.go
```diff
@@ -0,0 +1,82 @@
...
+func (res *FourWireTouchscreen) GetTouchPoint() (p touch.Point) {
+	p.X = int(res.ReadX())
+	p.Y = int(res.ReadY())
+	p.Z = int(res.ReadZ())
+	return
+}
+
+func (res *FourWireTouchscreen) ReadX() uint16 {
+	res.YM.Configure(machine.PinConfig{machine.PinInput})
+	res.YM.Low()
```
aykevl: You cannot normally set a pin high or low when it's an input. What is the intention here?

touch/resistive/fourwire.go
```diff
@@ -0,0 +1,82 @@
...
+
+type FourWireTouchscreen struct {
+	YP machine.Pin
+	YM machine.Pin
+	XP machine.Pin
+	XM machine.Pin
+
+	yp machine.ADC
+	xm machine.ADC
+	xp machine.ADC
```
aykevl: It looks like these are essentially duplicates of the pins declared above. It should be possible (and more efficient) to declare them inline:

```
machine.ADC{res.XP}.Get()
```

examples/touch/resistive/pyportal_touchpaint/main.go
bgould: Oh yes this definitely doesn't belong there, I will remove it.

touch/resistive/fourwire.go
bgould: Ok yes, I think that would be clearer.  I wasn't sure if creating a new instance of the struct for each call to Get() would have a negative impact on anything, so I went with just keeping a copy of it around.  I will change it over to the way you've suggested

Makefile
bgould: Will do

bgould:
Thank you very much for the feedback @aykevl , I think these are all good suggestions.  

> * Externally visible names must have documentation.

Yes, no problem, I will add.

> * What's the scale used here? It _appears_ be be 0..1023, but why this particular scale? For future proofing, it may be better to store the value in the range 0..65535 (the full uint16 range), similar to the return value of the ADC. Or otherwise some human-understandable range such as 0..1000 or 0..10000 (perhaps returning the result as an int/int32 instead of uint16).

This makes sense; I think I was thinking that scaling it down to 10 bits would be the lowest common denominator for AVR/Arduino or something, but what you're saying makes sense and wouldn't force lossiness on hardware that supports better precision.

> * In other drivers, I've prefixed sensor reading methods with `Read`, I think it's best to do that here too (as `Get` suggests a getter), such as `ReadTouchPoint`.

Sounds good to me, I will adjust the code accordingly

> Also, what's the reason for exporting the `ReadX` etc methods? It seems to me that all you need is a single method to read the touch point.

Originally I think I was exposing those to provide a low-level way to read the individual values, and was going to have GetTouchPoint() possibly implement things like debouncing and converting everything to a common scale.  I practice as I moved code around, things obviously didn't turn out like that so you're right it doesn't make a ton of sense to export these.

I will give this another shot and clean things up based on your feedback.

Makefile
bgould: This has been addressed in the latest set of commits.

touch/resistive/fourwire.go
bgould: This is addressed in the latest set of commits.

touch/resistive/fourwire.go
bgould: This is addressed in the latest set of commits.

examples/touch/resistive/pyportal_touchpaint/main.go
bgould: This is addressed in the latest set of commits.

deadprogram:
Hi @bgould seems like you left in the `ReadX()` methods. Are those supposed to be removed or is it fine the way it is? I do not have a strong opinion either way.

bgould:
> Hi @bgould seems like you left in the `ReadX()` methods. Are those supposed to be removed or is it fine the way it is? I do not have a strong opinion either way.

I ended up leaving them in but changed their purpose slightly and documented it accordingly.  Before the values returned from Read* were equivalent as if you called ReadTouchPoint which returns all three.  Now, ReadTouchPoint() returns these values but averages them across several reads.  The Read* methods on the other return the "raw" value from a single read.  I left these as exported in case someone wants to embed the struct and override ReadTouchPoint.

ReadTouchPoint is the method that defines the touch.Pointer interface and really the only one that most client code should care about.

deadprogram:
Finally got my PyPortal Pynt and had a chance to try this out. Worked great, the mini-paint application is too cool!

Now squashing/merging, thanks for another very nice contribution @bgould

PR #120, opened 2020-01-28
ws2812: work-arounds to allow Digispark to control WS2812 LEDs

deadprogram:
This PR contains work-arounds to allow Digispark to control WS2812 LEDs. Seems like we discovered some problems in the AVR LLVM backend.

### DIFF
--- a/examples/ws2812/digispark.go
+++ b/examples/ws2812/digispark.go
@@ -1,0 +1,10 @@
+// +build digispark
+
+package main
+
+import "machine"
+
+// This is the pin assignment for the Digispark only.
+// Replace neo in the code below to match the pin
+// that you are using if different.
+var neo machine.Pin = 0
--- a/examples/ws2812/main.go
+++ b/examples/ws2812/main.go
@@ -1,8 +1,7 @@
-// Connects to an WS2812 RGB LED strip with 10 LEDS, such as
-// on an Adafruit Circuit Playground Express board.
+// Connects to an WS2812 RGB LED strip with 10 LEDS.
 //
-// Replace machine.NEOPIXELS in the code below to match the pin
-// that you are using, if you have a different board.
+// See either the others.go or digispark.go files in this directory
+// for the neopixels pin assignments.
 package main
 
 import (
@@ -13,12 +12,15 @@ import (
 	"tinygo.org/x/drivers/ws2812"
 )
 
+var leds [10]color.RGBA
+
 func main() {
-	neo := machine.NEOPIXELS
+	led := machine.LED
+	led.Configure(machine.PinConfig{Mode: machine.PinOutput})
+
 	neo.Configure(machine.PinConfig{Mode: machine.PinOutput})
 
 	ws := ws2812.New(neo)
-	leds := make([]color.RGBA, 10)
 	rg := false
 
 	for {
@@ -33,7 +35,8 @@ func main() {
 			}
 		}
 
-		ws.WriteColors(leds)
+		ws.WriteColors(leds[:])
+		led.Set(rg)
 		time.Sleep(100 * time.Millisecond)
 	}
 }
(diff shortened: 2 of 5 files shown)


ws2812/ws2812_avr_digispark.go
```diff
@@ -0,0 +1,50 @@
...
+		nop
+		st    {portClear}, {maskClear} ; [2]   set output low (end of pulse)
+		nop                            ; [3]
+		nop
+		nop
+		subi  {i}, 1                   ; [1]   subtract one (for the loop)
+		brne  send_bit                 ; [1/2] send the next bit, if not at the end of the loop
+	`, map[string]interface{}{
+		"value":     c,
+		"i":         iv,
```
aykevl: Minor nit (feel free to fix or not): you could also just use `byte(8)` here, which has the same effect.

ws2812/ws2812_avr_digispark.go
```diff
@@ -0,0 +1,50 @@
+// +build digispark
+
+package ws2812
+
+// This file implements the WS2812 protocol for 16MHz AVR microcontrollers.
```
aykevl: I think this should have some explanation as to why there are two files for seemingly the same functionality. Something along the lines of "this is a slightly different implementation than the one for the atmega to work around a compiler bug".

examples/ws2812/digispark/main.go
```diff
@@ -0,0 +1,44 @@
...
+	"tinygo.org/x/drivers/ws2812"
+)
+
+var leds [4]color.RGBA
+
+func main() {
+	led := machine.LED
+	led.Configure(machine.PinConfig{Mode: machine.PinOutput})
+
+	var neo machine.Pin = 0
```
aykevl: Perhaps the two ws2812 examples could be merged into one? You could move the LED array length and the NeoPixel pin number into separate arduino.go and digispark.go files in the examples/ws2812 package (with appropriate build tags).
Additionally, this definitely should have a smoke test.

deadprogram: The part that seems to be important for the digispark variation is that `var leds [4]color.RGBA` is not inside of the `main()`.

Actually I can change the length back to 10 and it works fine.

Perhaps this can provide some clue about the nature of the backend problem.

In any case, I was not successful at getting the same example program to compile on both Arduino and Digispark.

conejoninja:
I can confirm this work fine on my digisparks :+1: 

I prefer to have both examples, so people will know digispark is a bit different and there's a workaround because its limitations.

deadprogram:
@conejoninja and @aykevl see my latest update to this branch. Basically the same example will run on CPE board with minimal change, using the build tags. Also now has smoke tests for both.

examples/ws2812/main.go
```diff
@@ -1,8 +1,8 @@
-// Connects to an WS2812 RGB LED strip with 10 LEDS, such as
-// on an Adafruit Circuit Playground Express board.
+// Connects to an WS2812 RGB LED strip with 4 LEDS
```
conejoninja: no longer true, it should be 10 LEDs. Also, not only on digispark

deadprogram: Good point, fixed.

deadprogram:
I think this PR is now ready to merge.

examples/ws2812/digispark/main.go
aykevl: (not relevant anymore, but perhaps useful information)
You could define the `leds` array inside the main function while keeping the length outside:

```go
const numLeds = 4
func main() {
    var leds [numLeds]color.RGBA
    // ...
}
```

(The `const` is important here).

examples/ws2812/digispark.go
```diff
@@ -0,0 +1,10 @@
+// +build digispark
+
+package main
+
+import "machine"
+
+// This is the pin assignment for the Digispark only.
+// Replace neo in the code below to match the pin
+// that you are using if different.
+var neo machine.Pin = 0
```
aykevl: This could be `const` instead of `var`. In practice it probably won't matter with the interp optimization pass, but `const` is a much stronger signal to the compiler.

PR #121, opened 2020-01-30
Adding scroll functionality for ili9341

bgould:
This PR adds a method for setting the vertical scroll address in the ILI9341 driver.

### DIFF
--- a/examples/ili9341/scroll/main.go
+++ b/examples/ili9341/scroll/main.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"image/color"
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/ili9341"
+)
+
+var (
+	display = ili9341.NewParallel(
+		machine.LCD_DATA0,
+		machine.TFT_WR,
+		machine.TFT_DC,
+		machine.TFT_CS,
+		machine.TFT_RESET,
+		machine.TFT_RD,
+	)
+
+	red   = color.RGBA{255, 0, 0, 255}
+	blue  = color.RGBA{0, 0, 255, 255}
+	green = color.RGBA{0, 255, 0, 255}
+	black = color.RGBA{0, 0, 0, 255}
+	white = color.RGBA{255, 255, 255, 255}
+)
+
+func main() {
+
+	machine.TFT_BACKLIGHT.Configure(machine.PinConfig{machine.PinOutput})
+
+	display.Configure(ili9341.Config{})
+	width, height := display.Size()
+
+	display.FillScreen(black)
+	machine.TFT_BACKLIGHT.High()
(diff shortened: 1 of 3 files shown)


conejoninja:
I would also add SetScrollArea and StopScroll

I found that ST7735 do also have scroll functionality (and probably ST7789 too). I used the same register name and functions and added 2 more to specify which area to scroll and to return to normal, althought SetScroll(0) works  too

```go

// SetScrollWindow sets an area to scroll with fixed top and bottom parts of the display
func (d *Device) SetScrollArea(topFixedArea, bottomFixedArea int16) {
	d.Command(VSCRDEF)
	d.Tx([]uint8{
		uint8(topFixedArea >> 8), uint8(topFixedArea),
		uint8(d.height - topFixedArea - bottomFixedArea>>8), uint8(d.height - topFixedArea - bottomFixedArea),
		uint8(bottomFixedArea >> 8), uint8(bottomFixedArea)},
		false)
}

// SetScroll sets the vertical scroll address of the display.
func (d *Device) SetScroll(line int16) {
	d.Command(VSCRSADD)
	d.Tx([]uint8{uint8(line >> 8), uint8(line)}, false)
}

// SpotScroll returns the display to its normal state
func (d *Device) StopScroll() {
	d.Command(NORON)
}
```

deadprogram:
@bgould not sure if you saw the comment from @conejoninja above?

bgould:
> I would also add SetScrollArea and StopScroll

@conejoninja Thanks, that is a great idea, I will get these added.

examples/ili9341/scroll/main.go
```diff
@@ -0,0 +1,49 @@
...
+	machine.TFT_BACKLIGHT.High()
+
+	display.FillRectangle(0, 0, width/2, height/2, white)
+	display.FillRectangle(width/2, 0, width/2, height/2, red)
+	display.FillRectangle(0, height/2, width/2, height/2, green)
+	display.FillRectangle(width/2, height/2, width/2, height/2, blue)
+	display.FillRectangle(width/4, height/4, width/2, height/2, black)
+
+	for scroll := int16(0); ; scroll = (scroll + 1) % 320 {
+		time.Sleep(7500 * time.Microsecond)
```
deadprogram: Is there a specific reason to wait first, before performing the scrolling?

conejoninja: My guess is "not", you could wait before or after the scroll, or it will move too fast. Since inside the loop there's only scroll+sleep, I think it doesn't make a difference

deadprogram: In that case, let's merge this PR.

Thanks again @bgould for another visually pleasing contribution. :scroll:

Issue #45, opened 2019-04-26
SPI Flash driver


deadprogram:
I'd like to have a driver for SPI flash drives similar to what https://github.com/adafruit/Adafruit_SPIFlash can do. Ideally, it would also support the CircuitPython extensions to the file system for when those are in use.

aykevl:
That would be useful. When defining such a driver, think of these things:

 * I think it would be best to split the filesystem from the block device driver. MicroPython has such a division and it works quite well: it enables other filesystems than just FAT (which is a rather bad filesystem for embedded devices except that it's simple). Maybe we could even use a generic FAT library, which would make testing much easier on a host system (use memory as backend, provide a FUSE interface, and stress-test the filesystem with all the tools that are out there already for testing filesystems).
 * The built-in flash of many chips can be used as well, so it would be useful to not have the FAT driver assume it uses SPI flash. In fact, I once got a FAT filesystem working on the nrf51 but it wasn't very usable due to the small amount of available flash on the chip (when using MicroPython).
 * There are basically two classes of flash: NAND and NOR. Most embedded devices have NOR flash while most SSDs have NAND flash. The difference is that NOR flash can be written to at a much smaller granularity, often just 4 bytes at a time. Also, page sizes tend to be smaller.
  * Page size varies by chip: the nrf51 has 1kB, nrf52 has 4kB and the flash chip usually shipped with the esp8266 also has a page size of 4kB.
  * An erased flash page is _usually_ all 0xff with newly written data setting bits to 0 (but not back to 1). Depending on the device, you may be able to set more bits in the word to 0 without erasing first. Some devices (IIRC the STM32L series, among others) have the bits reversed.

bgould:
FYI I'm making decent progress with this... I'm able to access the "internal" SPI flash on the Itsy Bitsy M0 and have implemented functions for doing low level IO operations on the block device.  Also I've forked https://github.com/mitchellh/go-fs and with a few tweaks have been able to use that start implementing functionality to read from a FAT filesystem.

@deadprogram I was wondering if you could please clarify what functionality you mean by "the CircuitPython extensions to the file system for when those are in use" - can you point me to any documentation on what those extensions do so that I can think about making accomodations for that as I am working?

deadprogram:
Great news about your progress!

I was looking and seems like the FAT filesystem used is now standardized? There appears to have been some MBR hacks previously, but looks like they were removed from the Adafruit code.

bgould:
Updates:

[...]

PR #124, opened 2020-02-03
Low-level IO driver for serial flash memory via SPI and QSPI

bgould:
Read functionality currently works using SPI, and QSPI on SAMD51.  Will work on QSPI for nRF chips next


bgould:
On atsamd51, to use regular SPI over the QSPI requires using a mechanism separate from SERCOM... however upon further investigation, nrf seems a little easier to use the existing SPI implementation, so I'm not going to focus on implementing QSPI on nrf right now.  Instead I'm going to work on cleaning up the code I've already got working.

conejoninja:
Great job! I tried the examples and work fine. I've been reading on this topic ( https://learn.adafruit.com/introducing-adafruit-itsybitsy-m4/circuitpython-storage ) and looks like it's needed to pull up some pins in order to get write permissions, at least in CircuitPython.

Have you looked into this too?

deadprogram:
@conejoninja I think that is a feature of CircuitPython itself:

https://github.com/adafruit/circuitpython/commit/1a0596a2fbb68c1285ec59606d03a610ee94ca47

We will probably need to implement something similar.

bgould:
I did notice pieces of that in the CircuitPython implementation... I think that is mostly related to the functionality they have where they present the flash chip as a USB drive to the host computer.  I think that they intend for most of the file to written from the host and read from CircuitPython.

I haven't gotten nearly so far yet; my intention for the scope of this driver is to be able to interface with SPI flash chips in a filesystem-agnostic way.  Some applications like data logging might not need a filesystem at all, so this could be immediately useful for that.  We should be able to support writing to/from the memory device and also write-protecting the device by just issuing the appropriate commands over SPI - https://github.com/tinygo-org/drivers/pull/124/files#diff-3eea81db4fb1722dbc92e4b8bcf45223R17

As far as next steps once the low level IO is working well enough, it might be possible to build some USB endpoints that allow for the host computer to format/write a FAT filesystem (or even just flash CircuitPython onto a board and write some files) and then I have some Go code that is mostly capable of reading those files (does not yet support writing though).  I'll probably maintain that FAT library outside of the drivers repo for a while until it is more battle-tested.  Eventually I would want to support a more full-featured solution of course, but this setup would at least provide a solution for things like being able to read wifi credentials from a file for example.  CircuitPython/Micropython use a embedded FAT implementation written in C that would be useful for that purpose, but it is going to take me a bit to wrap my head around how that should work (maybe some smarter will figure it out first :P)

As @aykevl pointed out previously, there are also embedded filesystems that are easier to implement than FAT.  I think he mentioned the micro:bit filesystem for instance, so I'm thinking if we can at least tackle the low-level IO first then we could go that direction as well.

conejoninja:
Cool! Thanks for the explanation. I was a bit puzzled , but much clear now :+1:

flash/transport_qspi_samd.go
```diff
@@ -0,0 +1,179 @@
...
+
+//go:inline
+func (q qspi) disableAndClearCache() {
+	CMCC.CTRL.ClearBits(CMCC_CTRL_CEN)
+	for CMCC.SR.HasBits(CMCC_SR_CSTS) {
+	}
+	CMCC.MAINT0.SetBits(CMCC_MAINT0_INVALL)
+}
+
+//go:inline
```
aykevl: `//go:inline` is something that should usually be left to the compiler. The compiler will do that almost always for very short functions and functions that are only called from one place.

Did you see a performance problem without `//go:inline`?

bgould:
> I didn't go through it all but there is one main concern I have with the code as it is now: QSPI is implemented directly in the flash package. I think it should go into the machine package to make it more widely available for other use cases than SPI flash.
> 
> Is there a reason to couple the flash package and the samd51 QSPI peripheral?

This a good question.  The QSPI peripheral for SAMD chips has 2 modes of operation; the first mode is as a general-purpose SPI peripheral that can be used with any sort of SPI device.  The second mode (which is the mode used in this component) is specific to serial flash memory devices.  In this mode, it is possible to perform single, dual, or quad bit transfers and to "memory map" the flash device into the 0x00400000-0x00500000 address space.

This first mode could definitely be made to fit into the current SPI abstraction, and would certainly make sense to have in the machine package if implemented.  I'm not sure there is a ton of advantage of this peripheral over using a SERCOM for SPI, except that from what I can tell the QSPI in SPI mode can run at the same speed as the CPU and the chip select line is handled automatically by the hardware.

Since the second mode is really only suitable for serial flash memory, I thought it would make sense to keep it in this driver.  I am fine with moving it out of here if that is preferred, would just need to brainstorm what the abstraction should look like for this mode of operation.

examples/flash/qspi/main.go
```diff
@@ -0,0 +1,233 @@
...
+
+func main() {
+
+	time.Sleep(3 * time.Second)
+
+	readyLED.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	readyLED.High()
+
+	readyLED.Low()
+	write("spi Configured. Reading flash info")
```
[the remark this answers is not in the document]
bgould: This is left over from code that I copied in from another program and is not relevant to the example.  I will change it.

flash/transport_qspi_samd.go
```diff
@@ -0,0 +1,179 @@
+// +build atsamd51
+
+package flash
+
+import (
+	. "device/sam"
```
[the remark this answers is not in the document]
bgould: Roger that, will change.

flash/transport_qspi_samd.go
bgould: I did see some measured improvement in another driver, but I did not measure anything for this use case.  I was just cargo-culting what I had done previously, so I will remove this.

aykevl:
> > Is there a reason to couple the flash package and the samd51 QSPI peripheral?
> 
> This a good question. The QSPI peripheral for SAMD chips has 2 modes of operation; the first mode is as a general-purpose SPI peripheral that can be used with any sort of SPI device. The second mode (which is the mode used in this component) is specific to serial flash memory devices.

Okay, thank you for explaining. It's not ideal but I agree it's best to keep it in the driver for now. Maybe there are other chips with a similar abstraction at which point it can be split up and moved into the machine package. Coming up with a good abstraction is almost impossible without at least two or three implementations to see whether the abstraction is the right one.

flash/transport.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func (err Error) Error() string {
+	switch err {
+	case ErrInvalidClockSpeed:
+		return "invalid clock speed"
+	case ErrInvalidAddrRange:
+		return "invalid address range"
+	default:
+		return "unspecified error"
```
aykevl: Nit: it would be helpful to include the source of the error, such as `flash: invalid clock speed`. That should help in the case the error is printed as-is (using `println` for example).

flash/attrs.go
```diff
@@ -0,0 +1,53 @@
...
+
+import "fmt"
+
+type JedecID struct {
+	ManufID  uint8
+	MemType  uint8
+	Capacity uint8
+}
+
+func (id *JedecID) AsUint32() uint32 {
```
aykevl: I think just calling this `Uint32` is more Go style. See for example [reflect](https://godoc.org/reflect#Value.Uint).

flash/device.go
```diff
@@ -0,0 +1,122 @@
...
+	return dev.transport.readMemory(addr, buf)
+}
+
+func (dev *Device) ReadStatus() (status byte, err error) {
+	buf := make([]byte, 1)
+	err = dev.transport.readCommand(CmdReadStatus, buf)
+	return buf[0], err
+}
+
+func (dev *Device) ReadStatus2() (status byte, err error) {
```
aykevl: What's the difference with `ReadStatus`?

bgould: There are multiple status registers on these chips, ReadStatus() reads the first and ReadStatus2() reads the second.  I've added comments to clarify that.

flash/attrs.go
bgould: Done

flash/transport.go
bgould: good point, added this.

flash/transport_qspi_samd.go
bgould: removed

flash/flash.go
```diff
@@ -0,0 +1,370 @@
...
+		return 0, err
+	}
+	return SerialNumber(uint64(sn[11]) | uint64(sn[10])<<0x8 |
+		uint64(sn[9])<<0x10 | uint64(sn[8])<<0x18 | uint64(sn[7])<<0x20 |
+		uint64(sn[6])<<0x28 | uint64(sn[5])<<0x30 | uint64(sn[4])<<0x38), nil
+}
+
+// ReadBuffer fills the provided buffer with memory read from the device
+// starting at the provided address.
+func (dev *Device) ReadBuffer(addr uint32, buf []byte) error {
```
aykevl: I think the `io.ReaderAt` interface would be a good fit here.

flash/flash.go
```diff
@@ -0,0 +1,370 @@
...
+	if err := dev.WaitUntilReady(); err != nil {
+		return err
+	}
+	return dev.trans.readMemory(addr, buf)
+}
+
+// WriteBuffer writes data to the device, one page at a time, starting at the
+// provided address. This method assumes that the destination is already erased.
+func (dev *Device) WriteBuffer(addr uint32, buf []byte) (n int, err error) {
```
aykevl: Same here, it seems to me like `io.WriterAt` would be a good fit here.
The extra requirement of needing to erase first still applies of course, but I don't think that disqualifies SPI flash from implementing the interface.

flash/flash.go
```diff
@@ -0,0 +1,370 @@
...
+
+// EraseBlock erases a block of memory at the specified address
+func (dev *Device) EraseBlock(addr uint32) error {
+	if err := dev.WaitUntilReady(); err != nil {
+		return err
+	}
+	if err := dev.WriteEnable(); err != nil {
+		return err
+	}
+	return dev.trans.eraseCommand(cmdEraseBlock, addr*BlockSize)
```
aykevl: It looks like `addr` isn't a real address but instead is a block number?
Same goes for `EraseSector`.

flash/flash.go
```diff
@@ -0,0 +1,370 @@
...
+		return err
+	}
+	if err := dev.WriteEnable(); err != nil {
+		return err
+	}
+	return dev.trans.eraseCommand(cmdEraseSector, addr*SectorSize)
+}
+
+// EraseChip erases the entire flash memory chip
+func (dev *Device) EraseChip() error {
```
aykevl: I would name this `EraseAll`, to erase the whole area covered by this driver (which for this driver is the entire chip). The reason is that other APIs may be provided for on-board flash. Most microcontrollers allow erasing/writing parts of their internal flash and it would be useful to provide a method that erases the whole area assigned for this purpose (but not the entire chip, which would also erase the firmware).

bgould:
I agree that it is possible and preferable to distill a smaller and more general interface out of this code, and also I agree 100% about having an implementation over on-chip flash (I was even considering if this package should be renamed to "spiflash" or something for that distinction).  The main reason I was more focused on the low-level implementation at this time is that I thought it might be good to try implementing a couple of filesystems first to see if a sensible interface emerges out of that.

That said, the FAT library I've been messing with has this interface, which sort of covers the aspects you mentioned (implements ReaderAt/WriterAt, etc): https://github.com/bgould/go-fs/blob/tinygo/block_device.go

On the other hand, I've also been experimenting with LittleFS which has a BlockDevice that looks more like this for its hardware callbacks:
```
type BlockDevice interface {
	ReadBlock(block uint32, offset uint32, buf []byte) error
	ProgramBlock(block uint32, offset uint32, buf []byte) error
	EraseBlock(block uint32) error
	Sync() error
}
```
That can probably be made to play nice with the aforementioned interface (as long as Sync() is accounted for)

Also I was planning to look into SPIFFS next which has roughly this interface for its callbacks (https://github.com/pellepl/spiffs/wiki/Integrate-spiffs):
```
  void my_spi_read(int addr, int size, char *buf)
  void my_spi_write(int addr, int size, char *buf)
  void my_spi_erase(int addr, int size)
```
so that is basically ReaderAt+WriterAt as you suggested, plus a method for erase

I will keep those things in mind and see if I can finish getting some filesystem code written to prove out what a small set of common functionality might look like, and then if you prefer I can go through this low-level driver and un-export functions that do not necessarily need to be exposed.

flash/flash.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+// Uint32 returns the JEDEC ID packed into a uint32
+func (id JedecID) Uint32() uint32 {
+	return uint32(id.ManufID)<<16 | uint32(id.MemType)<<8 | uint32(id.Capacity)
+}
+
+// String implements the Stringer interface to print out the hex value of the ID
+func (id JedecID) String() string {
+	return fmt.Sprintf("%06X", id.Uint32())
```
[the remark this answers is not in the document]
bgould: Good point, we can definitely do without fmt in this package

flash/flash.go
bgould: You're right, this is likely a bug, and certainly the variable name is misleading.  I will double check the functionality and fix up accordingly.

flash/flash.go
bgould: Sounds good, will rename

aykevl:
Thank you for the detailed explanation!

I've read a bit more about flash sizes (including [this post](https://electronics.stackexchange.com/questions/409822/nor-flash-memory-full-chip-vs-block-vs-sector-erase)) and it seems like:

 * "page size" refers to the smallest writable area. Because this is NOR flash, if you need to write a smaller area you have to extend it on both sizes with all-ones to emulate writing to a part of it (zeroes will not be set back to ones after being written).
 * "sector size" refers to the smallest erasable area.
 * "block size" is a larger area that can be erased, probably as an optimization.

So what do you think of this interface?

```go
type BlockFlash interface {
    // Size returns the size of this memory, in bytes.
    Size() int64

    // ReadAt reads the given number of bytes from the flash storage.    
    io.ReaderAt    
    
    // WriteAt overwrites an area of flash with the given data. The to be
    // written area must first be erased using EraseBlocks, the behavior is
    // otherwise memory dependent. Some flash memories will only set ones back
    // to zeroes, others may have a different behavior.
    // This function should work correctly for any address and buffer size, but
    // clients may use WriteBlockSize as an optimization.
    io.WriterAt
    
    // WriteBlockSize returns the block size in which data can be written to
    // memory. It can be used by a client to optimize writes, non-aligned writes
    // should always work correctly.
    WriteBlockSize() int64
    
    // EraseBlockSize returns the smallest erasable area on this particular chip
    // in bytes. This is used for the block size in EraseBlocks.
    EraseBlockSize() int64
    
    // EraseBlocks erases the given number of blocks. An implementation may
    // transparently coalesce ranges of blocks into larger bundles if the chip
    // supports this. The start and len parameters are in block numbers, use
    // EraseBlockSize to map addresses to blocks.
    EraseBlocks(start, len int64) error
}
```

I have no strong feelings about the particular naming of methods, but would like to make it as generic as reasonably possible (this is why I called `EraseBlockSize` that way and not `SectorSize`).
I don't have a lot of experience with SPI flash but I did look at on-chip flash in the past and that kind of flash tends to have very small write block sizes (just 2 or 4 bytes for the chips I looked at).

I used `int64` for offsets instead of `uint32`. This is for future extensibility and to match `io.ReaderAt`, the `Size()` method, etc. Using big integers is generally quite fast: it mainly requires a bit more storage and perhaps twice the operations. The only thing that is much slower is divide, but only if the compiler cannot optimize it to a multiply/bitshift/etc.

I intentionally used block numbers instead of addresses in `EraseBlocks`. This shields the SPI flash implementation from having to think about non-aligned cases with associated error conditions and forces the client to actually call `EraseBlockSize` to get the right block size. The block size will likely be core to the design of the filesystem so it's best to make that explicit. Also, by giving a range of blocks the implementation is free to use a larger erase size for consecutive blocks if that fits in the range.

>  The main reason I was more focused on the low-level implementation at this time is that I thought it might be good to try implementing a couple of filesystems first to see if a sensible interface emerges out of that.

That's a very smart choice. Maybe we should wait a bit until pinning down one particular API.

deadprogram:
Any action on this PR? I just really really want this functionality. :1234:

bgould:
@deadprogram I've been testing this out with https://github.com/bgould/go-littlefs for a bit and it working pretty well, especially for simple/common use cases.  I have a few tweaks to make as per the suggestions from @aykevl above, but those should be quick and then I think this can be reviewed/merged.  I'll try to get wrapped up in the next couple of days.

deadprogram:
That is very exciting!

bgould:
This is ready for review please.  I've incorporated the suggestions from @aykevl above, except that I did not actually declare the BlockFlash interface in the package yet.  The flash.Device struct does satisfy the suggested interface however, so it should be simple to add that later if it turns out that is the contract we'd like use for flash block devices.

This can be tested using examples/flash/console/spi (should work out of the box with Itsy Bitsy M0) and examples/flash/console/qspi (should work out of the box with PyPortal or Itsy Bitsy M4).  This example allows for querying the details about the chip and performing some low-level operations.

More practical examples how a filesystem can be implemented on top of this driver can be found at github.com/bgould/go-littlefs/tree/master/examples/flash/spi or github.com/bgould/go-littlefs/tree/master/examples/flash/qspi - these examples are similar to the ones above except that there commands like ls, cat, mkdir, etc for operating on files.  You'll first want to `format` and `mount` in order to initialize the filesystem on a new chip.  Also using those examples require the dev branch of TinyGo with this patch included: https://github.com/tinygo-org/tinygo/pull/997

aykevl:
@bgould I've experimented with the ability to mount filesystems in the os package, see https://github.com/tinygo-org/tinygo/pull/1012. Any feedback would be appreciated (whether it works for littlefs and whether we would need such an abstraction at all).

flash/flash.go
```diff
@@ -0,0 +1,407 @@
...
+		return err
+	}
+	if err := dev.trans.runCommand(cmdReset); err != nil {
+		return err
+	}
+
+	// Wait for the reset - 30us by default
+	const delay = int64(30 * time.Microsecond)
+	for stop := time.Now().UnixNano() + delay; stop > time.Now().UnixNano(); {
+	}
```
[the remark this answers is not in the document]
bgould: Yes you're right... when I started this a while back there was a bug in the scheduler that actually would freeze the program for sleeps under a certain threshold on samd chips, so I worked around it with the above loop.  I will change this to time.Sleep, as it won't hurt if it sleeps longer and maybe in the future even shorter sleep periods will be supported anyhow.

bgould:
Hi @aykevl I am taking a look at https://github.com/tinygo-org/tinygo/pull/1012 and will add some comments over there, I think it is a good idea and should be very useful I think.

bgould:
What do you think, can this be merged?  I think this is in a pretty good state for implementing filesystems on top of.

deadprogram:
This is just amazing work @bgould thank you for another impressive contribution.

Now squash/merging, I assume there will be another PR to add some docs on "how to wire everything up into a usable setup" as mentioned https://github.com/tinygo-org/drivers/issues/45#issuecomment-612093494

PR #126, opened 2020-02-06
README: Use degree sign instead of ordinal

xordspar0:
º is the masculine ordinal in Spanish. ° is the degree sign. They look different in some fonts.


aykevl:
Thank you! That's definitely hard to spot.
@deadprogram FYI

conejoninja:
I made this PR https://github.com/tinygo-org/drivers/pull/127 that changes it everywhere else.

PR #127, opened 2020-02-08
Fix degree sign

conejoninja:
Following PR #126 , the degree sign was used in several more places


deadprogram:
Thanks! Merging.

PR #128, opened 2020-02-21
semihosting: add example

aykevl:
I could also add a smoke test if needed.


deadprogram:
Thanks for adding this @aykevl now merging.

PR #129, opened 2020-02-23
some fixes on ili9341 driver

conejoninja:
- Add SetRotation on Configure func to avoid calling it later
- Fix SetPixel x coord


deadprogram:
Thanks for the fixes @conejoninja now merging.

PR #130, opened 2020-02-23
Wifinina typo

conejoninja:
Didn't test the change myself yet, but looks like a typo


deadprogram:
Seems like the change to `examples/ili9341/pyportal_boing/main.go` should not be in this PR.

The other change looks correct to me, good catch @conejoninja

conejoninja:
fixed, thanks

deadprogram:
Now merging, thanks @conejoninja

PR #133, opened 2020-03-01
new API for shifter driver.

conejoninja:
this doesn't break existing code, but it might no longer work as expected

### DIFF
--- a/examples/shifter/main.go
+++ b/examples/shifter/main.go
@@ -7,42 +7,47 @@ import (
 	"tinygo.org/x/drivers/shifter"
 )
 
+const (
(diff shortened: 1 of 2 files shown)


deadprogram:
I really love this change, it will make it so much simpler to use boards like the PyBadge.

Agreed that it changes the behavior, maybe we should add more to the GoDocs in this driver as part of this same PR to explain how to use it.

What do you think?

conejoninja:
Updated docs, not sure how better I could explain it, open to suggestions :)

deadprogram:
Looking good @conejoninja now merging.

PR #135, opened 2020-03-01
Shiftregister implementation

simulot:
Here is an implementation of Shift Output register Register using a IC like SN74HC595 using GPIO.

Tested with an Nucleo-F103RB and an Arduino Uno R3


Makefile
```diff
@@ -121,5 +121,7 @@ smoke-test:
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=arduino-nano33 ./examples/l9110x/speed/main.go
 	@md5sum ./build/test.hex
+	tinygo build -size short -o ./build/test.hex -target=nucleo-fb103rb ./examples/shiftregiser/main.go
```
conejoninja: wrong target, this makes the CI to fail

README.md
```diff
@@ -88,6 +88,7 @@ The following 44 devices are supported.
 | [Resistive Touchscreen (4-wire)](http://ww1.microchip.com/downloads/en/Appnotes/doc8091.pdf) | GPIO |
 | [Semihosting](https://wiki.segger.com/Semihosting) | Debug |
 | [Shift register](https://en.wikipedia.org/wiki/Shift_register#Parallel-in_serial-out_\(PISO\)) | GPIO |
```
conejoninja: I would rewrite both lines to 
- Shift register (PISO)
- Shift register (SIPO)

So the difference is clear between both options

shiftregister/shiftregister.go
```diff
@@ -0,0 +1,66 @@
+// Package shiftregister is for 8bit shift output register using 3 GPIO pins
+//
+// Tested with SN74HC595 with 8 outputs
```
conejoninja: Looks like the driver is compatible with several ICs, better to avoid specific wiring or datasheet, just name some of the compatible ICs like : https://github.com/tinygo-org/drivers/blob/master/shifter/shifter.go#L1

shiftregister/shiftregister.go
```diff
@@ -0,0 +1,66 @@
...
+		out:   Out,
+		bits:  Bits,
+	}
+}
+
+// Configure set hardware configuration
+func (d *Device) Configure() {
+	d.latch.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	d.clock.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	d.out.Configure(machine.PinConfig{Mode: machine.PinInput})
```
conejoninja: why is configured as input at first? but as output in WriteMask?

simulot:
Thanks for your wise comments.

This was my first PR. So, I'm not comfortable with the review process. Maybe you have some reading advices for me.

Thanks for debunking  the pin configuration error!
I'll apply your suggestions and will work on an API to manage register outputs individually.

conejoninja:
I hope it's your first PR of many more to come! The review process might be a little intimidating and difficult to read the comments, but we are here to help.

The first step is to get the [CI tests](https://circleci.com/gh/tinygo-org/drivers/395?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) to pass, right now is failing the smoke tests because of the target , [click here or in  "ci/circleci: build — Your tests failed on CircleCI
Details"](https://circleci.com/gh/tinygo-org/drivers/395?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) above. And them, some minor details :

- [ ] Fix smoke tests target (-target=nucleo**-**fb103rb  should be -target=nucleofb103rb)
- [ ] Change the line in README.md so it looks like the other existing shift register driver, I suggest something like :
    Shift register (PISO)
    Shift register (SIPO)
- [ ] Modify the comment/header of the driver file, to make it more general, use [this](https://github.com/tinygo-org/drivers/blob/master/shifter/shifter.go#L1) as an example 
- [ ] Modify the API to be able to set individual pins
- [ ] I personally would remove one of the examples' pin configuration. No other example uses +build tags. Assume the user is using only one of the boards and has wired it correctly.

README.md
simulot: Should I change the other entry in README.md?
I think we  should keep the existing driver's folder name as it is to not break existing code.

shiftregister/shiftregister.go
simulot: Oops... my fault.

simulot:
>  Fix smoke tests target (-target=nucleo-fb103rb should be -target=nucleofb103rb)

This one was source of trouble. `nucleofb103rb` is the build flag for the board, and `nucleo-fb103rb` is the target name.

shiftregister/shiftregister.go
simulot: I just have this one SN74HC595. No idea about  possible other references 

>     * Modify the comment/header of the driver file, to make it more general, use [this](https://github.com/tinygo-org/drivers/blob/master/shifter/shifter.go#L1) as an example

conejoninja: I'm not sure either, but here are some of the most common I found: SN74ALS164A , SN74AHC594 , SN74AHC595, ...

simulot:
* [X]  Fix smoke tests target (-target=nucleo-fb103rb  should be -target=**nucleo-f103rb**)
* [X]  Change the line in README.md so it looks like the other existing shift register driver, I suggest something like :
       Shift register (PISO)
       Shift register (SIPO) 
* [X]  Modify the comment/header of the driver file, to make it more general, use [this](https://github.com/tinygo-org/drivers/blob/master/shifter/shifter.go#L1) as an example
* [X ]  Modify the API to be able to set individual pins
* [X]  I personally would remove one of the examples' pin configuration. No other example uses +build tags. Assume the user is using only one of the boards and has wired it correctly.

README.md
```diff
@@ -87,8 +87,8 @@ The following 44 devices are supported.
 | [PCD8544 display](http://eia.udg.edu/~forest/PCD8544_1.pdf) | SPI |
 | [Resistive Touchscreen (4-wire)](http://ww1.microchip.com/downloads/en/Appnotes/doc8091.pdf) | GPIO |
 | [Semihosting](https://wiki.segger.com/Semihosting) | Debug |
-| [Shift register](https://en.wikipedia.org/wiki/Shift_register#Parallel-in_serial-out_\(PISO\)) | GPIO |
-| [Shift Registers Serial In Parallel Out (SIPO)](https://en.wikipedia.org/wiki/Shift_register#Serial-in_parallel-out_(SIPO)) | GPIO |
+| [Shift register (PISO)](https://en.wikipedia.org/wiki/Shift_register#Parallel-in_serial-out_\(PISO\)) | GPIO |
+| [Shift Registers (SIPO)](https://en.wikipedia.org/wiki/Shift_register#Serial-in_parallel-out_(SIPO)) | GPIO |
```
conejoninja: Really tiny detail and I think we are good to merge this, but the lines are different _register_ != _Registers_, please modify so they look alike:

[Shift register (PISO)]
[Shift register (SIPO)]

conejoninja: Also, remove these lines https://github.com/simulot/drivers/blob/shiftregister/shiftregister/shiftregister.go#L3-L8

conejoninja:
Thanks for these PR @simulot , now merging

PR #137, opened 2020-03-06
shifter: simplify API surface for PyBadge

deadprogram:
This PR is intended to simplify API surface and use build directive to directly match the PyBadge.


conejoninja:
Thanks, merging

PR #138, opened 2020-03-10
AMG88xx thermal camera module

conejoninja:
I have the feeling I'm missing something, but don't know what.

Interpolation to get a 24x24 image instead of 8x8 should be possible and will be added as example at a later time if I could, but since the driver is working I don't want to wait for this PR.


deadprogram:
This is just too cool. I want one of these cameras now!

Merging, thanks @conejoninja

conejoninja:
It's a feather so it fits in the GoBadge ! 
https://www.adafruit.com/product/3622
or if you prefer it from spain : https://tienda.bricogeek.com/adafruit-feather-wings/1035-camara-termica-amg8833-featherwing.html

PR #139, opened 2020-03-15
[WIP] MAX7219 8x8 LED Matrix

samclane:
[Whitepaper](https://www.maximintegrated.com/en/products/power/display-power-control/MAX7219.html)

Can use `maxSingle(row byte, col byte)` to write to the internal registers of the `max`, lighting up LEDs in row `row` with the bit-string `col`. 

Currently doesn't support daisy-chaining, as I can only test on 1 MAX chip.

Tested on Arduino Uno on Ubunutu Bullseye.


conejoninja:
Hi Sam, thanks for your contribution. I have several matrix modules and ICs+leds and can help you with the daisy-chaining code or verify it works, add it if you could.

Some additional notes while you are working on this PR:
- PR should be againts _dev_ instead of _master_ (I changed that for you)
- One single commits are easier to review instead of multiple, once you finish, if you can squash them into a single commit it will make the whole process way easier
- Functions has to be properly commented (see other drivers as example)
- The driver itself has to be added to README.md file and smoke tests to Makefile
- These lines https://github.com/samclane/drivers/blob/master/max7219/max7219.go#L20-L26 should be moved to a "Configure" function instead of "New" (see other drivers as example)

conejoninja:
It's been a while since the last update, is it ready to review? I'd like to add this driver as I have several max72xx modules lying around. The smoke test is failing, but not sure if something else needs to be modified before accepting this.

PR #141, opened 2020-03-30
Add driver for TMP102 low-power digital temperature sensor

yannishuber:
This adds a driver for the [TMP102](https://download.mikroe.com/documents/datasheets/tmp102-data-sheet.pdf) low-power digital temperature sensor.

Tested on the HiFive1 Rev B with the example code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ func main() {
 
 ## Currently supported devices
 
-The following 45 devices are supported.
+The following 46 devices are supported.
 
 | Device Name | Interface Type |
 |----------|-------------|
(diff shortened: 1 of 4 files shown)


deadprogram:
Hi, @yannishuber thank you for the contribution. A couple of comments:

- For all devices that return a temperature, we have been using this signature: https://github.com/tinygo-org/drivers/blob/master/bmp180/bmp180.go#L83 This allows us to avoid use of float values, unless really needed. Also by using this same interface, you can swap between any device in this category.

- We usually have a const named `Address` with the default I2C address for that device: https://github.com/tinygo-org/drivers/blob/master/bmp180/registers.go#L5-L6

What do you think?

aykevl:
Additionally to what Ron said, I think it's best to standardize on a single SI unit (Celsius). This means that all temperature sensors have a single interface. The conversion to Fahrenheit is very simple so if an application wants to display in Fahrenheit it can do that. There is no reason to repeat it in every temperature sensor driver.

yannishuber:
Thanks for the feedback, I made some changes accordingly. The signature of `ReadTemperature()` now matches the one of the other devices. I also removed the unit conversion.

deadprogram:
Looks good, thanks for making the changes.

Now squashing/merging.

PR #142, opened 2020-04-06
mqtt: reduce use of goroutines in router

deadprogram:
This PR reduces the use of goroutines in the MQTT router to not start a new goroutine for each invocation of each callback. This helps correct https://github.com/tinygo-org/tinygo/issues/1009 and also seems more realistic in an MCU.


aykevl:
I can't really comment on this but the change seems reasonable (particularly if it fixes a bug).

deadprogram:
OK, merging.

PR #143, opened 2020-04-10
Added HC-SR04 ultrasonic distance sensor.

conejoninja:


hcsr04/hcsr04.go
```diff
@@ -0,0 +1,81 @@
...
+
+const TIMEOUT = 23324 // max sensing distance (4m)
+
+// Device holds the pins
+type Device struct {
+	trigger machine.Pin
+	echo    machine.Pin
+}
+
+// New returns a new ultrasonic driver given 2 pins numbers (not pin object)
```
aykevl, line 20: > given 2 pins numbers (not pin object)

I'm not sure what to make of that. Is this maybe a comment from before the `machine.GPIO` -> `machine.Pin` transition?

conejoninja: Yes, fixed. I fixed too the driver were I copied that from

deadprogram:
Now squash/merging, thanks @conejoninja

PR #145, opened 2020-04-17
Improve performance of ILI9341 on ATSAMD5X

sago35:
I've made performance improvements.

1. clear+set in write8() is now set only.
    * Instead of being faster, it's more restrictive than ever before.
    * In the original, it had to be an 8-bit contiguous port, but this time it has to be an 8-bit contiguous port accessible by StoreUint8().
    * If this restriction is unacceptable, it is better not to merge these PRs.

### DIFF
--- a/ili9341/parallel_atsamd51.go
+++ b/ili9341/parallel_atsamd51.go
@@ -5,14 +5,14 @@ package ili9341
 import (
 	"machine"
 	"runtime/volatile"
+	"unsafe"
 )
 
 type parallelDriver struct {
 	d0 machine.Pin
 	wr machine.Pin
 
-	setPort *uint32
-	setMask uint32
+	setPort *uint8
 
 	clrPort *uint32
 	clrMask uint32
@@ -46,8 +46,12 @@ func (pd *parallelDriver) configure(config *Config) {
 	pd.wr.Configure(output)
 	pd.wr.High()
 
-	pd.setPort, _ = pd.d0.PortMaskSet()
-	pd.setMask = uint32(pd.d0) & 0x1f
+	// Calculates the address of the OUT register from the OUTSET register and obtains an address that allows 8-bit access.
+	//   OUT    : offset = 0x10
+	//   OUTSET : offset = 0x18
+	setPort, _ := pd.d0.PortMaskSet()
+	setMask := uint32(pd.d0) & 0x1f
+	pd.setPort = (*uint8)(unsafe.Pointer(uintptr(unsafe.Pointer(setPort)) - uintptr(8) + uintptr(setMask/8)))
 
 	pd.clrPort, _ = (pd.d0).PortMaskClear()
 	pd.clrMask = 0xFF << uint32(pd.d0)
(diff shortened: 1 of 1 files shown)


sago35:
test code : 7e992534be5ba34b5cba9cfba5ca0e28

* FillRectangle()
    * before : 35ms (100%)
    * with 1 : 28ms (80%)
    * with 1 and 2 : 23ms (66%)

* DrawRGBBitmap()
    * before : 83ms (100%)
    * with 1 : 59ms (71%)
    * with 1 and 2 : 52ms (63%)

sago35:
This PR will certainly make you faster, but it will also increase your limitations.
If it looks like it would be better to merge, I'll consider organizing the code a bit more.

Do you have any thoughts on that?

deadprogram:
@conejoninja what do you think?

ili9341/parallel_atsamd51.go
```diff
@@ -72,15 +80,52 @@ func (pd *parallelDriver) write16(data uint16) {
...
+		}
 	}
 }
 
 //go:inline
 func (pd *parallelDriver) write16sl(data []uint16) {
-	for i, c := 0, len(data); i < c; i++ {
+	c := len(data)
+	i := 0
+	for i = 0; i < c-8; i += 8 {
```
conejoninja, line 110: why 2 bucles ?
You are doing less loops, but  more code, is there something else?

conejoninja:
The code looks good, but I'd lie if I said I fully understand what is happening :shrug: 
maybe ping @bgould

bgould:
I haven't tried this extensively but this change seems to reduce the fps of the pyportal_boing example slightly... not sure if it is specific to that program, or if LLVM is doing some magic like unrolling the loop itself more efficiently, or something else.  Do you have that hardware? Are you able to see a similar result?

ili9341/parallel_atsamd51.go
sago35: I thought that loop unrolling would make it faster.
But the pyportal_boing demo is somehow slower.
https://en.wikipedia.org/wiki/Loop_unrolling

sago35:
pyportal_boing I tried it.
I wasn't aware of it, but it certainly slows down.

* pyportal_boing demo
    * before : 42 fps
    * with 1 : 51 fps
    * with 1 and 3 : 42 fps

There may be some kind of loop optimization being done in LLVM.
Interestingly, when the first loop is set to +=4 instead of +=8, it is initially 51 fps, but gradually drops to around 44 fps.
I don't know why.

I'll check a bit more if I know how to see the final assembler code.

aykevl:
LLVM does a lot of optimizations (including loop unrolling and loop re-rolling) that may all interact with each other. I've sometimes managed to speed things up or slown things down by changes that should have had the opposite effect. In general, making the code simpler to understand usually also makes it easier to understand for the compiler and thus easier to optimize.

In my experience, one of the few things that reliably increases performance is avoiding I/O operations. Those have side effects unknown to the compiler so cannot be optimized by it. See for example https://github.com/tinygo-org/drivers/pull/89. Another that may help is to avoid branches (if/else), those are not always well optimized by LLVM.

Also, if you run performance sensitive code, remember that TinyGo allows tuning the optimization level. By default it uses `-opt=z` meaning optimize for code size at all cost. You can try `-opt=s` (optimize for size without impacting performance too much) and `-opt=2` (optimize for performance). This setting can have quite a significant effect on the performance of code, but also on code size.

sago35:
I understood.

I think it's better to leave 2 and 3 to LLVM optimization.
1 is exactly what reduces the I/O operations.

I think this PR could be completed by making it just 1.
I force-pushed the PR.


benchmarks: 
tinygo build -o app.uf2 -target pyportal -size short -opt=* ./examples/ili9341/pyportal_boing

* before (-opt=z) : 42 fps
* before (-opt=2) : 38 - 42 fps
* before (-opt=s) : 38 - 42 fps
* with1 (-opt=z) : 51 fps
* with 1 (-opt=2) : 47 - 51 fps
* with 1 (-opt=s) : 47 - 51 fps

bgould:
Yep I tried this out with the latest push and there is a noticeable improvement.  Nice work, that is a good speedup IMO.  I'm in favor of this change

deadprogram:
Finally got a chance to test this, and it looks great! :globe_with_meridians: 

PR #150, opened 2020-05-06
Fix DrawFastHLine for ST77xx, SSD1331 and SSD1351

droid42:
DrawFastHLine internally uses FillRectangle(x,y,width,height,c), so height must be 1 to draw a horizontal line

I detected it trying my new 1,44" 128x128 TFT LCD with ST7735 driver, but fixed it in the other drivers, too.

Thank you for your fine work.


aykevl:
This looks like a bug, yes. While `DrawFastHLine` works in x/y coordinates, `FillRectangle` works with x/y/width/height.
@conejoninja can you take a look to check whether this is fix is correct?

conejoninja:
Sorry for the late reply, thanks for your contribution. Merging.

PR #154, opened 2020-05-22
apa102: avoid creating garbage

aykevl:
Avoid creating unnecessary garbage in the following ways:

  * Use the `Transfer` method instead of the `Tx` method for single byte transfers.
  * Use a statically allocated buffer for the fixed start-of-frame sequence.

Running this for a few minutes did not cause the garbage collector to run. Previously, it would run every second or so in a persistence-of-vision application, which was clearly visible.

---

Context: I want to build something like [this thing](https://lighttoys.cz/store/vpoi/V4/), which is seriously awesome.

### DIFF
--- a/apa102/apa102.go
+++ b/apa102/apa102.go
@@ -19,6 +19,8 @@ const (
 	GRB
 )
 
+var startFrame = []byte{0x00, 0x00, 0x00, 0x00}
+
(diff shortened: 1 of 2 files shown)


deadprogram:
Looks good, now merging.

PR #156, opened 2020-06-04
WIP: Wifinina refactor

bgould:
This PR includes a refactoring of the wifinina driver to make the code more readable.  It attempts to encapsulate the details of the low-level protocol in the `wifinina.Buffer` struct.  Additionally, a new "Transport" interface is introduced to decouple the driver from the TinyGo `machine` package; this for instance makes it possible to the driver with a serial converter such as FTDI 232H from standard Go.  Support for reading/formatting BSSIDs and MAC addresses is also improved, and UDP support is added and confirmed to work.

Code still needs cleanup but the bulk of the refactor is completed.


deadprogram:
Hi @bgould could you please rebase against `dev` branch? The CI should pass if you do so. Thanks!

net/net.go
```diff
@@ -18,7 +18,7 @@ func DialUDP(network string, laddr, raddr *UDPAddr) (*UDPSerialConn, error) {
 	listenport := strconv.Itoa(laddr.Port)
 
 	// disconnect any old socket
-	ActiveDevice.DisconnectSocket()
+	//ActiveDevice.DisconnectSocket()
```
deadprogram, line 21: You probably want to remove this, not just comment it out.

wifinina/tcp.go
```diff
@@ -74,24 +75,31 @@ func (drv *Driver) connectSocket(addr, portStr string, mode uint8) error {
...
 	}
 
-	return ErrConnectionTimeout
+	return nil
 }
 
-func (drv *Driver) ConnectUDPSocket(addr, sport, lport string) error {
-	return ErrNotImplemented
+func (drv *Driver) ConnectUDPSocket(addr, portStr, lport string) error {
+	//println("addr", addr)
```
deadprogram, line 99: Remove?

deadprogram:
Looks like that last commit is failing the tests, @bgould

bgould:
> Looks like that last commit is failing the tests, @bgould

Sometimes `goimports` likes to write package names like `github.com/tinygo-org/tinygo/src/runtime` instead of just `runtime` in my setup.  Fixed this, CI seems to be passing now.  Going back through and cleaning up commented out code and adding comments where some should be added.  After that I think this should be ready to remove WIP and review

deadprogram:
@bgould looks like you are going to need to resolve some merge conflicts, along with whatever else remains to be done on this PR. Thanks!

PR #158, opened 2020-06-14
Added support for the Bosch BMP280 temperature and pressure sensor.

nerdmeister:


deadprogram:
Hi @nerdmeister thanks for the contribution. Can you please add this sensor to the smoke tests like this: https://github.com/tinygo-org/drivers/blob/dev/Makefile#L136-L137

Thank you!

conejoninja:
It should be added to the README.md 's list too : https://github.com/nerdmeister/drivers/blob/bmp280/README.md

nerdmeister:
As requested, BMP280 has been added to the readme.md and the smoke-test.

examples/bmp280/main.go
```diff
@@ -0,0 +1,50 @@
...
+		t, err := sensor.ReadTemperature()
+		if err != nil {
+			println("Error reading temperature")
+		}
+
+		print("Temperature: ")
+		println(t)
+
+		// Temperature in degrees Celsius
+		//fmt.Printf("Temperature: %.2f\n", float32(t)/100)
```
conejoninja, line 36: I'd remove the commented lines, also, ReadTemperature returns milli Celsiuis, so it should be /1000 instead of /100, or am I missing someting?

bmp280/bmp280.go
```diff
@@ -0,0 +1,240 @@
...
+	Filter      Filter
+}
+
+type calibrationCoefficients struct {
+	// Temperature compensation
+	t1 uint16
+	t2 int16
+	t3 int16
+
+	// Pressure compensation
```
conejoninja, line 38: what do you think about using an array instead? 
```go
p [8]uint16 
```

bmp280/bmp280.go
```diff
@@ -0,0 +1,240 @@
...
+	// Calculate t_fine (temperature), used for the Pressure compensation
+	var1 := ((rawTemp >> 3) - int32(d.cali.t1<<1)) * int32(d.cali.t2) >> 11
+	var2 := (((rawTemp >> 4) - int32(d.cali.t1)) * ((rawTemp >> 4) - int32(d.cali.t1)) >> 12) *
+		int32(d.cali.t3) >> 14
+
+	t_fine := var1 + var2
+
+	rawPres := convert3Bytes(data[0], data[1], data[2])
+
+	// Pressure compensation
```
conejoninja, line 169: it would be nice if you could point to the part of the datasheet where this calculations come from

nerdmeister: Has been added.

examples/bmp280/main.go
nerdmeister: Thank for catching this one, ReadTemperature() and ReadPressure() now return in milli degrees celsius and milli pascal.

And the example has been adjusted to now properly show the temp in degrees celsius and pressure in hecto pascal.

bmp280/bmp280.go
nerdmeister: I would prefer to not change this, since the compensation formula from the datasheet is already hard to read and I don't want to add another layer of confusion by changing everything from say t1 to t[0] and p3 to p[2].

deadprogram:
Thank you very much for this contribution @nerdmeister now squash/merging. Thanks again!

PR #159, opened 2020-06-19
wifinina: Support UDP

bgould:
This PR adds UDP support for the wifinina driver.  Originally I was going to include this in https://github.com/tinygo-org/drivers/pull/156 but because of bug unrelated to UDP, refactoring will require more work.  In the meantime, UDP support can be added via this set of changes.


deadprogram:
Great work @bgould thank you very much for this, I need it! Now merging.

PR #160, opened 2020-06-25
LSM303AGR

alankrantas:
Works pretty well for me.  And since this is primary created for micro:bit v1.5, I included the math package to added some functions to provide data like pitch, roll, compass heading (the "compass" is not so accurate though, even in MakeCode) as well as the built-in temperature sensor.

### DIFF
--- a/examples/lsm303agr/main.go
+++ b/examples/lsm303agr/main.go
@@ -1,0 +1,47 @@
+/*
+Example code for package lsm303agr, which implements a driver for the LSM303AGR, 
+ a 3 axis accelerometer/magnetic sensor included onblard BBC micro:bits v1.5.
+
+Datasheet: https://www.st.com/resource/en/datasheet/lsm303agr.pdf
+micro:bit versions: https://tech.microbit.org/hardware/i2c/
+*/
+
+package main
+
+import (
+    "machine"
+    "time"
+    "tinygo.org/x/drivers/lsm303agr"
+)
+
+func main() {
+       
+    machine.I2C0.Configure(machine.I2CConfig{})
+    accel_mag := lsm303agr.New(machine.I2C0)
+    
+    if !accel_mag.Connected() {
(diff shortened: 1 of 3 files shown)


conejoninja:
I remember @aykevl had some functions for pitch/roll somewhere too, I'm unable to find it now. I think we could/should create accelerometer/magnometer/compasser interfaces and a package with those common functions. What do you think?

alankrantas:
@conejoninja probably be a good idea when there are more similar sensors added.

There are also formulas to calculate gyroscope motions, but I never understand them.

(And anyone knows what's the deal with the failing CI format check?)

deadprogram:
Hi @alankrantas you can run `go fmt ./examples/lsm303agr/main.go` which should resolve that, then please commit and push the updated version. Thanks!

alankrantas:
Tried to format it and upload the file directly. Still not working.

(It's dragon boat festival today, and I probably need to throw a few rice cakes to feed this CI monster.)

deadprogram:
Your last commit got past that, how it is a different error. Please rebase your branch against the current `dev` branch and push again.

alankrantas:
Ok...I have no idea how to do rebase online. Should I make a new PR? Or make a new branch at my fork?

deadprogram:
The easiest thing would be create a new branch based on `dev`, then add your new changes on top of that.

Also, while you are at it, please add a smoke test like this: https://github.com/tinygo-org/drivers/blob/dev/Makefile#L142-L143

Thank you.

aykevl:
Yes, I have a package here: https://github.com/aykevl/fusion
The main issue with it is that it doesn't expose yaw/pitch/roll but that could be added. So far I've mainly (tried to) use it for things that do not have an obvious up/down side so the yaw/pitch/roll format didn't make a lot of sense. Improvements are welcome.

PR #162, opened 2020-06-25
lsm303agr

alankrantas:
For LSM303AGR accelerometer/magnetic sensor on micro:bit v1.5.

### DIFF
--- a/lsm303agr/lsm303agr.go
+++ b/lsm303agr/lsm303agr.go
@@ -103,7 +103,7 @@ func (d *Device) Configure(cfg Configuration) {
 
 }
 
-/* read raw acceleration data (in ug/microgram) from all axis */
+/* read raw acceleration data (in ug or microgram) from all axis */
 func (d *Device) ReadAcceleration() (x int32, y int32, z int32) {
 
 	data1, data2, data3, data4, data5, data6 := []byte{0}, []byte{0}, []byte{0}, []byte{0}, []byte{0}, []byte{0}
@@ -126,19 +126,19 @@ func (d *Device) ReadAcceleration() (x int32, y int32, z int32) {
 		range_factor = 12 // the readings in 16G are a bit off
 	}
 
-	x = int32(int16((uint16(data1[0])<<8 | uint16(data2[0]))) >> 4 * range_factor) * 1000
-	y = int32(int16((uint16(data3[0])<<8 | uint16(data4[0]))) >> 4 * range_factor) * 1000
-	z = int32(int16((uint16(data5[0])<<8 | uint16(data6[0]))) >> 4 * range_factor) * 1000
+	x = int32(int16((uint16(data1[0])<<8|uint16(data2[0])))>>4*range_factor) * 1000
+	y = int32(int16((uint16(data3[0])<<8|uint16(data4[0])))>>4*range_factor) * 1000
+	z = int32(int16((uint16(data5[0])<<8|uint16(data6[0])))>>4*range_factor) * 1000
 	return
 }
 
-/* read pitch/roll degrees */
+/* read pitch and roll angles (in micro-degrees) */
 func (d *Device) ReadPitchRoll() (pitch int32, roll int32) {
 
 	x, y, z := d.ReadAcceleration()
 	xf, yf, zf := float64(x), float64(y), float64(z)
-	pitch = int32(math.Round(math.Atan2(yf, math.Sqrt(math.Pow(xf, 2)+math.Pow(zf, 2)))*(180/math.Pi)*100) / 100)
-	roll = int32(math.Round(math.Atan2(xf, math.Sqrt(math.Pow(yf, 2)+math.Pow(zf, 2)))*(180/math.Pi)*100) / 100)
+	pitch = int32((math.Round(math.Atan2(yf, math.Sqrt(math.Pow(xf, 2)+math.Pow(zf, 2)))*(180/math.Pi)*100) / 100) * 1000000)
+	roll = int32((math.Round(math.Atan2(xf, math.Sqrt(math.Pow(yf, 2)+math.Pow(zf, 2)))*(180/math.Pi)*100) / 100) * 1000000)
 	return
 
 }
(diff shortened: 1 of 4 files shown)


lsm303agr/lsm303agr.go
```diff
@@ -0,0 +1,196 @@
...
+	cmd[0] = byte(0xC0)
+	d.bus.WriteRegister(uint8(d.AccelAddress), TEMP_CFG_REG_A, cmd)
+
+	cmd[0] = byte(0x80 | d.MagPowerMode<<4 | d.MagDataRate<<2 | d.MagSystemMode)
+	d.bus.WriteRegister(uint8(d.MagAddress), MAG_MR_REG_M, cmd)
+
+}
+
+/* read raw acceleration data from all axis, 1000+ = 1g */
+func (d *Device) ReadAcceleration() (x int16, y int16, z int16) {
```
deadprogram, line 107: I think it would be a lot better to support the same API as the other devices in this category. For example, this function would be better defined as `func (d *Device) ReadAcceleration() (x int32, y int32, z int32)` and return values in µg (micro-gravity).

lsm303agr/lsm303agr.go
```diff
@@ -0,0 +1,196 @@
...
+	xf, yf, zf := float64(x), float64(y), float64(z)
+
+	pitch = float32(math.Round(math.Atan2(yf, math.Sqrt(math.Pow(xf, 2)+math.Pow(zf, 2)))*(180/math.Pi)*100) / 100)
+	roll = float32(math.Round(math.Atan2(xf, math.Sqrt(math.Pow(yf, 2)+math.Pow(zf, 2)))*(180/math.Pi)*100) / 100)
+	return
+
+}
+
+/* read magnetic field level from all axis */
+func (d *Device) ReadMagneticField() (x int16, y int16, z int16) {
```
deadprogram, line 148: In similar fashion to my previous comment, this function would be `func (d *Device) ReadRotation() (x int32, y int32, z int32)` and return values in µ°/s (micro-degrees/sec).

lsm303agr/lsm303agr.go
```diff
@@ -0,0 +1,196 @@
...
+	data1, data2 := []byte{0}, []byte{0}
+	d.bus.ReadRegister(uint8(d.AccelAddress), OUT_TEMP_H_A, data1)
+	d.bus.ReadRegister(uint8(d.AccelAddress), OUT_TEMP_L_A, data2)
+
+	t = int16((uint16(data1[0])<<8 | uint16(data2[0]))) >> 4
+	return
+}
+
+/* read temperature in Celsius */
+func (d *Device) ReadTemperature() (c float32) {
```
deadprogram, line 191: Same comment as before, but matching https://github.com/tinygo-org/drivers/blob/master/lsm6ds3/lsm6ds3.go#L169

lsm303agr/lsm303agr.go
alankrantas: I think currently it returns mg (if I lay the board flat on the table, the z axis reads a bit more than -1000, which is the same result I saw in MakeCode.) So times 1000 should be good?

lsm303agr/lsm303agr.go
alankrantas: I think it returns microgauss. In MakeCode it's microtesla.

lsm303agr/lsm303agr.go
aykevl: Please check this with the datasheet. Usually it is a power of two, so perhaps it's 1024. The datasheet will tell you exactly. Also, many of these chips can be configured to different sensitivities (for example, 2g, 4g, 8g, 16g for different applications), although initially using the default would be fine.

lsm303agr/lsm303agr.go
aykevl: One purpose of these APIs is to give a consistent interface for different drivers. It is not a consistent interface if every driver has a different scale (even if the signature is the same). There are many benefits to a consistent interface, such as that it is easier to learn. Therefore, I would suggest matching the existing drivers for consistency.

lsm303agr/lsm303agr.go
alankrantas: I think the [datasheet](https://www.st.com/resource/en/datasheet/lsm303agr.pdf) is poorly written; there's a lot of info it doesn't revel in there. 

In my experiments each axis always returns a value between +- 32768. The bigger the g sensitivity, the bigger the value scale would be (the observed Earth gravity reading become smaller in scale). I guess this is how it can represent a larger value in the same 2 bytes.

(And in normal or low power mode, only top 12 bits of the 2 bytes would be used, can't find the source now. I didn't implement the full 16-bit high resolution mode.)

The readings in 2g is exactly the same as what I've seen measured in MakeCode (1000~plus). In [MakeCode](https://makecode.microbit.org/reference/input/acceleration) the unit of raw axis data is labeled as mg or milligram. 1,000 mg = 1g (Earth gravity).

lsm303agr/lsm303agr.go
alankrantas: Oh, I misread the previous post. I thought it was about ReadMagneticField(). It's ReadRotation().

Will try to fix it.

alankrantas: Note: I can convert it to microdegree since the magnetic sensor is not a gyroscope. It sense the "direction" by Earth's magnetic field, not the rotating motion.

Or should I name it ReadCompass()?

lsm303agr/lsm303agr.go
```diff
@@ -0,0 +1,195 @@
+/*
```
deadprogram, line 1: Could you please make all of the comments follow the standard Go style, that also makes them work with GoDoc?

https://blog.golang.org/godoc

For example, the following comment block should be exactly like this instead:

```
// package lsm303agr implements a driver for the LSM303AGR,
// a 3 axis accelerometer/magnetic sensor which is included on BBC micro:bits v1.5.
//
// Datasheet: https://www.st.com/resource/en/datasheet/lsm303agr.pdf
```

deadprogram:
Thank you very much for all of the many changes requested @alankrantas and of course for adding this device!

I am going to squash/merge now. Thanks again!

PR #163, opened 2020-07-01
tmp102: add Connected func to check for device

yannishuber:
This PR adds a `Connected()` function for the tmp102 driver. This can be useful to test if the device is properly connected. For example in TinyHCI.


deadprogram:
Thanks for the addition, @yannishuber now merging.

PR #168, opened 2020-07-13
Fix WS2812 build tag to include Arduino Nano

alankrantas:
Using the tag ```arduino-nano``` doesn't seem to work. And as far as I know anything made with Atmega328p are either Uno or Nano clones.


deadprogram:
Hello @alankrantas can you please make sure that this PR is using the latest `dev` branch? From the CircleCI error it appears to not be.

Thank you!

PR #169, opened 2020-07-13
Fix WS2812 build for Nano

alankrantas:


deadprogram:
Thanks @alankrantas now merging.

aykevl:
This might be an issue for AVR boards that don't run at 16MHz but there aren't that many and we can fix it when needed. For now this seems like the better solution.

PR #171, opened 2020-07-13
ili9341: Cache address window to prevent sending unnecessary commands

ardnew:
This is a port of an optimization recently added to the Adafruit_ILI9341 and ILI9341_t3 Arduino drivers.

It works by caching the last row/column sent to the setWindow function, and only resending the respective commands when one of the drawing axes changes. It exhibits a surprisingly effective speedup for basically all drawing operations.

Retrying PR against dev branch instead of master

### DIFF
--- a/ili9341/ili9341.go
+++ b/ili9341/ili9341.go
@@ -19,6 +19,9 @@ type Device struct {
 	rotation Rotation
 	driver   driver
 
+	x0, x1 int16 // cached address window; prevents useless/expensive
+	y0, y1 int16 // syscalls to PASET, CASET, and RAMWR
+
 	dc  machine.Pin
 	cs  machine.Pin
 	rst machine.Pin
@@ -36,6 +39,10 @@ func (d *Device) Configure(config Config) {
 	d.width = config.Width
 	d.height = config.Height
 	d.rotation = config.Rotation
+
+	// try to pick an initial cache miss for one of the points
+	d.x0, d.x1 = -(d.width + 1), d.x0
+	d.y0, d.y1 = -(d.height + 1), d.y0
 
 	output := machine.PinConfig{machine.PinOutput}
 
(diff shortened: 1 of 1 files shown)


sago35:
It will be faster in normal use.
With pyportal_boing, the cache-hit rate is low, so it was not faster.
But it's not slow, so it's good.
I'd like to test it some more.


* pyportal_boing (WioTerminal (atsamd51p19 + SPI))
    * before : 40 - 42 fps
    * after : 40 - 42 fps


test code:

```go
// setWindow prepares the screen to be modified at a given rectangle
func (d *Device) setWindow(x, y, w, h int16) {
	//x += d.columnOffset
	//y += d.rowOffset
	wr := false
	x1 := x + w - 1
	if x != d.x0 || x1 != d.x1 {
		wr = true
		d.sendCommand(CASET, []uint8{
			uint8(x >> 8), uint8(x), uint8(x1 >> 8), uint8(x1),
		})
		d.x0, d.x1 = x, x1
	}
	y1 := y + h - 1
	if y != d.y0 || y1 != d.y1 {
		wr = true
		d.sendCommand(PASET, []uint8{
			uint8(y >> 8), uint8(y), uint8(y1 >> 8), uint8(y1),
		})
		d.y0, d.y1 = y, y1
	}
	if wr {
		d.sendCommand(RAMWR, nil)
		CacheFail++ // !!!
	} else {
		CacheHit++ // !!!
	}
}
```

result:

```
41  fps
hit 25 / fail 21207
```

ardnew:
Accidentally lost the hyperlinks when I copied the message from the bad PR I closed. So just for reference, the origin of the change is here:

https://github.com/PaulStoffregen/ILI9341_t3/pull/59

https://github.com/adafruit/Adafruit_ILI9341/pull/64

sago35:
LGTM

sago35:
@deadprogram 
Please review.

deadprogram:
Thanks for the contribution @ardnew and for looking it over @sago35 

Now squash/merging.

PR #172, opened 2020-07-15
all: changeover to eliminate all direct use of master/slave terminology

deadprogram:
This PR is to changeover to eliminate all direct use of master/slave terminology.

It is intended to match https://github.com/tinygo-org/tinygo/pull/1216 in the main TinyGo repo. Once that PR has been merged, then the tests for this PR will pass again.

See https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names

We will also need to change the master branch name to release, and change the default branch in Github.

### DIFF
--- a/wifinina/wifinina.go
+++ b/wifinina/wifinina.go
@@ -134,23 +134,23 @@ const (
 	CmdSetDigitalWrite = 0x51
 	CmdSetAnalogWrite  = 0x52
 
-	ErrTimeoutSlaveReady  Error = 0x01
-	ErrTimeoutSlaveSelect Error = 0x02
-	ErrCheckStartCmd      Error = 0x03
-	ErrWaitRsp            Error = 0x04
-	ErrUnexpectedLength   Error = 0xE0
-	ErrNoParamsReturned   Error = 0xE1
-	ErrIncorrectSentinel  Error = 0xE2
-	ErrCmdErrorReceived   Error = 0xEF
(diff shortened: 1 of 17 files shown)


deadprogram:
Now that  tinygo-org/tinygo#1216 has been merged into the `dev` branch and the tinygo/tinygo-dev docker container has these changes, the CircleCI tests in the drivers repo now pass.

This is now ready to review/merge.

wifinina/wifinina.go
```diff
@@ -296,8 +296,8 @@ func (d *Device) StartClient(addr uint32, port uint16, sock uint8, mode uint8) e
 		println("[StartClient] called StartClient()\r")
 		fmt.Printf("[StartClient] addr: % 02X, port: %d, sock: %d\r\n", addr, port, sock)
 	}
-	if err := d.waitForSlaveSelect(); err != nil {
-		d.spiSlaveDeselect()
+	if err := d.waitForDeviceSelect(); err != nil {
+		d.spiDeviceDeselect()
```
aykevl, line 300: The new name, according to https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names, is "chip select". In fact, "chip select" (or CS) seems to be quite common already. I would suggest changing to that instead of "device select".

deadprogram: Done.

deadprogram:
Now merging, thanks for the review @aykevl

PR #173, opened 2020-07-17
Add Hd44780i2c driver

alankrantas:
This is for the I2C version of HD44780 LCDs.

The driver is based on some C++/MicroPython drivers and I did try to match the existing HD44780 driver, however the control methods are a bit different (for example, there is no buffer; everything takes effect immediately).

Tested on Arduino Uno/Nano 33 IoT, micro:bit and Adafruit M4 Metro (although the Metro didn't like me to draw 5V power from its pin; only my 3.3V copy worked on it) with both 16x2 and 20x4 versions.


deadprogram:
Hi @alankrantas thank you very much for this contribution. Now squashing/merging.

PR #176, opened 2020-07-30
mcp23017: new driver for MCP23017 (I2C port expander)

rogpeppe:


deadprogram:
Hello @rogpeppe I think there are some pretty great ideas in this PR that could be incorporated into the overall design of drivers. But as a result, I require a bit more time to study what you are doing.

More soon.

rogpeppe:
Thanks for the update. Beyond the use of I2C as an interface (and
associated tests), I'm not sure if there's likely to be much that's
cross-applicable, but I look forward to your assessment :)

sago35:
An example of multidevice might be a good place to start.
It might be nice to have Toggle(), as in samd51.
https://github.com/tinygo-org/tinygo/blob/release/src/machine/machine_atsamd51.go#L542

Overall, this PR is great.

rogpeppe:
@sago35 Thanks for the feedback!
I added Toggle and friends, and also added a MultiDevice example.

go.mod
```diff
@@ -2,4 +2,7 @@ module tinygo.org/x/drivers
 
 go 1.14
 
-require github.com/eclipse/paho.mqtt.golang v1.2.0
+require (
+	github.com/eclipse/paho.mqtt.golang v1.2.0
+	github.com/frankban/quicktest v1.10.0
```
sago35: It adds a dependency for testing, but I can't decide if it's OK or not.
What do you think? @deadprogram @aykevl

sago35:
I don't have mcp23017, but it was easy to understand because of the test code.

go.mod
rogpeppe: I also wasn't sure if that would be OK either. I could just remove the test file and then it would be the same as all the other drivers, I guess :grin:

aykevl: I'm personally fine with extra dependencies, although I haven't looked into this particular one.

deadprogram:
Now that I merged in the `tester` package changes, @rogpeppe can you please revamp this PR so it uses that? Thanks again for getting this process of improved testing started!

Chillance:
So, this is nice because I'm currently using PCAL6416A myself (https://www.nxp.com/docs/en/data-sheet/PCAL6416A.pdf).
Question is though, should parts of code here be moved to a more generic one to more easier support similar devices?

jkaflik:
Hey folks,

what’s the status of this? Any blockers?

jkaflik:
Hi @rogpeppe !
Any chance to have it rebased soon? Would be good to have it in master :) thanks!

rogpeppe:
@jkaflik done (thanks for the ping and sorry for the delay - i haven't been using tinygo recently).
I made some changes to the `tester` package too.

deadprogram:
Thanks for updating this PR @rogpeppe took me a little while to digest the changes you made to `tester` but in the end, seemed like a good idea.

Now merging, thank you very much for the contribution!

jkaflik:
Thank you @rogpeppe @deadprogram ! :)

PR #177, opened 2020-07-30
bmi160: add initial support

aykevl:
Add initial support for the BMI160, which is present on the [GT832C_01](https://www.aliexpress.com/item/4000123520442.html). I haven't actually tested the example because I don't have a separate BMI160, but it does compile. However, this driver works in a project I'm working on.


deadprogram:
Looks good to me, thanks @aykevl now merging.

PR #178, opened 2020-07-31
Add TM1637 driver

alankrantas:
This is for 4-digit version of the TM1637 7-segment number display, which can sort of display texts too. There are 6-digit versions but they are not that common (at least I've never seen them in the shops).

Tested on Arduino Uno/Nano/Nano33 IoT, BBC micro:bit and Adafruit Metro M4. In fact it can also run on the Digispark if you keep the script simple.


deadprogram:
Sorry it was so long to review this contribution. Somehow it was overlooked.

I resolved the merge conflict and squashed the commits. It has now been merged into `dev`.

Thank you very much!

PR #180, opened 2020-08-06
Fix setWindow bug, add CS pin for Clue compatibility.

jmstriegel:
- corrected bit shift in setWindow that broke high addresses
- added csPin to constructor (will now work with adafruit clue)
- small adjustments and comments to init routine based on working arduino driver

### DIFF
--- a/st7789/st7789.go
+++ b/st7789/st7789.go
@@ -253,22 +253,21 @@ func (d *Device) SetRotation(rotation Rotation) {
 	madctl := uint8(0)
 	switch rotation % 4 {
 	case 0:
-		madctl = MADCTL_MX | MADCTL_MY | MADCTL_RGB
+		madctl = MADCTL_MX | MADCTL_MY
 		d.rowOffset = d.rowOffsetCfg
 		d.columnOffset = d.columnOffsetCfg
 		break
 	case 1:
-		madctl = MADCTL_MY | MADCTL_MV | MADCTL_RGB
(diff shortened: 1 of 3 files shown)


st7789/st7789.go
```diff
@@ -224,21 +253,22 @@ func (d *Device) SetRotation(rotation Rotation) {
 	madctl := uint8(0)
 	switch rotation % 4 {
 	case 0:
-		madctl = MADCTL_MX | MADCTL_MY
+		madctl = MADCTL_MX | MADCTL_MY | MADCTL_RGB
```
conejoninja, line 256: are you sure MADCTL_RGB (0x00) is required here?

jmstriegel: Yeah, not required at all. It doesn't affect the data byte, but I thought it was more clear that MADCTL has an effect on RGB/BGR memory ordering in addition to rotation.

jmstriegel: Latest commit removes this change.

jmstriegel:
Hi @conejoninja - checking to see if this patch needed anything else to merge. I'm making good progress on a further update that adds basic support for vsync timing. I'm new here, so please let me know if there's anything I should be doing differently when contributing updates.

conejoninja:
Merged! Great work, thanks for your contribution. 

Just one small detail, for future PR, please do them against `dev` branch, it's better if you use it as a base branch too for your changes.

jmstriegel:
Thanks! Will do.

PR #187, opened 2020-08-20
lis2mdl: add LIS2MDL magnetometer

deadprogram:
This PR move the functionality for the LIS2MDL magnetometer into a separate driver for clarity. A future PR will remove the duplication from the current LSM303AGR driver. Thanks @alankrantas for having done most of the work on this already.


conejoninja:
Looks good, merging... Thanks both of you for the great job.

PR #191, opened 2020-08-26
test: add tester package for mocking of I2C bus/devices

deadprogram:
This PR adds a `tester` package that contains I2CBus and I2CDevice mocks for use in writing unit tests for I2C devices.

Making this PR as draft, it is intended to serve as a starting point for conversation.

### DIFF
--- a/tester/device.go
+++ b/tester/device.go
@@ -1,0 +1,58 @@
+package tester
+
+import (
+	qt "github.com/frankban/quicktest"
+)
+
+const maxRegisters = 200
+
+// I2CDevice represents a mock I2C device on a mock I2C bus.
+type I2CDevice struct {
+	C    *qt.C
+	Addr uint8
+	// Registers holds the device registers. It can be inspected
+	// or changed as desired for testing.
+	Registers [maxRegisters]uint8
+	// If Err is non-nil, it will be returned as the error from the
+	// I2C methods.
+	Err error
+}
+
+// NewI2CDevice returns a new mock I2C device.
+func NewI2CDevice(c *qt.C, addr uint8) *I2CDevice {
+	return &I2CDevice{
+		C:    c,
+		Addr: addr,
+	}
+}
+
+// ReadRegister implements I2C.ReadRegister.
+func (d *I2CDevice) ReadRegister(r uint8, buf []byte) error {
+	if d.Err != nil {
+		return d.Err
+	}
+	d.AssertRegisterRange(r, buf)
+	copy(buf, d.Registers[r:])
+	return nil
+}
+
+// WriteRegister implements I2C.WriteRegister.
+func (d *I2CDevice) WriteRegister(r uint8, buf []byte) error {
+	if d.Err != nil {
+		return d.Err
+	}
+	d.AssertRegisterRange(r, buf)
+	copy(d.Registers[r:], buf)
+	return nil
+}
+
+// AssertRegisterRange asserts that reading or writing the given
+// register and subsequent registers is in range of the available registers.
+func (d *I2CDevice) AssertRegisterRange(r uint8, buf []byte) {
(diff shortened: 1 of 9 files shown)


tester/device.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	qt "github.com/frankban/quicktest"
+)
+
+const maxRegisters = 200
+
+// I2CDevice represents a mock I2C device on a mock I2C bus.
+type I2CDevice struct {
+	C    *qt.C
```
rogpeppe, line 11: I think I'd use an interface type rather than using *qt.C directly.

```
// Fataler is used by the I2CDevice type to abort when it's used in
// unexpected ways, such as reading an out-of-range register.
type Fataler interface {
    // Fatalf prints the Printf-formatted message and exits the current
    // goroutine.
    Fatalf(f string, a ...interface{})
}
```

Also, I don't think I'd bother exporting it - it's not a significant part of the user-visible API and the user should have access to it anyway.

Unfortunately I couldn't think of a better name for the interface :)

tester/device.go
```diff
@@ -0,0 +1,58 @@
...
+	Addr uint8
```
rogpeppe, line 12: I think I'd leave this unexported, and maybe define an exported `Addr` method instead.

tester/i2c.go
```diff
@@ -0,0 +1,52 @@
+package tester
+
+import (
+	qt "github.com/frankban/quicktest"
+)
+
+// I2CBus implements the I2C interface in memory for testing.
+type I2CBus struct {
+	C       *qt.C
```
rogpeppe, line 9: Again, I'd use an interface type and unexport this.

tester/i2c.go
```diff
@@ -0,0 +1,52 @@
...
+	Devices []*I2CDevice
```
rogpeppe, line 10: I don't think I'd export this either.

tester/i2c.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Tx implements I2C.Tx.
+func (bus *I2CBus) Tx(addr uint16, input, output []byte) error {
+	// TODO: implement this
+	return nil
+}
+
+// FindDevice returns the device with the given address.
+func (bus *I2CBus) FindDevice(addr uint8) *I2CDevice {
```
rogpeppe, line 44: Perhaps this would also be best left unexported.

rogpeppe (approved):
This should be really useful - nice to see it migrating :)

tester/device.go
```diff
@@ -0,0 +1,58 @@
+package tester
+
+import (
+	qt "github.com/frankban/quicktest"
+)
+
+const maxRegisters = 200
```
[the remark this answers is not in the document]
deadprogram: Done.

tester/device.go
deadprogram: I chose "Failer".

tester/device.go
deadprogram: Done.

tester/i2c.go
deadprogram: Done.

tester/i2c.go
deadprogram: Done.

tester/i2c.go
deadprogram: Not sure on this one, I left it for now.

deadprogram:
I think this is now ready for merge. 

Once merged, the next step will be switching over all of the current I2C devices to use the interface, and have a "sanity" test using `Connected()`. Then individual drivers can have further tests added as time permits.

sago35:
I think it's very good.

deadprogram:
OK, now merging. Thanks for the feedback, everyone!

Issue #151, opened 2020-05-16
SPI not declared by package machine


ArtemkaKun:
Hi,
I have an SSD1306 OLED 128x32 display. When I try to use an example for SSD1306 driver and build, have 2 errors:
`../../../../root/go/src/tinygo.org/x/drivers/ssd1306/ssd1306.go:38:19: SPI not declared by package machine
../../../../root/go/src/tinygo.org/x/drivers/ssd1306/ssd1306.go:63:25: SPI not declared by package machine
`
Any thoughts on what can be wrong? SPI type actually declared in machine package :/

conejoninja:
Hello, which command are you using to build? Looks like the target is missing.

ArtemkaKun:
> Hello, which command are you using to build? Looks like the target is missing.

Hi. I have an Arduino Nano, so I use `tinygo flash -target=arduino-nano  -port /dev/ttyUSB0 main.go`
(main.go it's just a simple go file that contains code from SSD1306 example)

conejoninja:
Unfortunately, SPI support is missing for the Nano :( 

https://tinygo.org/microcontrollers/arduino-nano/

ArtemkaKun:
> Unfortunately, SPI support is missing for the Nano :(
> 
> https://tinygo.org/microcontrollers/arduino-nano/

Oh, sorry for my incompetence. I try to connect to my display with I2C.
Here is the code
`func main() {
	machine.I2C0.Configure(machine.I2CConfig{
		Frequency: machine.TWI_FREQ_400KHZ,
	})

	display := ssd1306.NewI2C(machine.I2C0)
	display.Configure(ssd1306.Config{
		Address: ssd1306.Address_128_32,
		Width:   128,
		Height:  32,
	})

	display.ClearDisplay()

	x := int16(0)
	y := int16(0)
	deltaX := int16(1)
	deltaY := int16(1)
	for {
		pixel := display.GetPixel(x, y)
		c := color.RGBA{255, 255, 255, 255}
		if pixel {
			c = color.RGBA{0, 0, 0, 255}
		}
		display.SetPixel(x, y, c)
		display.Display()

		x += deltaX
		y += deltaY

		if x == 0 || x == 127 {
			deltaX = -deltaX
		}

		if y == 0 || y == 31 {
			deltaY = -deltaY
		}
		time.Sleep(1 * time.Millisecond)
	}`

conejoninja:
[...]

PR #193, opened 2020-08-27
i2c: switch all i2c drivers definitions to use i2c bus interface type

deadprogram:
This PR switches all i2c drivers definitions to use i2c bus interface type instead of machine package concrete type.

Once it is merged into `dev`, then individual drivers can have tests added as time permits.

### DIFF
--- a/adt7410/adt7410.go
+++ b/adt7410/adt7410.go
@@ -1,8 +1,9 @@
 package adt7410 // import "tinygo.org/x/drivers/adt7410"
 
 import (
-	"machine"
 	"time"
+
+	"tinygo.org/x/drivers"
 )
 
 type Error uint8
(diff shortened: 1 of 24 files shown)


deadprogram:
I am working in branch https://github.com/tinygo-org/drivers/tree/adt7410-tests-improvements on improvements to the ADT7410 based on this branch.

aykevl:
I checked code size differences using `make smoke-test` and there are indeed some differences. Often in the 20-100 byte range, but sometimes larger. The largest I've seen was somewhere in the 500 byte range (although there is also a _decrease_).

I think this is worth the cost, especially considering the testability improvements and increased flexibility. Interfaces will allow the use of software I2C for any driver.

aykevl (approved):

deadprogram:
Thanks for the review, now merging.

conejoninja:
Now that this has been merged, should we do something similar with SPI? That would be helpful with issue #151 were a driver implements both I²C and SPI, but the board only implements one of those and fails to compile.

deadprogram:
Yes, that is for sure the next step.

PR #153, opened 2020-05-20
ILI9341 TFT driver (SPI)

sago35:
This PR adds SPI support for the ILI9341.

PR #196, opened 2020-09-02
ILI9341: add SPI driver for STM32F4

ardnew:
This PR adds an SPI-based ILI9341 driver for STM32F4 targets such as TinyGo boards `stm32f4disco` and `feather-stm32f405`.

Also added is the `pyportal_boing` example configured for the `feather-stm32f405`.

Note the `feather-stm32f405` board is still a WIP in the dev branch of tinygo. Adding this PR as a draft until that board and its SPI implementation are merged to make sure I'm submitting this PR correctly


ili9341/spi_stm32f4.go
```diff
@@ -0,0 +1,142 @@
...
+import (
+	"device/stm32"
+	"machine"
+)
+
+type spiDriver struct {
+	bus machine.SPI
+}
+
+func NewSpi(bus machine.SPI, dc, cs, rst machine.Pin) *Device {
```
aykevl, line 14: The Go convention is to keep abbreviations capitalized.

```suggestion
func NewSPI(bus machine.SPI, dc, cs, rst machine.Pin) *Device {
```

aykevl (commented):
What does this code do? Is it using a standard SPI connection or is it doing something special?

ardnew:
> What does this code do? Is it using a standard SPI connection or is it doing something special?

Which code are you referring to? If you're asking about all of it, then yes it's just using a standard SPI connection, along with the STM32F4-specific peripheral interface registers.

Similar drivers exist for SPI on SAMD21 and SAMD51, which this was largely inspired from. See:

https://github.com/tinygo-org/drivers/blob/dev/ili9341/spi_atsamd21.go

https://github.com/tinygo-org/drivers/blob/dev/ili9341/spi_atsamd51.go

ili9341/spi_stm32f4.go
ardnew: This change will be inconsistent with the other ILI9341 SPI drivers (see my previous comment). Is that OK? Or should we update those as well?

Amblyopius: All of them have to be changed or you can't use code across multiple devices with ILI9341 as in some cases you'd have to call NewSPI and in others NewSpi. Note that changing it also breaks any existing code using ILI9341 package (do we care at the moment?).

Note that current ILI9341 drivers are limited as they are write-only which means you have a 76800 pixels buffer that you can't read even though you might want to. Keeping a copy of it in memory is 150KB (assuming 16bit color) which for most devices is most of your RAM so rarely an option. The ATSAMD51 write functions also seem to have some unexplained fixation on playing around with SPI CTRLB register while neither ATSAMD51 nor ILI9341 manual seem to imply any reason to do so. I plan to experiment a bit by:
- Changing the ILI9341 SPI writes for ATSAMD51 to check if my understanding of ATSAMD51 SPI and the effect on ILI9341 is correct
- Adding ILI9341 SPI reads for ATSAMD51 and adding some of the read commands to ILI9341

bgould: My opinion is that `NewSPI` as per Ayke's suggestion is a better name more consistent with other drivers as well.  The other implementation for atsamd21 and atsamd51 should also be changed to keep things consistent, which will be a breaking change but I don't think that will be a big deal since the code is relatively newly released and the project is not at 1.0 yet.  The wioterminal examples will need to be updated to the new name if it is changed.  I'll go ahead and open a PR to update the names for the examples and the samd implementations.

ardnew: Updated the function names included in this PR (5f5f8a9), and the other drivers are being updated in PR #199 by @bgould. Marking this issue resolved.

sago35:
@ardnew 
I would like you to make a PR for the SPI branch of stm32f405.
I have a feather-stm32f405 and SPI-ILI9341 and I'm going to review this PR.

ardnew:
@sago35 just added the PR here: https://github.com/tinygo-org/tinygo/pull/1377

ili9341/spi_stm32f4.go
```diff
@@ -0,0 +1,142 @@
...
+func (pd *spiDriver) write8(b byte) {
+	if !pd.bus.Bus.CR1.HasBits(stm32.SPI_CR1_SPE) {
+		pd.bus.Bus.CR1.SetBits(stm32.SPI_CR1_SPE)
+	}
+
+	pd.setWord(b, true, true)
+
+	pd.bus.Bus.CR1.ClearBits(stm32.SPI_CR1_SPE)
+}
+
```
sago35, line 38: The processing for `SPI_CR1_SPE` does not seem to be necessary, and removing it will increase the speed from 68 fps to 88 fps.
Other functions are the same.

Any reason to use `SPI_CR1_SPE` ?

```suggestion
func (pd *spiDriver) write8(b byte) {
	pd.setWord(b, true, true)
}
```

sago35:
Please add the following examples as well.
I've confirmed that it works in my environment.

* examples/ili9341/basic
* examples/ili9341/scroll

ili9341/spi_stm32f4.go
```diff
@@ -0,0 +1,142 @@
...
+// puts a single 8-bit word in the SPI data register (DR).
+// if first (first word being transmitted) is false, waits for the SPI transmit
+// buffer empty bit (TXE) is set before putting the word in DR.
+// if last (last word being transmitted) is true, waits for the SPI transmit
+// buffer empty bit (TXE) is set and SPI bus busy bit (BSY) is clear before
+// returning.
+// for all wait operations, a fixed number of wait iterations (const tryMax) are
+// performed before a timeout is assumed.
+// if timeout occurs, returns false. otherwise, returns true.
+func (pd *spiDriver) setWord(word uint8, first bool, last bool) bool {
```
sago35, line 109: setWord() appears to be working well and efficiently.
Very nice.

examples/ili9341/pyportal_boing/feather-stm32f405.go
```diff
@@ -0,0 +1,34 @@
...
+		machine.SPI0,
+		dcPin,
+		csPin,
+		machine.D8, // if wired to 3.3V, pick an unused pin
+	)
+
+	// ILI9341's LED pin. set this to an unused pin (but not NoPin!) if
+	// wired via resistor straight to 3.3V. the boing example tries to
+	// set this pin and will panic if NoPin is used.
+	backlight = machine.D13
```
sago35, line 24: I hadn't noticed.
It is better to have the following changes made.

```go
// examples/pyportal_boing/main.go - main()
	if backlight != machine.NoPin {
		// configure backlight
		backlight.Configure(machine.PinConfig{machine.PinOutput})
		backlight.High()
	}
```

I think the target source code is as follows.

* examples/ili9341/basic/main.go
* examples/ili9341/pyportal_boing/main.go
* examples/ili9341/scroll/main.go

examples/ili9341/pyportal_boing/main.go
```diff
@@ -58,6 +58,7 @@ func main() {
 	backlight.High()
 
 	display.SetRotation(ili9341.Rotation270)
+	time.Sleep(50 * time.Millisecond)
```
sago35, line 61: Any reasons for the addition?

ili9341/spi_stm32f4.go
ardnew: I wasn't sure about that either, perhaps it clears certain flags (e.g., overrun/underrun?) in the SPI peripheral once the transaction has completed? You wouldn't want them to persist across transactions. 

But for good connections with a well-behaving slave, you can probably get by **without** doing it. 

Note that the motivation for this disabling and reenabling the peripheral comes from STM32CubeIDE-generated driver code, which also performs this set/clear on CR1 SPE in its Transmit/Receive functions - both their HAL driver and their LL driver do this.

For reference, I will attach my STM32CubeIDE project configured for the STM32F405 Feather. It has UART, I2C, and SPI interfaces all configured on their correct pins, and the system clock is configured with the 12 MHz HSE crystal onboard the feather.

examples/ili9341/pyportal_boing/main.go
ardnew: Unfortunately I don't have any documentation to justify this, only a use case:

Before the boing demo enters its main loop (drawing the ball frames), it first draws the entire 1-bit background across the screen. 

With my Feather, the top ~0.5cm of this background (including the top of the grid) was not being drawn, and would remain either white or have a scrambled gray coloring -- it seemed like the first SPI transactions were starting too quickly and the screen wasn't able to draw the data in time. The rest of the graphics continued normally without any problem. 

After adding this short delay, I never saw the problem again. 

I'll try to recreate and take a screenshot.

ili9341/spi_stm32f4.go
ardnew: I suggest removing this logic from the ILI9341 driver as you're implying. 

This logic would be better suited in `(SPI).Transfer` where robustness and bus stability is more important than performance.

Do you agree? Or would you remove it altogether?

aykevl:
> > What does this code do? Is it using a standard SPI connection or is it doing something special?
> 
> Which code are you referring to? If you're asking about all of it, then yes it's just using a standard SPI connection, along with the STM32F4-specific peripheral interface registers.

The thing is, if it can work with standard SPI (no STM32 specific registers) then that would be much better as it would support any chip with SPI support. That said, it would be great to add the necessary interfaces directly to the machine package so that chips don't have to be coupled to the drivers.

ardnew:
Oh I see. I thought this chip coupling was convention, since there exists SAMD21 and SAMD51 specific drivers in this package already. Your comment applies to those as well then.

That being said, I highly doubt we can achieve the SPI efficiency required for the `pyportal_boing` demo using the `(SPI).Transfer` method alone. These chip-specific drivers take advantage of the ILI9341's API, eliminating overhead performed with usual SPI transfers (e.g., not waiting for TXE/BSY, or toggling SPE, for every byte clocked out). @sago35 may be able to confirm or refute this from experience.

ardnew:
Indeed, I can't find a single other driver that includes chip-specific code. I would have to defer to others on how to expose this functionality in the machine package.

Amblyopius:
I think I can chip in a bit as I've been playing around with the ILI9341 drivers and have a local version that also reads (ATSAMD51 only currently). For ATSAMD51 (and ATSAMD21) SPI there's separate routines in order to have a send only option rather than using default SPI transfer which always reads too. I would assume this was mainly done for speed (though I have not tested yet if it really makes a difference). Essentially it puts the ATSAMD51/21 SPI in write only mode by disabling the receiver (using CTRLB.RXEN). It is true that having the receiver on is pointless in most cases as ILI9341 is not returning anything useful.

Also abstracting it in write8, write8n ... for ILI9341 driver was likely done because ILI9341 isn't always connected by SPI so an intermediate layer between the actual driver routines and the hardware layer might be needed. That said it should be trivial to have a generic SPI driver that works board independent and is simply based on generic SPI transfer function (which is exactly what I did to have reads). Having board specific SPI routines should provide some sort of benefit to be justified. If the other connection options provide an identical transfer function you could probably do away with a lot of code.

sago35:
> Oh I see. I thought this chip coupling was convention, since there exists SAMD21 and SAMD51 specific drivers in this package already. Your comment applies to those as well then.
> 
> That being said, I highly doubt we can achieve the SPI efficiency required for the `pyportal_boing` demo using the `(SPI).Transfer` method alone. These chip-specific drivers take advantage of the ILI9341's API, eliminating overhead performed with usual SPI transfers (e.g., not waiting for TXE/BSY, or toggling SPE, for every byte clocked out). @sago35 may be able to confirm or refute this from experience.

Waiting for TXE will result in several times worse performance.
For example, in the samd41 example, the performance goes from 40 fps to 10 fps.
Therefore, I think the ILI9341 driver needs to manipulate registers independently instead of using SPI.

https://github.com/tinygo-org/drivers/pull/153#issuecomment-631448289
https://github.com/tinygo-org/drivers/pull/153#issuecomment-632033771

ili9341/spi_stm32f4.go
sago35: In my opinion, it should be removed from ILI9341.
And as you say, I think we should leave it in SPI.Transfer().


https://github.com/tinygo-org/tinygo/blob/7fc5095b6b768a6e441752223724b27adaf2f014/src/machine/machine_stm32_spi.go#L150

aykevl:
If disabling the SPI receiver is the only difference, then it should be relatively straightforward to implement in the machine package. In fact, some chips already have support for this.

See: https://github.com/tinygo-org/tinygo/blob/dev/src/machine/machine_nrf.go#L438-L452
In this case the TXD byte is double buffered, which ensures the SPI bus is in constant use (no pause between sent bytes).
Other chips probably work differently.

Essentially, you can implement a custom `SPI.Tx` function in the machine package that switches to write-only behavior when the rx byte slice is empty. This has the big advantage that it improves SPI for all drivers, not just this ili9341 driver.

ardnew:
> Essentially, you can implement a custom `SPI.Tx` function in the machine package that switches to write-only behavior when the rx byte slice is empty. This has the big advantage that it improves SPI for all drivers, not just this ili9341 driver.

Clever! I'd really like to attempt implementing this, but would be a separate effort from the current PR for SPI on the STM32F405. 

This ILI9341 driver PR is still in draft, so if I can get that working on the machine-side, we can remove 90% of the additions in this PR.

PR #197, opened 2020-09-04
ILI9341: Fix for RAMWR bug + cosmetic changes to comments

Amblyopius:
RAMWR can not be skipped as you may call setwindow even if device is not ready for data write to buffer.
Data can only be written if last command was Memory Write or Write Memory Continue which was not guaranteed
There's no performance loss as having the exact same window on a consecutive call is extremely unlikely

Also fixed some typos and added a few missing comments to please go-lint

### DIFF
--- a/ili9341/ili9341.go
+++ b/ili9341/ili9341.go
@@ -20,7 +20,7 @@ type Device struct {
 	driver   driver
 
 	x0, x1 int16 // cached address window; prevents useless/expensive
-	y0, y1 int16 // syscalls to PASET, CASET, and RAMWR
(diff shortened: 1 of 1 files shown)


bgould:
Thank you @Amblyopius this looks good, thank you for pointing out the detail about the RAMWR and for cleaning up the comments.  I will try it out on some hardware later, and presuming everything is working I think this can be merged.

deadprogram:
@bgould did you happen to get a chance to try this out yet?

deadprogram:
Tested out on Pyportal worked exactly as expected.

Now merging, thanks for the contribution @Amblyopius

PR #198, opened 2020-09-04
ws2812: add support for the Xtensa architecture

aykevl:
This allows WS2812 LEDs to be controlled by an ESP32. Right now it only supports 160MHz operation, which is the default with current ESP32 support in TinyGo.

This PR depends on https://github.com/tinygo-org/tinygo/pull/1354 to work.


deadprogram:
Cool work, thanks @aykevl

PR #199, opened 2020-09-06
Renamed NewSpi() to NewSPI() in accordance with Go naming conventions

bgould:
This change renames the NewSpi() constructors for the ILI9341 driver to better follow Go naming conventions, as pointed out in https://github.com/tinygo-org/drivers/pull/196#discussion_r483991299

This is technically a breaking change but I think the disruption should be minimal as the current names have only been a released version of the drivers for about a month.  @sago35 what do you think?

### DIFF
--- a/examples/ili9341/basic/atsamd21.go
+++ b/examples/ili9341/basic/atsamd21.go
@@ -9,7 +9,7 @@ import (
 )
 
 var (
-	display = ili9341.NewSpi(
+	display = ili9341.NewSPI(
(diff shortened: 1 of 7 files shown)


sago35:
@bgould 
I agree with this change.
It's a bad name for the function I originally gave it.

so LGTM.

sago35 (approved):

deadprogram:
I should have caught that on the original PR. Thanks for the change @bgould now merging.

PR #200, opened 2020-09-09
ws2812: add support for ESP8266

aykevl:
This patch adds support for the ESP8266 chip by adding support for 80MHz operation. This should also work when the ESP32 is changed to 80MHz, but I didn't test it (as there is not currently a way to do that).

Verified on a NodeMCU dev board with an ESP8266. Depends on https://github.com/tinygo-org/tinygo/pull/900.


deadprogram:
:traffic_light: merging. Thanks @aykevl

PR #207, opened 2020-10-16
ws2812: make AVR support more robust

aykevl:
* Merge AVR support for Digispark and non-Digispark
* Fix the error "inline assembly requires more registers than available"
* Use a single file for all AVR targets, in a similar style as the
  Xtensa support.


deadprogram:
Thanks @aykevl works as expected. Merging.

PR #210, opened 2020-11-01
drv883x: add drv8838 device

elliots:
Some questions:

- I called it PhaseEnableDevice as that's what's in the pdf, is there a better name?

- I have only added the 8838 and not the 8837 as I don't have one of those. But it seems simple enough, is it ok to add without actually testing it?

image

https://www.ti.com/lit/ds/symlink/drv8838.pdf

### DIFF
--- a/drv8838/drv8838.go
+++ b/drv8838/drv8838.go
@@ -1,0 +1,74 @@
+// Package drv8838 provides a driver to the DRV8838 Low-Voltage H-Bridge Driver
+// typically used to control DC motors.
+//
+// Datasheet: https://www.ti.com/lit/ds/symlink/drv8838.pdf
+//
+package drv8838 // import "tinygo.org/x/drivers/drv8838"
+
+import (
+	"machine"
+)
+
+const (
+	DIRECTION_FORWARD  = false
+	DIRECTION_BACKWARD = true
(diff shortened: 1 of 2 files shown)


deadprogram:
Hi @elliots thank you for the contribution. Since you do not have a DRV8837 to test with, perhaps this PR should only cover the DRV8838 and renamed `drv8838` instead of `drv883x` until such time that the functionality for that device can be verified.

What do you think?

elliots:
Done.

deadprogram:
Sorry it has been so long to review this.

I notice that this PR needs a smoke test added. Would you like to do it, or would you prefer that I do so.

Thank you.

drv8838/drv8838.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// Device is a motor with direction (phase), speed (enable) and sleep controlled by a DRV8838
+type Device struct {
+	enable machine.PWM
+	phase  machine.Pin
+	sleep  machine.Pin // Optional
+}
+
+// Device returns a new motor device
```
thunderstorm99, line 24:
```suggestion
// NewDevice returns a new motor device
```
Fixed name in comment of function

PR #213, opened 2020-12-10
rgb75: Add GPIO-based driver for HUB75 RGB LED matrix panels

ardnew:
There already exists a [`hub75` driver in this repo](https://github.com/ardnew/drivers/tree/master/hub75), but it is incompatible with the pinout used on the Adafruit Matrix Portal M4 board: the driver uses SPI for clock+data, but the HUB75 interface on this board is not connected to SPI-capable GPIO pins.

This driver implements an abstracted bit-banged driver `rgb75` (in `rgb75/rgb75.go`), interface `Hub75` (in `rgb75/native/hub75.go`), an implementation of that interface for SAMD51 (specifically, for the Adafruit Matrix Portal M4, [tinygo#1529](https://github.com/tinygo-org/tinygo/pull/1529), in `rgb75/native/hub75_atsamd51.go`), and finally an example program to drive the display (in `examples/rgb75/main.go`).


aykevl:
I haven't yet looked into the code itself, but would it be possible to merge this with the existing hub75 driver? The method for sending data is different, but it seems to me that most other parts (OE pin timing, adapting to the various display layouts, framebuffers, maybe parts of the bitplanes, etc) are the same between the drivers. The difference is that one sends data by sending one long bitstring for all colors via SPI while the other uses direct port writes.

Also, while I like the example I would prefer if it was shorter. The goal of an example is to show how to use the driver in a kind of "hello world" fashion, while you've written an entire animation.

ardnew:
> would it be possible to merge this with the existing hub75 driver?

Yes, it should  be possible of course, but I suspect you and I will have vastly different opinions on how to implement this dual-role (SPI+GPIO) capability internally. 

Instead of fumbling around trying to explain my thoughts, let me describe how I would prefer to approach this with pseudo-Go source:

### Let user select which mode to use (SPI, GPIO, _**or CUSTOM**_) by providing an object that complies with exported interface type
```go
type DataClocker interface {} // methods for controlling clock and sending RGB data

type SPIDataClock struct {}   // defines methods for DataClocker compliance using SPI interface
type GPIODataClock struct {}  // defines methods for DataClocker compliance using GPIO interface

type HUB75 struct { 
    dc DataClocker // common driver code calls DataClocker methods to abstract the physical transmission medium
    ...
}

// with this design, the user can provide a completely different transmission mechanism than the two ones (SPI+GPIO) we 
// provide, making it possible for someone to implement UART/CAN (+DMA?) support without driver modification.
func New(dc DataClocker, ...) *HUB75 {
    return &HUB75{dc: dc, ...}
}
```

### Let user select which mode to use (SPI, GPIO) by simply using desired constructor; all interface-related types are private, simplifying API
```go
type dataClocker interface {} // methods for controlling clock and sending RGB data

type spiDataClock struct {}   // defines methods for dataClocker compliance using SPI interface
type gpioDataClock struct {}  // defines methods for dataClocker compliance using GPIO interface

type HUB75 struct { 
    dc dataClocker // common driver code calls dataClocker methods to abstract the physical transmission medium
    ...
}

// with this design, the user is not responsible for (or have access to) the dataClocker interface abstraction, making it simpler to
// use, but also less flexible.
func NewSPI(...) *HUB75 {
    return &HUB75{dc: spiDataClock{...}, ...}
}
func NewGPIO(...) *HUB75 {
    return &HUB75{dc: gpioDataClock{...}, ...}
}
```

> The method for sending data is different, but it seems to me that most other parts (OE pin timing, adapting to the various display layouts, framebuffers, maybe parts of the bitplanes, etc) are the same between the drivers. The difference is that one sends data by sending one long bitstring for all colors via SPI while the other uses direct port writes.

This is all correct with respect to the driver package in this PR. I haven't verified those statements hold for the existing `hub75` driver package we are wanting to merge.

> Also, while I like the example I would prefer if it was shorter. The goal of an example is to show how to use the driver in a kind of "hello world" fashion, while you've written an entire animation.

Indeed, a simple example is needed. Arduino libraries often provide multiple example programs ranging in complexity from "Hello World" to "Host a chat relay server". Would you be opposed to including TWO example programs with this driver, assuming they are adequately documented and identified?

PR #215, opened 2020-12-20
pin: implement driver.Pin interface

deadprogram:
This PR implements the `drivers.Pin` interface so that many drivers can remove another dependency on the machine package.

The following drivers however still use `machine.Pin` each for various reasons:

- HD44780
- ILI9341
- L293x
- L9110x
- MCP3008
- microbitmatrix
- shifter
- wifinina
- ws2812

Each of these will require some resolution or just leave them how they are.
 
The drivers that use the new pin interface each has a very rudimentary unit test in place to serve as a starting point for more complete testing of that driver.

The new `make unit-test` task run the units for all drivers except those that have dependencies on `machine`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,12 +161,11 @@ endif
 	@md5sum ./build/test.hex
 
 DRIVERS = $(wildcard */)
-NOTESTS = build examples flash semihosting pcd8544 shiftregister st7789 microphone mcp3008 gps microbitmatrix \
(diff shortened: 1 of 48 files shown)


deadprogram:
Closing in favor of a somewhat different approach once https://github.com/tinygo-org/tinygo/pull/1537 or something else like that exists.

PR #216, opened 2020-12-21
Better interface "ReadTime" instead of "Time" for DS1307

conejoninja:
Same interface as DS3231 and the same format as the other drivers


deadprogram:
Uniformity of naming is good. At some point soon we can start to define real interfaces.

PR #219, opened 2021-01-02
Add bmp388 support

mcbunkus:
Add initial support for the bmp388 temperature and pressure sensor. I only have one microcontroller to test it on, but it's working without issues and the tests compile.


deadprogram:
Hello, @AgentMacklin thank you for the contribution In general it looks really good.

However, in order to match the other devices of this kind, you should change the signature of your functions a bit. For example, take a look at the `mag3110` driver(https://github.com/tinygo-org/drivers/blob/e376785596dc8269f3e8aa42a9bf75fb1457febc/mag3110/mag3110.go#L52): 

```go
func (d Device) ReadTemperature() (int32, error)
```

The reason that we use integer instead of floating point values is to avoid FP storage and operations to reduce the target memory footprint.

Can you please change the function signatures to match the existing functions of the same kind?

Thanks!

mcbunkus:
Shoot, I should have seen that earlier, I'll make the changes. Thanks!

bmp388/bmp388.go
```diff
@@ -0,0 +1,246 @@
...
+	// Turning on the pressure and temperature sensors and setting the measurement mode
+	err = d.writeRegister(REG_PWR_CTRL, PWR_PRESS|PWR_TEMP|byte(d.Config.Mode))
+
+	// Configure the oversampling, output data rate, and iir filter coefficient settings
+	err = d.writeRegister(REG_OSR, byte(d.Config.Pressure|d.Config.Temperature<<3))
+	err = d.writeRegister(REG_ODR, byte(d.Config.ODR))
+	err = d.writeRegister(REG_IIR, byte(d.Config.IIR<<1))
+
+	if err != nil {
+		return errors.New("bmp388: failed to configure sensor, check connection")
```
aykevl: Creating new error objects can be expensive. I believe they do not currently cause a heap allocation anymore, but they are still expensive. So can you please move these errors to global error values, such as this?

```go
var errConfiguration = errors.New("bmp388: failed to configure sensor, check connection")
```

Also note that there is no guarantee an error will be returned when the connection is broken.

bmp388/bmp388.go
```diff
@@ -0,0 +1,246 @@
...
+	return int32(temp), nil
+}
+
+// ReadPressure returns the pressure in centipascals, i.e 10132520 / 100 = 101325.20 Pa
+func (d *Device) ReadPressure() (int32, error) {
+
+	_, err := d.ReadTemperature()
+	if err != nil {
+		return 0, err
+	}
```
aykevl: Why are you reading the temperature here? Is reading the temperature required by the datasheet?

bmp388/bmp388.go
```diff
@@ -0,0 +1,246 @@
...
+// ReadAltitude estimates the altitude above sea level in centimeters.
+func (d *Device) ReadAltitude() (alt int32, err error) {
+	press, err := d.ReadPressure()
+	if err != nil {
+		return
+	}
+	// pulled from bme280 driver, modified a bit to return altitude in centimeters
+	alt = int32(4433000.0 * (1.0 - math.Pow(float64(press)/float64(d.Config.SeaLevelPressure), 0.1903)))
+	return
+}
```
aykevl: I don't know what @deadprogram thinks about it but my personal opinion is that this should not be put in the driver itself. Instead, such related algorithms can be put in a separate package (maybe part of the drivers repository, maybe not) for miscellaneous calculations and conversions. Because while this is undoubtedly useful, it is relevant to all pressure sensors and not just this pressure sensor.

bmp388/bmp388.go
```diff
@@ -0,0 +1,246 @@
...
+// Connected tries to reach the bmp388 and check its chip id register. Returns true if it was able to successfully
+// communicate over i2c and returns the correct value
+func (d *Device) Connected() bool {
+	data, err := d.readRegister(REG_CHIP_ID, 1)
+	return err == nil && data[0] == CHIP_ID // returns true if i2c comm was good and response equals 0x50
+}
+
+// SetMode changes the run mode of the sensor, NORMAL is the one to use for most cases. Use FORCED if you plan to take
+// measurements infrequently and want to conserve power. SLEEP will of course put the sensor to sleep
+func (d *Device) SetMode(mode Mode) error {
```
aykevl, line 207: We might eventually want to standardize this sort of thing, but for now it's okay.

bmp388/registers.go
```diff
@@ -0,0 +1,84 @@
+// Package bmp388 provides a driver for Bosch's BMP388 digital temperature & pressure sensor.
+// The datasheet can be found here: https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bmp388-ds001.pdf
+package bmp388
+
+const ADDRESS byte = 0x77 // default I2C address
```
aykevl: The convention in Go is to not use all caps like here.

```suggestion
const Address byte = 0x77 // default I2C address
```

(also, the `byte` type is unnecessary but fine).

bmp388/bmp388.go
mcbunkus: Yes, there is a value called `tlin` from the temperature compensation calculations that the pressure calculations use too. I used `ReadTemperature` here because it already does all the compensation work to calculate `tlin` then returns the temperature by dividing by a constant. I'll refactor this to be less confusing.

bmp388/bmp388.go
mcbunkus: I agree that that's a better way to do it. There's more than one way to estimate the altitude, so it can be left up to the end user on how they want to do it. I'll pull this out.

mcbunkus:
I've added all the changes and updated the example accordingly. Please let me know if I missed anything.

Thanks

examples/bmp388/main.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/bmp388"
+)
+
+func main() {
+
+	time.Sleep(time.Second)
```
deadprogram: Is there a specific reason for this startup delay in the example?

examples/bmp388/main.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	for {
+		temp, err := sensor.ReadTemperature() // returns the temperature in centicelsius
+		press, err := sensor.ReadPressure()   // returns the pressure in centipascals
+
+		if err != nil {
+			println(err)
+		} else {
+			fmt.Printf("Temperature: %d cC\r\n", temp)
```
deadprogram: I would recommend avoid using the `fmt` package in examples, to avoid the extra memory usage that reflection requires.

examples/bmp388/main.go
mcbunkus: No not really, it's left over from testing, I removed it.

examples/bmp388/main.go
mcbunkus: No problem, I removed it.

deadprogram:
Looks good @AgentMacklin now going to squash/merge. Thank you!

PR #220, opened 2021-01-07
P1AM base controller driver

quentinmit:
This PR adds basic support for the P1AM base controller. This is an embedded peripheral that is present on the [P1AM-100 board](https://facts-engineering.github.io/modules/P1AM-100/P1AM-100.html). Since this is a built-in peripheral, there's no configuration available for what SPI port and pins to use, etc. I included an example program so you can see what the API is like. I don't have a full set of expansion modules, so I've only been able to test/implement digital I/O modules.

### DIFF
--- a/p1am/internal/cmd/gen_defines/main.go
+++ b/p1am/internal/cmd/gen_defines/main.go
@@ -1,0 +1,133 @@
+package main
+
+import (
+	"bytes"
+	"go/format"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"text/template"
+)
+
+var tmpl = template.Must(template.New("main").Parse(`package p1am
+
+//go:generate go run ./internal/cmd/gen_defines
+
+type ModuleProps struct {
+	ModuleID                                 uint32
+	DI, DO, AI, AO, Status, Config, DataSize byte
+	Name                                     string
+}
+
+var modules = []ModuleProps{
+{{.MDB -}}
+}
+
+var defaultConfig = map[uint32][]byte{
+	{{range .Configs -}}
+	0x{{.ID}}: // {{.Name}}
+	{{index $.DefaultConfigs .Name}},
+	{{end}}
+}
+
+{{range .Defines}}
+const {{.Name}} = {{.Value}}{{.Comment -}}
+{{end}}
+`))
+
+func findLibrary() string {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		log.Fatal(err)
+	}
+	for _, dir := range []string{
+		"Documents/Arduino",
+		"Arduino",
+	} {
+		dir = filepath.Join(home, dir, "libraries/P1AM/src")
+		if _, err := os.Stat(dir); err == nil {
+			return dir
+		}
+	}
+	return ""
+}
+
+func definitions(path string, delim string) []string {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return strings.Split(string(data), delim)
+}
+
+var (
+	mdbRE    = regexp.MustCompile(`(?s)mdb\[\] = \{\s*(.+)\}`)
+	configRE = regexp.MustCompile(`(?s)const char (.*?)\[\] = (.+)`)
+	caseRE   = regexp.MustCompile(`(?s)case 0x([^:]+):\s+return \(char\*\)(.+)`)
+	defineRE = regexp.MustCompile(`(?ms)^\s*#define (\S+)\s+(\d+|0x[0-9a-fA-F]+)(\s+.*?)?\s*$`)
+)
+
+func main() {
+	base := findLibrary()
+	if base == "" {
+		log.Fatal("can't find Arduino library")
+	}
+	var data = struct {
+		MDB            string
+		DefaultConfigs map[string]string
+		Configs        []struct {
+			ID   string
+			Name string
+		}
+		Defines []struct {
+			Name    string
+			Value   string
+			Comment string
+		}
+	}{
+		DefaultConfigs: make(map[string]string),
+	}
+	for _, line := range definitions(filepath.Join(base, "Module_List.h"), ";") {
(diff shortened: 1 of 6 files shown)


Nerzal:
Some things, that will help to get this PR going. 

You might want to update the branch by merging in the current dev branch.
Also you can add the driver to the Readme and also increment the number of supported devices in there. 
And you also want to add the main from the example to the tests in the Makefile  like in this example:
https://github.com/tinygo-org/drivers/blob/5741ceb9d179dab8b88e290fda2ad5486345303c/Makefile#L138

quentinmit:
@Nerzal Thanks, I have addressed those three comments.

deadprogram:
Hello, there. First of all, thanks for working on this contribution. Sorry I have not responded more quickly. 

The biggest thing that has prevented me from digging in further to this PR is the perl script generator. I think we run the risk of adding some code that no one will be able to maintain. We have actually done a considerable amount of effort to replace our original python code for generating the `device` files in the TinyGo repo with Go programs.

So I am not sure how to proceed further. What do you think?

sago35:
It would be better to convert the script to Go instead of Perl.

Alternatively, it might not be necessary to push it to tinygo-org/drivers, since it's simple.

quentinmit:
> Hello, there. First of all, thanks for working on this contribution. Sorry I have not responded more quickly.
> 
> The biggest thing that has prevented me from digging in further to this PR is the perl script generator. I think we run the risk of adding some code that no one will be able to maintain. We have actually done a considerable amount of effort to replace our original python code for generating the `device` files in the TinyGo repo with Go programs.
> 
> So I am not sure how to proceed further. What do you think?

The perl script just converts the constants from the upstream library's header files into the .go file that is checked in here; users wouldn't have to run it themselves unless there's an update upstream that they want to include before someone gets around to updating the .go file. I intentionally committed the generated .go file so this is just a maintenance task, not a part of the build process.

I guess it's mostly just a series of regexes, so it ought to be fairly simple to convert to Go if you really want... I'm not thrilled with that but if it's your only comment on this PR I'm happy to do it. (But, uh, transforming a list of constants is a very small part of the PR - I'm much more interested in the structure of the driver itself. You don't need to understand the perl code at all in order to review that.)

quentinmit:
@deadprogram @sago35 I converted the Perl script to a Go program... it is now exactly twice as long, and IMO less readable. WDYT?

p1am/internal/cmd/gen_defines/main.go
```diff
@@ -0,0 +1,133 @@
...
+		"Documents/Arduino",
+		"Arduino",
+	} {
+		dir = filepath.Join(home, dir, "libraries/P1AM/src")
+		if _, err := os.Stat(dir); err == nil {
+			return dir
+		}
+	}
+	return ""
+}
```
sago35, line 56: Thank you very much. I think it's good.

Please write the version of the arduino-codes you used as the base in the commit message or source code.

I think it is the following, but there may be another version as Arduino library.

* https://github.com/facts-engineering/P1AM/tree/da6d3a1463c17aa3913be83df678e222857638d6

quentinmit: I've added a comment linking to the version of the Arduino library I used as a base (v1.0.1).

deadprogram:
Hi @quentinmit I think there are just a couple of things still needed.

The first is since this driver cannot have a unit test written due to using `machine.SPI` instead of `drivers.SPI` interface, it needs to be added here in order to prevent the current CircleCI build failure: https://github.com/tinygo-org/drivers/blob/dev/Makefile#L176-L178

Once that is done, then just need to rebase against the `dev` branch, then squash the commits in the PR.

Thank you very much.

quentinmit:
@deadprogram I added it to `NOTESTS`, though it should be possible to use build tags and a local interface definition to get (most of) the code built for testing. (I think it would mostly involve declaring an unexported interface to use in lieu of `machine.SPI` in the `P1AM` struct and move the `Controller` variable definition into a separate file that is restricted to the `p1am` build tag. Then you could theoretically write unit tests with a mock SPI interface.)

I saw the CircleCI failure but I can't get them to cough up the build log without granting them write access to every repo on my account (!). I'm not willing to do that, sorry.

The dev branch is already merged in as of this morning. Isn't squashing something that would happen as part of merging the PR?

deadprogram:
Thanks for adding to `NOTESTS`. As you might see, the CI is now passing.

>  it should be possible to use build tags and a local interface definition to get (most of) the code built for testing

That is certainly true. See the drivers that use a mock I2C for unit tests:
https://github.com/tinygo-org/drivers/blob/release/tester/i2c.go

Someone needs to do some similar work to provide a mock SPI interface, that would make it a lot easier.

Regarding the squashing, for TinyGo it is just up to the individual preference. I can squash commits when merging, or the contributor can do so.

quentinmit:
> Regarding the squashing, for TinyGo it is just up to the individual preference. I can squash commits when merging, or the contributor can do so.

@deadprogram Personally, I try to structure my commits so they're meaningful to someone who comes later. If' you're okay with that, I'm happy to do a rebasing pass over these commits to squash them into a minimal set of functional blocks, and them you can commit with a rebase onto the tip of the branch. But if you're going to just squash them all into one commit, you should just do that.

deadprogram:
We generally try to make it so each commit is atomic, and only contains a single complete unit of functionality for that change. Hence why we try to invite people to rebase their own commits. Otherwise, we just squash them together, as long as everything else seems alright.

Hopefully that makes sense in context for this project.

quentinmit:
> We generally try to make it so each commit is atomic, and only contains a single complete unit of functionality for that change. Hence why we try to invite people to rebase their own commits. Otherwise, we just squash them together, as long as everything else seems alright.
> 
> Hopefully that makes sense in context for this project.

I gotcha, when you asked me to squash above I interpreted that as a request for a single commit. I've force-pushed a set of commits that each contain a single complete unit of functionality.

deadprogram:
Thank you @quentinmit for making all of the requested changes. Now merging at long last!

PR #222, opened 2021-01-24
microbitmatrix: support for microbit v2

deadprogram:
This is WIP on support for the microbit v2.

V2 is not yet working, but at least allows both v1 and v2 to be compiled.

### DIFF
--- a/microbitmatrix/microbitmatrix-v1.go
+++ b/microbitmatrix/microbitmatrix-v1.go
@@ -1,0 +1,112 @@
+// +build microbit
+
+// Package microbitmatrix implements a driver for the BBC micro:bit's LED matrix.
+//
+// Schematic: https://github.com/bbcmicrobit/hardware/blob/master/SCH_BBC-Microbit_V1.3B.pdf
+//
+package microbitmatrix // import "tinygo.org/x/drivers/microbitmatrix"
+
+import (
+	"machine"
+	"time"
+)
+
+var matrixRotations = [4][5][5][2]uint8{
+	{ // 0
+		{{0, 0}, {1, 3}, {0, 1}, {1, 4}, {0, 2}},
+		{{2, 3}, {2, 4}, {2, 5}, {2, 6}, {2, 7}},
+		{{1, 1}, {0, 8}, {1, 2}, {2, 8}, {1, 0}},
+		{{0, 7}, {0, 6}, {0, 5}, {0, 4}, {0, 3}},
+		{{2, 2}, {1, 6}, {2, 0}, {1, 5}, {2, 1}},
+	},
+	{ // 90 CCW
+		{{0, 2}, {2, 7}, {1, 0}, {0, 3}, {2, 1}},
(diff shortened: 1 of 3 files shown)


deadprogram:
@conejoninja maybe you can help me take a look into why it is not working for v2?

deadprogram:
This PR for sure requires https://github.com/tinygo-org/tinygo/pull/1595 but still does not work as expected.

conejoninja:
I'm unable to find a diagram of the v2, but pin numbers seem ok. What issues are you experiencing? Let's try to enable one led and see if that works. Does the code work on v1?

deadprogram:
This program can run on the v2: https://github.com/tinygo-org/tinygo/blob/microbit-v2-matrixpins/src/examples/microbit-blink/microbit-blink.go

conejoninja:
We'll need to perform some tests. I couldn't find the electric diagram, nor some example code, maybe we need to made some changes on the order of the columns/rows in "Display" function.


Please try this:
test 1
```go 
for {
        display.EnableAll()
        time.Sleep(1000*time.Millisecond)
        display.DisableAll()
        time.Sleep(1000*time.Millisecond)
}
```


test 2
```go 
x := int16(1)
y := int16(2)
c := color.RGBA{255, 255, 255, 255}
for {
        display.SetPixel(x,y,c)
        display.Display();
        time.Sleep(1000*time.Millisecond)
        display.DisableAll()
        time.Sleep(1000*time.Millisecond)
}
```
change values of x,y to see if it lit the correct pixel, then try enabling a whole row or column

deadprogram:
Thanks for the assist @conejoninja now working with my latest commit and https://github.com/tinygo-org/tinygo/pull/1595

deadprogram:
The build failures are not related to this PR. See https://github.com/tinygo-org/tinygo/issues/1599

deadprogram:
Refactored to eliminate duplicate code with v1/v2 and rebased against `dev` so tests should now be passing.

deadprogram:
@conejoninja ready for review again, and is fully working.

microbitmatrix/microbitmatrix-v1.go
```diff
@@ -0,0 +1,112 @@
...
+	for i := machine.LED_COL_1; i <= machine.LED_COL_9; i++ {
+		d.pin[i-machine.LED_COL_1].Low()
+	}
+	for i := machine.LED_ROW_1; i <= machine.LED_ROW_3; i++ {
+		d.pin[i-machine.LED_COL_1].High()
+	}
+}
+
+// Size returns the current size of the display.
+func (d *Device) Size() (w, h int16) {
```
conejoninja, line 110: Could this be in microbitmatrix.go to avoid repetition? Both matrices are 5x5

conejoninja (changes requested):
func Size() is repeated in v1 and v2, since both are 5x5 we could put it in microbit.go

deadprogram:
I kept them separate only because there might be other matrix displays we can support with only a few changes.

conejoninja:
In that case, if it works, let's merge it. Great work, thanks.

PR #224, opened 2021-01-27
all: fix main package in examples

aykevl:
The package with the main function should always have the name main.
This was not the case in three packages.

This was silently allowed before, but since a TinyGo change
(https://github.com/tinygo-org/tinygo/pull/1592) this now results in a
linker failure.

Perhaps this should result in a better error message in TinyGo. However,
the example code also needs to be fixed, so hence this PR.

### DIFF
--- a/examples/hcsr04/main.go
+++ b/examples/hcsr04/main.go
@@ -1,4 +1,4 @@
-package hcsr04
+package main
 
 import (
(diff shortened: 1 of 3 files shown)


deadprogram:
That makes sense, thanks for finding it and for the explanation.

Now merging.

PR #225, opened 2021-01-30
Device/max7219

Nerzal:
Add an implementation for the max7219 IC, which is commonly used to drive 7 segment displays and 8x8 led matrices. 
Works like a charm on my tests :)


Nerzal:
build fails, caused by: "package machine is not in GOROOT"

conejoninja:
Hi Nerzal, thanks for your contribution, I've been wanting to driver a driver like this for some time so I could finally use the max7219s I have. After a quick review, here are some comments:
- I think it's better to rename device.go to max7219.go to match the style of the rest of the drivers
- In fact, if the code is compatible with max7219 && max7221 the driver should be max72xx instead
- Comment every public function, so GoDoc could be generated properly
- A smoke test should be added in Makefile : https://github.com/tinygo-org/drivers/blob/release/Makefile#L10
- It needs to be added to the List in README.md and the number of supported drivers bumped too : https://github.com/tinygo-org/drivers/blob/release/README.md

Nerzal:
Okay, i'll update the pr in some minutes. 
The only thing, that this currently does not support is daisy chaining. And i sadly don't have time to check (and missing hardware to daisy chain). 
But i guess, that functionality could be added later :)

Nerzal:
@conejoninja done, but i don't get, why the build fails

Nerzal:
Btw some additional thoughts: 
As is is possibly to control different kind of devices, i did not directly want to integrate the 7 segment logic in that package. 
I guess it's best to add new packages for 7 segment display, 8x8 matrix and all other devices you can control with that one.

deadprogram:
We have both unit tests (mostly i2c, executed using normal go) and also smoke tests (built using tinygo)

Once we have interfaces for everything this will not be a problem. In the meantime, so that the tests can pass please add your package to the list here: https://github.com/tinygo-org/drivers/blob/dev/Makefile#L170-L172 thanks!

Nerzal:
oh thank you, i'll do that :)

Nerzal:
The tests do still fail. 
But it worked in my other pullrequest

max72xx/max72xx.go
```diff
@@ -0,0 +1,96 @@
+// Driver works for max7219 and 7221
+// Datasheet: https://datasheets.maximintegrated.com/en/ds/MAX7219-MAX7221.pdf
+package max72xx
+
+import (
+	"machine"
+)
+
+type Device interface {
```
deadprogram: Not sure why this is declared as an interface here. Seems like simpler to just implement directly without the indirection.

max72xx/max72xx.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type device struct {
+	bus  machine.SPI
+	load machine.Pin // load
+}
+
+// NewDriver creates a new max7219 connection. The SPI wire must already be configured
+// The SPI frequency must not be higher than 10MHz.
+func NewDevice(load machine.Pin, bus machine.SPI) Device {
```
deadprogram: The other SPI drivers generally have the SPI bus param first. Also isn't "load" really the chip select pin? e.g.

```go
func NewDevice(bus machine.SPI, cs machine.Pin) Device
```

deadprogram:
I made a couple of comments.

Another general comment would be that some of the code in the example seems like perhaps it should be in the driver itself. Otherwise only using `WriteCommand()`  seems very low level and a bit harder to use.

What do you think?

conejoninja:
I agree `WriteCommand()` is not clear what the parameters should be to display something on the display, I expect the first param to be a command. Maybe we should add a wrapper that's more clear, `sendChar` or `sendByte` ¿?. Due to the nature of the driver, used in multiple different displays (8-segments display, 8x8 led matrices,...), it will be hard to get a better common interface.

I'd like to see support to daisy-chain too. An old arduino library for max72xx has the following interface, I don't really like the names, but I like the spirit :
- void maxSingle( byte reg, byte col) // if you only have one max72xx
- void maxAll (byte reg, byte col) // display the same in all max72xx daisy-chained
- void maxOne(byte maxNr, byte reg, byte col) // send a reg/col to specific daisy-chained max72xx

max72xx/max72xx.go
Nerzal: I can change the order of the parameters. 

The load pin actually is the cs pin. 
The datasheet refers to it as load (cs) pin, which is why i used the same wording. But i can also just rename it to cs

max72xx/max72xx.go
Nerzal: removed it

max72xx/max72xx.go
Nerzal: fixed :)

Nerzal:
It is true, that the WirteCommand function is very low level. 
I just could not figure out a useful abstraction, that can be used for all the different devices, which this driver is capable of handling. 

@conejoninja i could try to add daisy chaining functionality, but i currently only have 1 single max7219 chip, so i need to order some more first, so i could actually test it.

Nerzal:
I ordered 5 more max7219 chips, but they take 1-3 weeks until they arrive, but i can update the PR as soon as i have them and introduce daisychaining functionalities. :)

Makefile
```diff
@@ -165,13 +165,15 @@ endif
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=circuitplay-express ./examples/lis2mdl/main.go
 	@md5sum ./build/test.hex
+	tinygo build -size short -o ./build/test.hex -target=arduino-nano33 ./examples/max72xx/main.go
+	@md5sum ./build/t
 	tinygo build -size short -o ./build/test.hex -target=feather-m0 ./examples/dht/main.go
-	@md5sum ./build/test.hex
+	@md5sum ./build/t
```
deadprogram: Same problem as the other PR.

Nerzal: fixed, sorry for that

deadprogram:
I squshed and merged into `dev` by way of a series of complicated `git rebase -i`.

Next time, @Nerzal you need to rebase against dev more frequently. The result is everyone will probably need to use `-f` to force rebase against dev for now.

Anyhow, closing this PR, as the feature is now in `dev` branch. Thanks for working on it @Nerzal

Nerzal:
Oh sorry for that!

PR #228, opened 2021-02-01
adc: update drivers with ADC to use new config struct

deadprogram:
This PR updates the drivers with ADC to use new config struct, which is now required for the CI build to pass.


conejoninja:
The change required was already merged in tinygo/dev, so merging this. Thanks

PR #230, opened 2021-02-20
fix hd44780

ncw:
- ws2812: make AVR support more robust
- spi: remove machine.SPI and replace with drivers.SPI interface for almost all SPI drivers
- spi: incorporate change from GH issue feedback
- drivers/flash: restore previous calls directly to machine package until we implement SetClockSpeed()
- test: run unit tests against i2c drivers and any spi drivers without direct gpio
- Better interface "ReadTime" instead of "Time" for DS1307
- Add bmp388 support (#219)
- mcp23017: new driver for MCP23017 (I2C port expander)
- mcp23017: implement pin toggling
- mcp23017: use new tester package
- all: fix main package in examples
- microbitmatrix: matrix now working on microbit v2
- microbitmatrix: refactor to eliminate duplicate code with microbit v1/v2
- st7789: add scrolling functions to match st7735
- st7789: correct errors on various godoc comments
- adc: update drivers with ADC to use new config struct
- hd44780: add a mode to work with boards where the RW pin is grounded


ncw:
Ooops sorry, wrong branch

PR #231, opened 2021-02-20
hd44780: add a mode to work with boards where the RW pin is grounded

ncw:
On some HD44780 boards (eg the Keyestudio LCD1602 expansion shield),
the RW pin isn't brought out and is permanently grounded.

This means that the board can't be read from, and in particular the
busy status can't be read.

This patch adapts the package to work with boards like these.

To signal this to the package, set the RW pin to machine.NoPin.

The package will then disallow all reading and use adjustable timing
based writing. The timing can be adjusted the configuration.

### DIFF
--- a/hd44780/gpio.go
+++ b/hd44780/gpio.go
@@ -57,9 +57,16 @@ func (g *GPIO) SetCommandMode(set bool) {
 	}
 }
 
+// WriteOnly is true if you passed rw in as machine.NoPin
+func (g *GPIO) WriteOnly() bool {
+	return g.rw == machine.NoPin
+}
+
 // Write writes len(data) bytes from data to display driver
 func (g *GPIO) Write(data []byte) (n int, err error) {
-	g.rw.Low()
+	if !g.WriteOnly() {
+		g.rw.Low()
+	}
 	for _, d := range data {
 		g.write(d)
 		n++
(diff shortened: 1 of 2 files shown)


deadprogram:
Seems to make sense to me. @conejoninja any comments?

hd44780/gpio.go
```diff
@@ -11,6 +11,7 @@ type GPIO struct {
 	en       machine.Pin
 	rw       machine.Pin
 	rs       machine.Pin
+	noRW     bool
```
conejoninja: this is not used anywhere

conejoninja:
Also, here : https://github.com/ncw/drivers/blob/fix-hd44780/hd44780/hd44780.go#L223
could we change it to 
```go
for d.busy(false) {
```


It will not make any difference, but I guess it make sense to use the internal function. Anything else looks good to me, thanks for the contribution

hd44780/gpio.go
ncw: Well spotted! Will fix.

ncw:
> Also, here : https://github.com/ncw/drivers/blob/fix-hd44780/hd44780/hd44780.go#L223
> could we change it to
> 
> ```go
> for d.busy(false) {
> ```
> 
> It will not make any difference, but I guess it make sense to use the internal function. Anything else looks good to me, thanks for the contribution

Sure thing - that makes sense :-)

I've pushed those changes now. Let me know if you'd like any other changes.

PS Having a lot of fun with tinygo - thanks for a great project :-)

deadprogram:
@conejoninja any other feedback before we merge?

conejoninja:
looks good to me

conejoninja:
Now merged, thanks for your contribution @ncw

PR #232, opened 2021-02-22
Draft: websocket draft

Nerzal:
This PR is not meant to be merged right now and should currently only be considered as a base of discussion on how and where we can implement websockets and how to integrate them in mqtt.


Nerzal:
Okay so this WASM package needs to be separated from the other packages as dependencies lead to a build failure.

Nerzal:
i'll open a new PR later, when i have a better place and a new plan :D

Issue #117, opened 2020-01-06
Driver request for DHT-11 and DHT-22


homebrew79:
I would like to request a driver for the DHT-11 and DHT-22 Temperature/Humidity sensor. I have seen a few libraries on github for go, but they seem to be specific to raspberry pi.

conejoninja:
DHT-11/22 are quite common and it would be good to implement them.

I left this  ( https://github.com/MichaelS11/go-dht ) as a reference for whoever want to take the challenge , this might be a good start since it's a pure go implementation but it depends on "runtime/debug". DHT-xx are timing sensitive and garbage collection might interfere in it.

alankrantas:
Tried to write one as an experiment. It...kind of works? But there are still some weird issues. I'll leave this here and see if it can be solved later.

This is based on my micro:bit MakeCode driver (which both DHT11 and DHT22 works). 

I tested this script on micro:bit; only DHT11 works, kind of. The checksum is always a bit off, but humidity and temperature readings looks ok, so I simply ignore checksum here. (The value of 2nd and 4th bytes are very small so I ignore them as well.)

DHT22 doesn't work. All bytes are either 255 or 254 (almost all bits are read HIGH - why?). So I haven't bothered to calculate results for DHT22. On Arduino Uno both DHT11 and DHT22 are like this as well. 

The sensors I have are the 3-pin PCB version, which already pulled high data pins. They performs the same on 3.3V or 5V.

```golang
package main

import (
	"machine"
	"time"
)

type Device struct {
	Pin machine.Pin
}

func New(pin machine.Pin) Device {
	return Device{pin}
}

func (d *Device) Measure() (humidity uint8, temperature uint8, err error) {
	
	var data [40]bool
	var result [5]uint8
	
	// initialize
	d.Pin.Configure(machine.PinConfig{Mode: machine.PinOutput})
	d.Pin.Set(false)
	time.Sleep(time.Millisecond * 18)
	d.Pin.Configure(machine.PinConfig{Mode: machine.PinInput})
	time.Sleep(time.Microsecond * 20)
	
	// wait for ready signal
	for d.Pin.Get() {}
	for !d.Pin.Get() {}
	for d.Pin.Get() {}
	
	// read 40 bits from DHT11
	for i := 0; i < 40; i++ {
	    for d.Pin.Get() {}
	    for !d.Pin.Get() {}
	    time.Sleep(time.Microsecond * 28)
	    data[i] = d.Pin.Get()
	}
	
[...]

PR #235, opened 2021-03-03
DHTXX driver

pleomaxx3002:
Implementation of DHTXX driver based on AdaFruit's driver. #117  

Tested with DHT11 and AdaFruit feather-m0


deadprogram:
Hello @pleomaxx3002 thank you for this contribution.

Could you please add an example under the `examples` directory, and also add it to the "smoketest" make task similar to the other drivers? https://github.com/tinygo-org/drivers/blob/release/Makefile#L158-L159

Thanks!

deadprogram:
Thanks for the example @pleomaxx3002 one last change I'd would like to ask can you please eliminate the camel case file names and rename them to all lower case? Thank you very much.

deadprogram:
Thank you for making all of the requested updates on this driver @pleomaxx3002 now squash/merging.

Thanks again!

PR #236, opened 2021-03-05
all: do not take the pointer of an I2C object

aykevl:
This is in preparation of PR https://github.com/tinygo-org/tinygo/pull/1693, which makes `machine.I2C0` and similar objects of pointer type, so they can be freely passed around.


deadprogram:
Thanks @aykevl now merging.

PR #237, opened 2021-03-05
all: use interfaces for UART objects

aykevl:
This makes it possible to replace UART objects with dummy implementations, for example. Or allows changing the `machine.UART` type to `*machine.UART` without breaking compatibility.

Similar to https://github.com/tinygo-org/drivers/pull/208 and https://github.com/tinygo-org/drivers/pull/193.

### DIFF
--- a/espat/espat.go
+++ b/espat/espat.go
@@ -20,17 +20,17 @@ package espat // import "tinygo.org/x/drivers/espat"
 
 import (
 	"errors"
-	"machine"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


deadprogram:
Excellent. This will help greatly with testing as well as implementation.

deadprogram:
@aykevl I think this branch needs to be rebased against `dev`.

aykevl:
Done.

deadprogram:
Now merging.

PR #240, opened 2021-03-28
Add INA260

kenbell:
Adds the INA260, which is a current measuring device.

As part of this change, to add unit-tests, the unit test mock was re-factored.  The existing mock assumed all registers were 1-byte in length and that reading 2-bytes would read 2 registers.  This doesn't apply to all devices.  The existing mock is moved into `I2CContiguousRegisterDevice`, a new mock `I2CClassicDevice` and an interface is defined to unify the two.

Not sure about the new names for the mocks - very open to suggestions for better names.

### DIFF
--- a/ina260/ina260.go
+++ b/ina260/ina260.go
@@ -1,0 +1,112 @@
+package ina260
+
+import "tinygo.org/x/drivers"
+
+// Device wraps an I2C connection to an INA260 device.
+type Device struct {
+	bus     drivers.I2C
+	Address uint16
+}
+
+// Config holds the configuration of the INA260 device.
+type Config struct {
+	// One of AVGMODE_XXX
+	AverageMode byte
+
+	// One of CONVTIME_XXXXUSEC
+	VoltConvTime byte
+
+	// One of CONVTIME_XXXXUSEC
+	CurrentConvTime byte
+
+	// Multiple of MODE_XXXX
+	Mode byte
+}
+
+// New creates a new INA260 connection. The I2C bus must already be
+// configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func New(bus drivers.I2C) Device {
+	return Device{
+		bus:     bus,
+		Address: Address,
+	}
+}
+
+// Configure sets up the device.
+//
+// This only needs to be called to override built-in defaults. By default,
+// the device starts with:
+//
+// * AverageMode = AVGMODE_1
+// * VoltConvTime = CONVTIME_1100USEC
+// * CurrentConvTime = CONVTIME_1100USEC
+// * Mode = MODE_CONTINUOUS | MODE_VOLTAGE | MODE_CURRENT
+//
+func (d *Device) Configure(cfg Config) {
+	var val uint16
+
+	val = uint16(cfg.AverageMode&0x7) << 9
+	val |= uint16(cfg.VoltConvTime&0x7) << 6
+	val |= uint16(cfg.CurrentConvTime&0x7) << 3
+	val |= uint16(cfg.Mode & 0x7)
+
+	d.WriteRegister(REG_CONFIG, val)
+}
+
+// Resets the device, setting all registers to default values
+func (d *Device) Reset() {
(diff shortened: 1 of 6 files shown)


deadprogram:
Was just looking over the changes made to the mock. I think that it would be a much simpler to add a property like `UsesContiguousRegister` instead of adding an additional type.

What do you think?

kenbell:
Data structure is quite different - flat array of bytes vs map of register index to byte array.  It feels like two implementations to me (otherwise every function will be and `if ... else ...` block).

edit: The other thing is the tests directly access the 'Register' member of the structs, so not sure how that would be resolved since the tests assume different data types for that member.

aykevl:
I think the contiguous design is actually better than the map design, even for existing tests. It's relatively easy to read/write 16 bit values to an array of 8-bit values (just some shifts and masks) and it maps closer to the hardware it should mock. You can use a single array to store the data (`[256]byte`) and provide helper functions for setting/reading 16 bit addresses and even a special constructor that maps a `map[uint8]uint16` of values to this array. Memory consumption should not be a concern as it's just 256 bytes.

Maybe it's easier to do this change in a separate PR.

kenbell:
OK, I see what you're saying.  Could possibly do it by having a 'stride' member, which acts as a multiplier on the register number for ReadRegister / WriteRegister.

I'll raise a separate PR to tweak the mock.

aykevl:
Oops I've looked at the datasheet and now I better understand the problem. Apparently the INA260 stores 16 bit values for each address. That's quite unique, I haven't seen any other I2C device like this. Usually they have 8 bit registers and split 16 bit values over two registers instead.

> OK, I see what you're saying. Could possibly do it by having a 'stride' member, which acts as a multiplier on the register number for ReadRegister / WriteRegister.

Yes, that's one solution. But now that I see the INA260 is different from other I2C devices, it makes a bit more sense to introduce a new kind of I2C device (although I don't like the code duplication).

One thing I don't really agree with though is the naming. You've named 8-bit devices `I2CContiguousRegisterDevice` and 16-bit devices `I2CClassicDevice`. That makes it sound like `I2CClassicDevice` is the more common form, which it is not (as far as I know). If you keep them in one struct, I would suggest keeping `I2CContiguousRegisterDevice` as `I2CDevice` and naming the 16-bit version something different, maybe `I2CDevice16`?

What might be even better (but IMHO not necessary just to get the INA260 merged) is this: splitting the data store and the I2C mock interface. The data store could implement just the `Tx` method in whatever way the device wants, and the `I2CDevice` object could then implement `ReadRegister` and `WriteRegister` on top of it.

kenbell:
I was already most of the way through doing a stride implementation, so I've raised PRs for a couple of alternative options:
#243 - Single mock device with variable stride
#244 - Two mock implementations with a common interface (for the bus to use)

Happy to go with whichever is preferred.

kenbell:
I've got another i2c device i want to add that does not have 'registers', but has variable-length 'commands' - so I think #244 is the best approach to introduce an interface so additional implementations can be provided.

I've rebased this onto #244

deadprogram:
@kenbell I think this branch is going to need to be rebased again against `dev`, since I merged #244 first.

kenbell:
@deadprogram - thanks!  I've rebased.

ina260/ina260.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Multiple of MODE_XXXX
+	Mode byte
+}
+
+// New creates a new INA260 connection. The I2C bus must already be
+// configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func New(bus drivers.I2C, address uint16) Device {
```
deadprogram: I think it is better to define this as `func New(bus drivers.I2C) Device` and then just set `Address` to special value only if needed before using. See https://github.com/tinygo-org/drivers/blob/release/adt7410/adt7410.go#L39-L45

kenbell: That makes sense - pushed an update

deadprogram:
Great work on this @kenbell thank you very much! Now merging.

PR #244, opened 2021-03-31
Add 16-bit register mock device

kenbell:
Introduces an interface to abstract the device implementation (needed by the bus) and has separate implementations for 8-bit and 16-bit device mocks.


deadprogram:
Seems like all feedback so far has been addressed.

@aykevl what do you think of the changes?

deadprogram:
Looked again, and it for sure seems like it addresses the core issues around testing i2c.

Now merging, thanks for working on this @kenbell

PR #248, opened 2021-04-03
drivers/flash: fix EraseBlocks method which is erasing sectors instead

dmlambea:
Currently, the call to EraseBlocks is iterating from the starting block and calling EraseSector for every index, which will surely erase the wrong areas.

This PR fixes it by calling EraseBlock instead.


sago35 (approved):
Thank you for your contribution.
I have confirmed that it correctly erases the data in block (64 * 1024 bytes).

deadprogram:
Thank you @dmlambea for the fix and @sago35 for verifying.

Now merging.

PR #251, opened 2021-04-14
tone: add package for producing tones using the PWM interface

aykevl:
I'm only really familiar with Dutch tones and prononctuation, but I tried to use an English equivalent. Here, two-tone sirens are standard for all emergency vehicles.

Warning: the Circuit Playground Bluefruit is surprisingly loud!


deadprogram:
Tested and it is in fact quite loud! :speaker: 

I would like to merge this, but it will fail the CI until #252 is merged. :smiley_cat:

deadprogram:
Now merging, thanks for the new driver package @aykevl

PR #252, opened 2021-04-17
pwm: update drivers with PWM to use new interface

deadprogram:
Now that https://github.com/tinygo-org/tinygo/pull/1121 has been merged into TinyGo, this PR is needed to handle the changes to the new PWM API.

### DIFF
--- a/examples/apa102/itsybitsy-m0/main.go
+++ b/examples/apa102/itsybitsy-m0/main.go
@@ -15,7 +15,7 @@ import (
 var (
 	apa apa102.Device
 
-	led   = machine.PWM{machine.LED}
+	pwm   = machine.TCC0
 	leds  = make([]color.RGBA, 1)
 	wheel = &Wheel{Brightness: 0x10}
 )
@@ -27,12 +27,19 @@ func init() {
 	apa = apa102.NewSoftwareSPI(machine.PA00, machine.PA01, 1)
 
 	// Configure the regular on-board LED for PWM fading
-	machine.InitPWM()
-	led.Configure()
-
+	err := pwm.Configure(machine.PWMConfig{})
+	if err != nil {
+		println("failed to configure PWM")
+		return
+	}
 }
 
 func main() {
+	channelLED, err := pwm.Channel(machine.LED)
+	if err != nil {
+		println("failed to configure LED PWM channel")
+		return
+	}
 
 	// We'll fade the on-board LED in a goroutine to show/ensure that the APA102
 	// works fine with the scheduler enabled.  Comment this out to test this code
@@ -43,11 +50,11 @@ func main() {
 				brightening = !brightening
 				continue
 			}
-			var brightness uint16 = uint16(i) << 8
+			var brightness uint32 = uint32(i)
 			if !brightening {
-				brightness = 0xFFFF - brightness
+				brightness = 256 - brightness
 			}
-			led.Set(brightness)
+			pwm.Set(channelLED, pwm.Top()*brightness/256)
 			time.Sleep(5 * time.Millisecond)
 		}
 	}()
--- a/l293x/l293x.go
+++ b/l293x/l293x.go
@@ -56,49 +56,71 @@ func (d *Device) Stop() {
 	d.en.Low()
 }
 
+// PWM is the interface necessary for controlling the motor driver.
+type PWM interface {
+	Configure(config machine.PWMConfig) error
+	Channel(pin machine.Pin) (channel uint8, err error)
+	Top() uint32
+	Set(channel uint8, value uint32)
+	SetPeriod(period uint64) error
+}
+
 // PWMDevice is a motor with speed control.
 // a1 and a2 are the directional GPIO pins.
 // en is the PWM pin that controls the motor speed.
 type PWMDevice struct {
 	a1, a2 machine.Pin
-	en     machine.PWM
+	spc    uint8
+	pwm    PWM
 }
 
-// NewWithSpeed returns a new PWMMotor driver that uses a PWM pin to control speed.
-func NewWithSpeed(direction1, direction2 machine.Pin, speedPin machine.PWM) PWMDevice {
+// NewWithSpeed returns a new PWMMotor driver that uses an already configured PWM channel
+// to control speed.
+func NewWithSpeed(direction1, direction2 machine.Pin, spc uint8, pwm PWM) PWMDevice {
 	return PWMDevice{
-		a1: direction1,
-		a2: direction2,
-		en: speedPin,
+		a1:  direction1,
+		a2:  direction2,
+		spc: spc,
+		pwm: pwm,
 	}
 }
 
-// Configure configures the PWMDevice.
-func (d *PWMDevice) Configure() {
+// Configure configures the PWMDevice. Note that the PWM interface and
+// channel must already be configured, this function will not do it for you.
+func (d *PWMDevice) Configure() error {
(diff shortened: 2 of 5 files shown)


examples/apa102/itsybitsy-m0/main.go
```diff
@@ -47,7 +56,7 @@ func main() {
 			if !brightening {
 				brightness = 0xFFFF - brightness
 			}
-			led.Set(brightness)
+			pwm.Set(channelLED, uint32(brightness)/0xFFFF)
```
aykevl: If you want to set the PWM to a certain fraction, you need to use the `Top()` method. For example:

```suggestion
			pwm.Set(channelLED, pwm.Top() * 256 / brightness)
```

This also requires removing `uint16(i) << 8`: there is no need for increasing the range here.

l293x/l293x.go
```diff
@@ -61,44 +61,67 @@ func (d *Device) Stop() {
 // en is the PWM pin that controls the motor speed.
 type PWMDevice struct {
 	a1, a2 machine.Pin
-	en     machine.PWM
+	en     machine.Pin
+	pwm    *machine.TCC
```
aykevl: To make this driver portable, the l293x package should define a new `PWM` interface and use that. For an example, see:

https://github.com/tinygo-org/drivers/blob/8d776b9ebbf51f7b0b761e883394a5cf3553fc74/tone/tone.go#L7-L14

l293x/l293x.go
```diff
@@ -61,44 +61,67 @@ func (d *Device) Stop() {
...
 }
 
 // Configure configures the PWMDevice.
-func (d *PWMDevice) Configure() {
+func (d *PWMDevice) Configure() error {
 	d.a1.Configure(machine.PinConfig{Mode: machine.PinOutput})
 	d.a2.Configure(machine.PinConfig{Mode: machine.PinOutput})
-	d.en.Configure()
+	err := d.pwm.Configure(machine.PWMConfig{
+		Period: 16384e3, // 16.384ms
```
aykevl: Does driving motors depend on this particular frequency?
If not, a more portable solution would be to let users of this API provide an already configured PWM interface object and channel number so that it is easy to drive multiple motors from a single PWM if the PWM has multiple channels.

l9110x/l9110x.go
```diff
@@ -52,39 +52,59 @@ func (d *Device) Stop() {
 // PWMDevice is a motor with speed control.
 // ia and ib are the directional/speed PWM pins.
 type PWMDevice struct {
-	ia, ib machine.PWM
+	ia, ib machine.Pin
+	pwm    *machine.TCC
```
aykevl: Same here.

examples/apa102/itsybitsy-m0/main.go
deadprogram: I think you meant `pwm.Set(channelLED, pwm.Top()*brightness/256)`.

Anyhow changes made and force pushed to branch.

l293x/l293x.go
deadprogram: Done.

l293x/l293x.go
deadprogram: Done.

l9110x/l9110x.go
deadprogram: Also done.

deadprogram:
Updated and ready for review @aykevl

deadprogram:
Reminder again that this PR is ready for review. :smile_cat:

l293x/l293x.go
```diff
@@ -56,49 +56,71 @@ func (d *Device) Stop() {
...
 
+// PWM is the interface necessary for controlling the motor driver.
+type PWM interface {
+	Configure(config machine.PWMConfig) error
+	Channel(pin machine.Pin) (channel uint8, err error)
+	Top() uint32
+	Set(channel uint8, value uint32)
+	SetPeriod(period uint64) error
+}
+
```
sago35, line 67: It might be better to move `type PWM interface` to `./pwm.go` .

l9110x/l9110x.go
```diff
@@ -49,42 +49,52 @@ func (d *Device) Stop() {
...
 
+// PWM is the interface necessary for controlling the motor driver.
+type PWM interface {
+	Configure(config machine.PWMConfig) error
+	Channel(pin machine.Pin) (channel uint8, err error)
+	Top() uint32
+	Set(channel uint8, value uint32)
+	SetPeriod(period uint64) error
+}
+
```
sago35, line 60: same here.

sago35:
Other than the above, I think the rest is good.

examples/apa102/itsybitsy-m0/main.go
```diff
@@ -27,12 +27,21 @@ func init() {
 	apa = apa102.NewSoftwareSPI(machine.PA00, machine.PA01, 1)
 
 	// Configure the regular on-board LED for PWM fading
-	machine.InitPWM()
-	led.Configure()
-
+	err := pwm.Configure(machine.PWMConfig{
+		Period: 16384e3, // 16.384ms
```
aykevl: You could consider leaving this out, as the default value of `Period` is explicitly defined as being a good value for LEDs. So `machine.PWMConfig{}` is good enough.

l293x/l293x.go
aykevl: No, I think it's better to keep it in the driver. The reason is that (unlike SPI or I2C), the capabilities of each PWM vary. For example: the tone driver (https://github.com/tinygo-org/drivers/pull/251) requires a `SetPeriod` method while the servo driver (https://github.com/tinygo-org/drivers/compare/dev...servo) doesn't need this method. If there is a PWM that does not support the `SetPeriod` method, it can still be used for a servo. If the `PWM` interface is shared between all drivers, such a PWM (without SetPeriod) would not be usable by any driver.

That said, most of the methods here are unused so might be removed from this interface (for example, `SetPeriod` is unused).

l293x/l293x.go
```diff
@@ -56,49 +56,71 @@ func (d *Device) Stop() {
...
+// Forward turns motor on in forward direction at specific speed as a percentage.
+func (d *PWMDevice) Forward(speed uint32) {
+	if speed > 100 {
+		speed = 100
+	}
+
 	d.a1.High()
 	d.a2.Low()
-	d.en.Set(speed)
+	d.pwm.Set(d.spc, speed)
```
aykevl: The channel value is not a percentage. I think you meant something like this:

```suggestion
	d.pwm.Set(d.spc, d.pwm.Top() * speed / 100)
```

(The same goes for other `Forwards` and `Backwards` calls.)

l293x/l293x.go
sago35: I understood.

deadprogram: Also, at least for now, none of the interfaces defined in this package have any reference to the `machine` package. This is what makes it possible to run the unit tests just using "standard" Go.

examples/apa102/itsybitsy-m0/main.go
deadprogram: Done.

l293x/l293x.go
deadprogram: OK, that makes total sense, thanks for the correction. 

Done.

deadprogram:
OK, it looks like finally NOW this is done. Ready again for review.

deadprogram:
All feedback has been addressed, so now merging. Thanks @aykevl for the reviews.

PR #253, opened 2021-04-25
ENC28J60 driver working with blinky example

### LINKED ISSUE #238: Support ENC28J60 ethernet card
[the IC](https://www.microchip.com/wwwproducts/en/en022889).

If no one is working on implementing it, I'll do it! I've already done work to port https://github.com/muanis/arduino-projects, specifically [this file](https://github.com/muanis/arduino-projects/blob/master/libraries/etherShield/enc28j60.c).

I'm just a little unsure how testing works with TinyGo.

soypat:
Resolves #238 
Examples in [`examples_test.go`](https://github.com/soypat/drivers/blob/dev/enc28j60/examples_test.go).

There are some things I'd like to make better, for example instead of using a buffer to store the entire ethernet frame, a `io.ReaderCloser` and `io.Writer` + `io.Copy` could be used to marshal and unmarshal frames. This could reduce RAM usage from numbers like 500-1500 range to 32 bytes.

Driver folders I've added:
* [`enc28j60`](https://github.com/soypat/drivers/tree/dev/enc28j60) I/O, Configuring and routines for ENC28J60 IC
* [`bytealg`](https://github.com/soypat/drivers/tree/dev/bytealg) Needed a strings.Index for bytes with minimal memory allocation
*  [`frame`](https://github.com/soypat/drivers/tree/dev/frame) Need an abstraction for Marsahlling & Unmarshalling ethernet/ip/tcp frames.
* [`encoding/hex`](https://github.com/soypat/drivers/tree/dev/encoding/hex) Primitives to convert/print binary byte -> hex bytes 

Driver folders edited
* [`net`](https://github.com/soypat/drivers/tree/dev/net) Added `HardwareAddress` type.


soypat:
Still working on a improved version at https://github.com/soypat/drivers/tree/swtch. 

Basically brings the following to the table:
* State machine approach which is much slimmer on stack usage
* No more global buffer, all data is read from ENC28J60 directly to the IP/Ethernet/TCP frame


Still trying to work out some kinks in it.

deadprogram:
This is some really important work. Let us all know anything we can do to help out!

soypat:
@deadprogram Thanks for the supportive message :smile: ! I've uploaded my work so far here https://github.com/soypat/ether-swtch. I'm struggling on how the tests would look like for the package. Once I got tests, the rest should be streamlined. I'd also be inclined to asking what everyone thinks about the API and if there are better ideas out there. I've rewritten it a couple of times but there's always something that ends up bothering me.

soypat:
Scratch the above comment. I'm getting somewhere. Peter bourgon gave me some excellent tips on the #networking slack channel. porting old driver to new api @ https://github.com/soypat/enc28j60

soypat:
Please take a look at https://github.com/soypat/enc28j60. The new API is ready. A few issues with it's current implementation:

* Depends on github.com/soypat/net until https://github.com/tinygo-org/tinygo/pull/1975 is merged into `main`
* Depends on github.com/soypat/ether-swtch external dependency for Ethernet/IP/TCP logic

Should we wait for #1975 before I submit a new pull request with this driver? Should [`ether-swtch`](https://github.com/soypat/ether-swtch) be part of tinygo drivers, for the time being I am the only maintainer.

Below is an example on the usage of `enc28j60`+`ether-swtch` libraries:
```go
	var (
                spiCS = machine.D53
		MAC  = net.HardwareAddr{0xDE, 0xAD, 0xBE, 0xEF, 0xFE, 0xFF}
		MyIP = net.IP{192, 168, 1, 5} //static setup is the only one available
	)
	// 8MHz SPI clk for older than Rev 6 boards (See Rev. B4 Silicon Errata)
	machine.SPI0.Configure(machine.SPIConfig{Frequency: 8e6})

	e := enc28j60.New(spiCS, machine.SPI0)

	err := e.Init(MAC)
	if err != nil {
		println(err.Error())
	}
	const okHeader = "HTTP/1.0 200 OK\r\nContent-Type: text/html\r\nPragma: no-cache\r\n\r\n"

	timeout := time.Second * 1
	swtch.HTTPListenAndServe(e, MAC, MyIP, timeout, func(URL []byte) (response []byte) {
		return []byte(okHeader + "Hello world!")
	}, panic)
```

PR #256, opened 2021-05-08
Enable reset screen for SSD1306 via I2C

ysoldak:
This enables use of 128x64 OLED screen on XIAO Expansion Board (SSD1306 driver) w/o weird jump of (0,0) point on reset.

Code that I tested it with: https://github.com/ysoldak/tinygo-stuff/blob/master/examples/xiao-expansion-board/oled-snake/oled-snake.go

I've tried resetting display 200 times (see commented out block there) and experienced no buffer corruption or any other bad effects (as comment for SPI warns about). So I assume it's safe to do resets while on I2C.

Weird thing though... Screen is 128x64 and Seeeduino's docs say address `0x78` shall be used, but it works handsomely with `0x3C` that is "Address_128_32" in `registers.go`

### DIFF
--- a/examples/ssd1306/i2c_128x64/main.go
+++ b/examples/ssd1306/i2c_128x64/main.go
@@ -1,0 +1,60 @@
+// This example shows how to use 128x64 display over I2C
+// Tested on Seeeduino XIAO Expansion Board https://wiki.seeedstudio.com/Seeeduino-XIAO-Expansion-Board/
+//
+// According to manual, I2C address of the display is 0x78, but that's 8-bit address.
+// TinyGo operates on 7-bit addresses and respective 7-bit address would be 0x3C, which we use below.
+//
+// To learn more about different types of I2C addresses, please see following page
+// https://www.totalphase.com/support/articles/200349176-7-bit-8-bit-and-10-bit-I2C-Slave-Addressing
+
+package main
+
+import (
+	"machine"
+
+	"image/color"
+	"time"
+
+	"tinygo.org/x/drivers/ssd1306"
+)
+
+func main() {
+	machine.I2C0.Configure(machine.I2CConfig{
+		Frequency: machine.TWI_FREQ_400KHZ,
+	})
+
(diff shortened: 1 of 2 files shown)


ysoldak:
The document mentioning "The I2C slave address is 0x78."
https://files.seeedstudio.com/wiki/Seeeduino-XIAO-Expansion-Board/document/Seeeduino%20XIAO%20Expansion%20board_v1.0_SCH_200824.pdf

From https://wiki.seeedstudio.com/Seeeduino-XIAO-Expansion-Board/

sago35 (approved):
Looks good to me.
Thank you for contribution.

sago35:
> Weird thing though... Screen is 128x64 and Seeeduino's docs say address 0x78 shall be used, but it works handsomely with 0x3C that is "Address_128_32" in registers.go

I think 0x78 is an 8-bit address notation.
As a 7-bit address, it would be 0x3C, so I don't think it's wrong.

https://www.totalphase.com/support/articles/200349176-7-bit-8-bit-and-10-bit-I2C-Slave-Addressing

The following is the source code for tinygo.

https://github.com/tinygo-org/tinygo/blob/release/src/machine/i2c.go#L26

```go
// WriteRegister transmits first the register and then the data to the
// peripheral device.
//
// Many I2C-compatible devices are organized in terms of registers. This method
// is a shortcut to easily write to such registers. Also, it only works for
// devices with 7-bit addresses, which is the vast majority.
func (i2c *I2C) WriteRegister(address uint8, register uint8, data []byte) error {
	buf := make([]uint8, len(data)+1)
	buf[0] = register
	copy(buf[1:], data)
	return i2c.Tx(uint16(address), buf, nil)
}
```

sago35:
Please register `examples/ssd1306/i2c_128x64` .
This will make it easier for people who use the XIAO Expansion Board to try it out.

ysoldak:
@sago35 @deadprogram added example

deadprogram:
Thanks for adding this @ysoldak now merging.

PR #257, opened 2021-05-09
pcf8563: add support for pcf8563 real time clock

sago35:
Created a driver for the PF8563 RTC used in the XIAO-Expansion-board.

https://www.seeedstudio.com/Seeeduino-XIAO-Expansion-board-p-4746.html

`SetTimer()` and `SetAlarm()` are a bit limited in their functionality.
If additional functionality is needed, it will be created in another PR or something.

## PCF8563

> The PCF8563 is a CMOS1 Real-Time Clock (RTC) and calendar optimized for low power consumption. A programmable clock output, interrupt output, and voltage-low detector are also provided. All addresses and data are transferred serially via a two-line bidirectional I2C-bus. Maximum bus speed is 400 kbit/s. The register address is incremented automatically after each written or read data byte

https://www.nxp.com/docs/en/data-sheet/PCF8563.pdf


pcf8563/pcf8563.go
```diff
@@ -0,0 +1,249 @@
...
+// New creates a new PCF8563 connection. I2C bus must be already configured.
+func New(i2c drivers.I2C) Device {
+	return Device{
+		bus:     i2c,
+		Address: PCF8563_ADDR,
+	}
+}
+
+// Reset resets the `control and status registers`.
+func (d *Device) Reset() (err error) {
```
deadprogram, line 32: Should this be renamed to `Configure()`? or should there perhaps be a `Configure()` function added that calls this function?

examples/pcf8563/time/main.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+var (
+	i2c = machine.I2C0
+	rtc = pcf8563.New(i2c)
+)
+
+func main() {
+	i2c.Configure(machine.I2CConfig{Frequency: machine.TWI_FREQ_400KHZ})
+
```
deadprogram, line 19: Is this missing a call to `rtc.Reset()` here?

pcf8563/pcf8563.go
sago35: Calling `Reset()`, for example, will initialize Alarm and Timer.

The RTC may be battery-backed even if the microcontroller is powered off. If the RTC is battery-backed, the user will want to use the previous Alarm and Timer, so putting `Reset()` in `Configure()` will not work.
`Reset()` and `Configure()` also have different meanings.

So I think it's good to leave it as it is now. Whether or not to call `Reset()` depends on how to use it.

examples/pcf8563/time/main.go
sago35: Here, `Reset()` may or may not be called, but this example may make some people wonder.
I will add a call to the `Reset()` function here, and also add some more information in the comments of the `Reset()` function.

deadprogram:
Thanks for the update @sago35 can you please squash commits? Thank you!

deadprogram:
Thanks @sago35 now merging.

PR #234, opened 2021-02-26
Fix Wifnina for Firmware >= 1.2.4

Nerzal:
I know the fix is kinda dirty, but i couldn't wrap my head around how to get it into the existing padding logic. 
If anyone has an idea, tell me :) 

I tested the mqttclient, tcpclient and udpstation and they all worked. 
Only thing i wasn't able to fix is the tls issue, guess I'll need more time for that. 

Tested with Firmware 1.2.4 and 1.4.3 :) 

Fixes: #233

Issue #233, opened 2021-02-25
WIFININA: WLAN connection cannot be established 


Nerzal:
I had some problems maintaining a stable connection with Firmware version 1.2.1, so i tried nearly every other Firmware version i could find. 

The problem that i do now face is, that I'm unable to connect to the WLAN now. Also tried it with the connectToAP() function from the examples, which also does not lead to an established connection anymore. 

To verify, that i did not break anything when flashing a new firmware, i tried the Arduino IDE example sketch, which successfully connects and maintains a connection. 

So does anyone have a trick, what to do?

Example function:

```go
func connectToAP() {
	time.Sleep(2 * time.Second)
	println("Connecting to " + ssid)
	adaptor.SetPassphrase("mySSID", "mySecret")
	for st, _ := adaptor.GetConnectionStatus(); st != wifinina.StatusConnected; {
		println("Connection status: " + st.String())
		time.Sleep(1 * time.Second)
		st, _ = adaptor.GetConnectionStatus()
	}
	println("Connected.")
	time.Sleep(2 * time.Second)
	ip, _, _, err := adaptor.GetIP()
	for ; err != nil; ip, _, _, err = adaptor.GetIP() {
		println(err.Error())
		time.Sleep(1 * time.Second)
	}
	println(ip.String())
}
```

Output: 

```
Connection status: No SSID Available
Connection status: No SSID Available
Connection status: No SSID Available
Connection status: No SSID Available
Connection status: Disconnected
Connection status: Disconnected
Connection status: Disconnected
....
```

Nerzal:
I'm currently trying to analyze whats going on here. 

My router logs arent that helpful, but atleast i can see, that the arduino actually talks to the router. 

```
(W005) WLAN-Anmeldung via <2.4> Ghz Frequenzband ist gescheitert: Autorisierung fehlgeschlagen. Rechnername: <arduino-nano-iot-33>, Mac-Adresse: <24:62:AB:B8:F8:84>
``` 

That roughly translates to: Connection attempt to 2.4 Ghz failed: Authorization failed. 
So i think, that maybe the password isn't being transmitted correctly.

Nerzal:
For documentation purposes so i (and others) can find the interesting changes in the wifinina driver lib for newer firmware versions.

Nerzal:
Closing this issue, as it is fixed :)

deadprogram:
Reopening, will close once released. Has been labeled for "next release".

[...]

PR #260, opened 2021-05-10
Fix WiFiNINA + TLS

ysoldak:
Built on top of @Nerzal fix.
Now it can connect to AP. (See #233 and #234)

Verified Web and NTP client examples work (refactored them a bit on the way too).
Added example for TLS client + respective changes to wifinina driver to make TLS work.

Verified on Arduino Nano 33 IoT and NINA FW version 1.4.5 (latest)
(As of today you can't flash 1.4.5 from Arduino IDE. Hack is to download 1.4.5 binary and drop it in place of 1.4.3)

P.S. I know, nicer version of wifinina driver in works, but at least this can serve the purpose until new version is ready?

### DIFF
--- a/examples/wifinina/ntpclient/main.go
+++ b/examples/wifinina/ntpclient/main.go
@@ -23,33 +23,46 @@ const ntpHost = "129.6.15.29"
 
 const NTP_PACKET_SIZE = 48
 
+// these are the default pins for the Arduino Nano33 IoT.
+// change these to connect to a different UART or pins for the ESP8266/ESP32
 var (
+
+	// these are the default pins for the Arduino Nano33 IoT.
+	uart = machine.UART2
+	tx   = machine.NINA_TX
+	rx   = machine.NINA_RX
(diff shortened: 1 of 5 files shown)


deadprogram:
Just did some testing, and it worked perfectly with the latest BIN file from https://github.com/arduino/nina-fw/releases

BTW the easiest way to update to the latest version is to run esptool.

You do need to put the passthru INO on the board first. After that you then can run the following command:

```shell
esptool --chip esp32 --port /dev/ttyACM0 --baud 115200 --before default_reset --after hard_reset write_flash 0 ~/Downloads/NINA_W102-v1.4.5.bin
```

deadprogram:
This is really great work getting it into at least usable shape. Thank you @ysoldak for taking the initiative to work on results of all the investigation/work done by @Nerzal on top of all the work originally done by @bgould 

Whew!

Now merging.

Nerzal:
This is really great! Thank u!

PR #262, opened 2021-05-11
wifinina: update docs to simplify the nina-fw update process

deadprogram:
This PR updates the docs to simplify the nina-fw update process, for those who need to update to the latest wifi firmware.


deadprogram:
Added a second commit that updated the docs and examples to work with the various Adafruit boards with ESP32 WIFi processors.

deadprogram:
Merging.

PR #263, opened 2021-05-12
Smoke tests refactoring

ysoldak:
Paint me crazy and feel free to reject.

Could not stand that long list of hardcoded smoke tests in the Makefile and embarked on wild refactoring.
(Boosted my Makefile-foo on the way.)

Every example now has `targets.txt` file with list of targets we want to build this example for.
Values extracted from old Makefile.
Some examples (mostly wifinina and espat) been missing targets, added "arduino-nano33".

`targets.txt` file name is random, any better name?

What about that `ifneq ($(AVR), 0)`? Seem to work handsomely w/o.


ysoldak:
I was thinking to use `// +build` headers instead of `targets.txt`.
The problem though is how to infer target from build tags, was not able to solve that.

I don't like `target.txt` files much myself, but at least it's manageable compared with huge list in Makefile.

aykevl:
> I was thinking to use `// +build` headers instead of `targets.txt`.
> The problem though is how to infer target from build tags, was not able to solve that.

Yes, that's exactly why it's not a good idea. There is no way to do that reliably.

I actually quite like the simplicity of having a single Makefile that runs all tests. For example, it allows the flexibility of adding extra command line parameters if you want.

> What about that `ifneq ($(AVR), 0)`? Seem to work handsomely w/o.

Because it's not always easy to run AVR tests on some operating systems: it requires that some extra tools are installed.

ysoldak:
I agree with flexibility.
Keeping eye on all examples are actually buildable (included in Makefile) -- that's hard. If we have such a goal, of course.

I have another approach, w/o `targets.txt` files, will make a separate PR. Will see if you like it better.

ysoldak:
OK, will not open yet another PR.
Instead, here is the diff if you are fancy on how we can make it w/o `target.txt` files, but inferring build targets from build tags in each example `main.go`
https://github.com/tinygo-org/drivers/compare/dev...ysoldak:smoke-tests-tags

Pros:
- Do not need to track individual examples, they are all covered given they have `+build` headers
- Smaller Makefile

Cons:
- Not flexible, one command to build 'em all
- `+build` headers  may prevent people from building examples for their targets (shall build tags be wider then?)
- When I Wrote It, Only God and I Knew the Meaning; Now God Alone Knows

So, not pushing this. As said, feel free to reject. That was an interesting journey to Makefile-land.

PR #264, opened 2021-05-12
servo: add driver using PWM

aykevl:
This PR adds support for controlling servos. For example, on the Arduino Uno it should be able to control up to 6 servos jitter-free when using all available PWM pins.

I haven't added support for setting a position in degrees, mainly because this varies by servo and it's probably necessary to configure the bounds in some way. Therefore, I added just `SetMicroseconds`. This makes the API possible to use and leaves the possibility of adding a `SetPosition` in the future.


deadprogram:
Please resolve merge conflict @aykevl thanks.

aykevl:
Done.

deadprogram:
Thanks @aykevl now merging!

PR #267, opened 2021-05-26
wifinina: fix rssi

ysoldak:
Fixes GetCurrentRSSI() for WiFiNINA driver.

Also, changes in "connect" example:
+ Print current SSID
+ Wait for correct time before printing it out
+ Cleanup


deadprogram:
Thank you for the fix @ysoldak now merging.

PR #268, opened 2021-05-27
wifinina: avoid fmt dependency

ysoldak:
This avoids `fmt` package dependency in wifinina driver, that shrinks compiled flash size significantly.
`fmt` is still used for debug printouts yet happily optimised out by the compiler when debugging is disabled.

Backstory.
I've noticed my program is `70k` big on flash even after "z" optimisation.
According to `-size=full` output, `fmt` package was responsible for quite big chunk of that.
Odd, since I've specifically avoided `fmt` in my code.
Analysis shows `wifinina` depends on `fmt` and after eliminating this dependency, my code is `50k` now.

I had to use `strconv` to replace `fmt`, still the size decrease is significant.
Please tell if there's a better way to convert between bytes and string representation of bytes than `Itoa` and `Atoi`.

Example was updated too in this PR.

Some stats

Old example and old driver
```
tinygo build -target arduino-nano33 -size=full -o build.hex -opt=z .
   code  rodata    data     bss |   flash     ram | package
   ...
   9420     201       0       0 |    9621       0 | fmt
   ...
  55400       -    1384    4380 |   56784    5764 | (all)
```

New example and old driver
```
tinygo build -target arduino-nano33 -size=full -o build.hex -opt=z .
   code  rodata    data     bss |   flash     ram | package
   ...
   9440     201       0       0 |    9641       0 | fmt
   ...
  55496       -    1384    4380 |   56880    5764 | (all)
```

New example and new driver
```
tinygo build -target arduino-nano33 -size=full -o build.hex -opt=z .
   code  rodata    data     bss |   flash     ram | package
   ...
  30332       -    1360    4380 |   31692    5740 | (all)
```

### DIFF
--- a/examples/wifinina/connect/main.go
+++ b/examples/wifinina/connect/main.go
@@ -2,9 +2,8 @@
 package main
 
 import (
-	"encoding/binary"
-	"fmt"
 	"machine"
+	"strconv"
 	"time"
 
 	"tinygo.org/x/drivers/wifinina"
@@ -80,7 +79,7 @@ func printRSSI() {
 		println("Unknown (error: ", err.Error(), ")")
 		return
 	}
-	println(fmt.Sprintf("%d", rssi))
+	println(strconv.Itoa(int(rssi)))
 }
 
 func printIPs() {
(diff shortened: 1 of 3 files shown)


deadprogram:
This is really great @ysoldak thank you so much for making things smaller.

Now merging.

PR #269, opened 2021-05-29
apa102: use 4-byte buffer to improve speed

aykevl:
Instead of sending the APA102 data byte by byte, it's more efficient to send multiple bytes at once (especially when the SPI peripheral uses DMA). This requires the `apa102.Device` object to be a pointer receiver to avoid excessive heap allocations.

This commit also just happens to work around a hardware bug on the nrf52832:
https://infocenter.nordicsemi.com/index.jsp?topic=%2Ferrata_nRF52832_Rev2%2FERR%2FnRF52832%2FRev2%2Flatest%2Fanomaly_832_58.html&anchor=anomaly_832_58

### DIFF
--- a/apa102/apa102.go
+++ b/apa102/apa102.go
@@ -27,44 +27,46 @@ var startFrame = []byte{0x00, 0x00, 0x00, 0x00}
 type Device struct {
 	bus   drivers.SPI
 	Order int
+	buf   [4]byte
 }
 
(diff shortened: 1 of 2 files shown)


deadprogram:
:traffic_light: :traffic_light: :traffic_light: :traffic_light: 

Merging, thanks @aykevl

PR #270, opened 2021-05-30
bmi160: avoid heap allocations

aykevl:
The nrf52 series chips use DMA for writing and reading SPI data, which means that the transmit and receive buffer slices escape. This can easily be solved by pre-allocating the buffer in the BMI160 driver and reusing this buffer every time.

### DIFF
--- a/bmi160/bmi160.go
+++ b/bmi160/bmi160.go
@@ -13,6 +13,8 @@ type DeviceSPI struct {
 	// Chip select pin
 	CSB machine.Pin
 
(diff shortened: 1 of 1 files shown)


deadprogram:
I'm sure if we look thru all the drivers, there must be some others with this same sort of heap allocation we can correct.

Now merging, thanks @aykevl

aykevl:
Yes, probably. But this is one I hit in a project and made a PR for so that I don't need to worry about it anymore :)

Issue #204, opened 2020-10-09
ws2812 driver and interrupts


ewwaller:
The ws2812 on an ARM Cortex M4 [Adafruit Feather M4] seems to require that interrupts be disabled during the write.  Without the guard, the data sent to the LED controller is occasionally garbled.   I disabled interrupts just before the in line assembly WriteByte in ws2812_m4_120m.go, and re-enabled them immediately after.  The LEDs no longer displayed intermittent flashes (visible when the LED is continuously updated at 500Hz).  As a sanity check, I ensured I was able to accurately catch 10hKz pulse trains on an edge triggered interrupt with the LED updating with Interrupt disable without dropping any interrupts.

Also, the timing is a little wrong -- to hit the pulsewidths in the spec sheet, I had to remove a few NOPs from the initial high period (8, IIRC)

deadprogram:
That sounds very interesting. Are you going to submit a PR with those changes, possibly?

ewwaller:
I will be happy to.  I'm a little leary as I am still a newcomer to both Go and Tinygo; but I've much experience with embedded processors.  Let me recheck the timing this weekend, and will create a pull request.

aykevl:
Previously this was difficult, but now we have the runtime/interrupt package that you can use for this purpose.

deadprogram:
This has been released in 0.17.1. Thanks!

PR #271, opened 2021-05-30
WS2812: Disable interrupts before sending data

dbridges:
This PR disables interrupts on M0 and M4 processors before sending data to the WS2812. Interrupts are re-enabled after all data is sent.

Currently the WS2812 driver causes intermittent flashes on incorrect LEDs in the strip (tested on a Circuit Playground Express, and also reported by @ewwaller in this [issue](https://github.com/tinygo-org/drivers/issues/204)).

Disabling interrupts solved this problem on a Circuit Playground Express, but I am unable to test on other boards. Because @ewwaller also reported this issue and fix for an M4 board, I went ahead and added this code to all M0 and M4 processors. Also, the Adafruit NeoPixel Arduino library disables interrupts as well.

### DIFF
--- a/ws2812/ws2812_avr.go
+++ b/ws2812/ws2812_avr.go
@@ -7,6 +7,7 @@ package ws2812
 import (
 	"device/avr"
 	"machine"
+	"runtime/interrupt"
 )
 
 // Send a single byte using the WS2812 protocol.
@@ -17,6 +18,7 @@ func (d Device) WriteByte(c byte) error {
 	// Probably this is about pointer registers, which are very limited on AVR.
 	port, maskSet := d.Pin.PortMaskSet()
 	_, maskClear := d.Pin.PortMaskClear()
+	mask := interrupt.Disable()
 
 	switch machine.CPUFrequency() {
 	case 16e6: // 16MHz
@@ -51,8 +53,10 @@ func (d Device) WriteByte(c byte) error {
 			"maskClear": maskClear,
 			"portClear": port,
 		})
+		interrupt.Restore(mask)
 		return nil
 	default:
+		interrupt.Restore(mask)
 		return errUnknownClockSpeed
 	}
 }
--- a/ws2812/ws2812_m0_16m.go
+++ b/ws2812/ws2812_m0_16m.go
@@ -7,6 +7,7 @@ package ws2812
 
 import (
 	"device/arm"
+	"runtime/interrupt"
 )
 
 // Send a single byte using the WS2812 protocol.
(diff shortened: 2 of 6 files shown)


aykevl:
Hmmm, yeah maybe the driver itself should disable interrupts because it is so timing critical. However note that this could have unexpected consequences.

That said, the correct way to disable interrupts is in this way:

```go
import "runtime/interrupt"

// ...
mask := interrupt.Disable()
// do something
interrupt.Restore(mask)
```

This takes care of all the various platform specific oddities. For example, on nrf chips it's not allowed to call `arm.DisableInterrupts()` when the SoftDevice is enabled (not currently implemented in `runtime/interrupt`, but will eventually be).

dbridges:
I agree disabling interrupts could have unexpected consequences, but the driver as is emits a lot of unexpected flashes.

@aykevl I can revise the PR to use `runtime/interrupt`. Do you suggest keeping the interrupt disable in `WriteByte` (since it is exported), or instead just disabling in `Write` and `WriteColors` to avoid repeatedly enabling and disabling the interrupts?

aykevl:
> Do you suggest keeping the interrupt disable in `WriteByte` (since it is exported), or instead just disabling in `Write` and `WriteColors` to avoid repeatedly enabling and disabling the interrupts?

I'm not entirely sure either. Doing it for the whole string at once is faster (not sure by how much) but has the downside of keeping interrupts disabled for a long period, which might interfere with other activities. Doing it per byte doesn't have this issue (interrupts can still run, just a little bit delayed) but is a bit slower and has the potential of causing the same issue if an interrupt takes more than ~6µs (see [here](https://wp.josh.com/2014/05/13/ws2812-neopixels-are-not-so-finicky-once-you-get-to-know-them/)).

Perhaps we can start by disabling interrupts for each byte and see how well that goes? I don't think it will be _that_ big of an overhead. Although, if you want to measure it, that would be great!

dbridges:
@aykevl Ready for re-review.

deadprogram:
Looks good to me. @aykevl any other feedback before I merge?

aykevl:
Yes, looks good to me. I'm only a bit uncertain about `defer`: it may be a bit too expensive. For these simple cases, it may be better to just insert the call at the end of the block.

dbridges:
I originally had the `defer` in there because some architectures used `switch` statements, but commit has been amended to remove `defer`.

deadprogram:
Thanks for make the requested changes @dbridges and thanks for reviewing @aykevl now merging.

PR #274, opened 2021-06-10
ili9341: add standard SPI driver

sago35:
This PR adds a standard SPI driver for ili9341.
Unlike the approach in PR #242 , we tried to use the driver dedicated to each board if available.
For example, if there is no board-specific driver, such as the NRF52840, then the standard SPI driver is used.

The frame rate for `examples/ili9341/pyportal_boaring` is as follows

* specific driver
    * wioterminal (atsamd51)
        * 42 fps
* std-spi-driver
    * feather-nrf52840 (nrf52840)
        * 23 fps
    * feather-stm32f405 (stm32f405)
        * 30 fps

According to #196, the spi driver for stm32f405 seems to have a speed of 68 - 88 fps.
Once #196 is merged, the ili9341 on STM32F405 will run faster.
Until then, we can use the std-spi-driver.

* https://github.com/tinygo-org/drivers/pull/196#discussion_r487686223

### DIFF
--- a/examples/ili9341/pyportal_boing/main.go
+++ b/examples/ili9341/pyportal_boing/main.go
@@ -25,7 +25,7 @@ const (
 )
 
 var (
-	frameBuffer = [(graphics.BALLHEIGHT + 8) * (graphics.BALLWIDTH + 8)]uint16{}
+	frameBuffer = [(graphics.BALLHEIGHT + 8) * (graphics.BALLWIDTH + 8) * 2]uint8{}
 
 	startTime int64
 	frame     int64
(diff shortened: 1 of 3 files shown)


sago35:
Note that std-spi-driver does not differ in speed from specific-driver when used with atsamd51.

sago35:
I have fixed the commit message and force-pushed it.

sago35:
The following code, created by @ardnew , still works now.
With `-opt=2`, the fps is 68 fps.
After we merge this PR, it' s a good idea to use the SPI driver from #196 for the STM32F405.

 #196

deadprogram:
Thanks @sago35 worked correctly on my PyPortal. My frame rate was 37-38 fps:

```
tinygo flash -target pyportal ./examples/ili9341/pyportal_boing/
```

Then adding `-opt=2` I got 47-48 fps

deadprogram:
Now merging, thanks for working on this.

PR #279, opened 2021-06-23
sdcard: add write functionality to examples/sdcard/tinyfs

sago35:
Update tinyfs to add write functionality.

See :  https://github.com/tinygo-org/tinyfs/pull/5

I have also implemented the help command in examples/sdcard/tinyfs.
This change is not originally necessary, but I added it because I think the write command is confusing.


deadprogram:
Thanks @sago35 now merging.

PR #280, opened 2021-06-26
ws2812: rename the pin to ws2812

sago35:
Changed the name of the pin from NEOPIXELS to WS2812.
This means that most boards with WS2812 can try examples without changing the source code.
I tried it with feather-nrf52840-sense, and it worked fine!

see.
https://github.com/tinygo-org/tinygo/pull/1940


aykevl:
Thank you! This change is a good idea.

Issue #282, opened 2021-07-02
RTL8720DN: ConnectToAP() does not proceed as expected


deadprogram:
When using the examples for the RTL8720DN and the `debug=true`, the program does not call `ConnectToAP()` until the serial port from the computer connects to the board, and then does not proceed with the `ConnectToAP()` call until the user disconnects the serial connection. This is very odd behavior.

sago35:
WaitSerial() is executed only when debug=true, but this is undocumented and confusing to the user.
I will figure out how to fix it.
I think it would be better to put it in the documentation first.

https://github.com/tinygo-org/drivers/blob/release/examples/rtl8720dn/webclient-tinyterm/wioterminal.go#L54-L56

sago35:
https://www.twitch.tv/videos/1074459848?t=01h09m58s

In your video, you were unable to proceed from rpc_wifi_connect().
In my environment, if I wait 5 seconds, I get a response (rx).
In your video, you terminated picocom, but what happens if you don't terminate it?

deadprogram:
This was already addressed I think, so now closing.

PR #283, opened 2021-07-04
rtl8720dn: add info when debug mode is enabled

sago35:
This PR fixes issue #282


deadprogram:
That is certainly an improvement. Thanks @sago35 now merging.

PR #284, opened 2021-07-06
add first revision of new ENC28J60 driver

soypat:
Rewrote ENC28J60 driver with more go-like approach to packet reading. I left the arduino-inspired functions PacketSend and PacketRead in case someone was looking to directly port some arduino program to Go.

### DIFF
--- a/datagrammer.go
+++ b/datagrammer.go
@@ -1,0 +1,36 @@
+package drivers
+
+import (
+	"io"
+	"time"
+)
+
+// Datagrammer represents a reader/writer of data packets received over stream.
+// These packets are low level representations of what could be an Ethernet/IP/TCP transaction.
+// An example of an IC which implements this is the ENC28J60.
+type Datagrammer interface {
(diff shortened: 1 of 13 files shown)


soypat:
~I've been thinking, maybe `NextPacket` should have a deadline/timeout argument as it probably should not hang indefinitely waiting on a packet.~ Done. It'd be great if people with non-AVR boards could test a simple HTTP server (like in the examples in the test file) and see if it is stable. The only boards I have are arduinos and Picos (no SPI on them). With the arduino the server works fine for some time and then panics due to nil pointer deref.

soypat:
Seems to work OK. Still gotta work on the TCP control flow which hangs on not fulfilled requests (not part of this PR).

I could use suggestions on which functions to inline with `//go:inline` pragma. I'm guessing the I/O should be all inlined for starters.

soypat:
Alright. I feel like the bugs have been ironed out. Please take a look @aykevl

PR #285, opened 2021-07-08
rtl8720dn: add examples/rtl8720dn/webserver

sago35:
I have created a webserver example using wioterminal and rtl8720dn.

image

Many of the files under tinygo.org/drivers/net/http have been adapted from the original Go code.
Notice that examples/rtl8720dn/webserver/main.go is almost identical to the Go code.

https://github.com/tinygo-org/drivers/blob/048d6975b0eff32c727904f23f4a109e76548871/examples/rtl8720dn/webserver/main.go#L59-L69


bgould:
That is awesome.  The `net/http` package is generic such that it can be used on top of espat or wifinina or another implementation, is that correct?

sago35:
At the moment, I think the interface definition is not good.
However, it can be made into a webserver by implementing ListenAndServe().

https://github.com/tinygo-org/drivers/blob/048d6975b0eff32c727904f23f4a109e76548871/net/http/driver.go#L3-L5

It could probably be split into smaller interfaces than ListenAndServe(), but I haven't considered it yet.
Maybe "`net.Conn` + Accept()" would be nice.

https://github.com/tinygo-org/drivers/blob/048d6975b0eff32c727904f23f4a109e76548871/rtl8720dn/http.go#L196-L217

sago35:
I did some more research.
It can be made into a webserver by implementing the following

* Implement ListenAndServe()
* Implement the following for Request
    * https://github.com/tinygo-org/drivers/blob/048d6975b0eff32c727904f23f4a109e76548871/rtl8720dn/http.go#L103
    * https://github.com/tinygo-org/drivers/blob/048d6975b0eff32c727904f23f4a109e76548871/rtl8720dn/http.go#L117
    * https://github.com/tinygo-org/drivers/blob/048d6975b0eff32c727904f23f4a109e76548871/rtl8720dn/http.go#L124

deadprogram:
:spider_web: merging, great work!

PR #287, opened 2021-07-20
rtl8720dn: add support for http.Get(), http.Post() and cookie

sago35:
This PR adds support for the following.

* http.Get() (client and server)
* http.Post() (client and server)
* http.Cookie (client and server)

The number of examples has not increased.
Instead, the webclient and tlsclient examples have been made very simple.
The difference between http and https has been absorbed by the net/http package, so now it is simply a matter of calling http.
Get().
This is written the same way as in regular Go.

The main changes in this PR are as follows
This PR can be merged after #286  has been merged.

```
$ git diff --name-status rtl8720dn_mqtt
M       examples/rtl8720dn/tlsclient/main.go
M       examples/rtl8720dn/webclient-tinyterm/main.go
M       examples/rtl8720dn/webclient/main.go
M       examples/rtl8720dn/webserver/main.go
A       net/http/client.go
A       net/http/cookie.go
A       net/http/cookiejar/jar.go
A       net/http/cookiejar/punycode.go
A       net/http/jar.go
M       net/http/request.go
M       net/http/response.go
A       net/http/tinygo.go
A       net/ipsocki.go
M       rtl8720dn/netdriver.go
M       rtl8720dn/rtl8720dn.go
```

PR #291, opened 2021-08-05
wifinina: change examples/wifinina to use http.Get()

sago35:
Changed to use tinygo.org/drivers/net/http.Get(), added in #287
With this change, webclient and tlsclient now have almost the same source code.
This means that there is no longer a difference between http and https.

```
$ diff examples/wifinina/webclient/main.go examples/wifinina/tlsclient/main.go
2,3c2
< // and sends a HTTP request to retrieve a webpage, based on the following
< // Arduino example:
---
> // and sends a HTTPS request to retrieve a webpage
5c4,6
< // https://github.com/arduino-libraries/WiFiNINA/blob/master/examples/WiFiWebClientRepeating/
---
> // You shall see "strict-transport-security" header in the response,
> // this confirms communication is indeed over HTTPS
> // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
26c27
< const url = "http://tinygo.org/"
---
> const url = "https://www.example.com/"
39c40
< var buf [0x400]byte
---
> var buf [0x1000]byte
```

The bad thing about this PR is that the memory usage has probably increased.
The current drivers/net/http are implemented in such a way that the entire response is put into memory.
This may be a problem for samd21.

If memory usage is a problem, it may be a good idea to leave the current webclient/tlsclient as is and create a new example.

### DIFF
--- a/examples/wifinina/http-get/main.go
+++ b/examples/wifinina/http-get/main.go
@@ -4,6 +4,12 @@
 //
 // https://github.com/arduino-libraries/WiFiNINA/blob/master/examples/WiFiWebClientRepeating/
 //
+// This example will not work with samd21 or other systems with less than 32KB
+// of RAM.  Use the following if you want to run wifinina on samd21, etc.
+//
+// examples/wifinina/webclient
+// examples/wifinina/tlsclient
+//
 package main
 
 import (
@@ -23,6 +29,7 @@ const ssid = ""
 const pass = ""
 
 // IP address of the server aka "hub". Replace with your own info.
+// Can specify a URL starting with http or https
 const url = "http://tinygo.org/"
 
 // these are the default pins for the Arduino Nano33 IoT.
--- a/examples/wifinina/http-get/main.go
+++ b/examples/wifinina/http-get/main.go
@@ -1,0 +1,154 @@
+// This example opens a TCP connection using a device with WiFiNINA firmware
+// and sends a HTTP request to retrieve a webpage, based on the following
+// Arduino example:
+//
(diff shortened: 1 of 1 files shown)


sago35:
I got an Arduino MKR WiFi 1010 board and tried it out.
This board is a combination of samd21 + NINA-W102.

https://github.com/tinygo-org/tinygo/pull/2056

Unfortunately, it seems to be an out-of-memory error.

It seems that samd21 is giving me memory errors after about 5-10 accesses in most cases.
I am planning to check what the problem is.
With pyportal, there is no error even if you access the file more than 1000 times.

```
HTTP/1.1 301 Moved Permanently
Cache-Control: public, max-age=0, must-revalidate
Content-Length: 34
Content-Type: text/plain
Date: Tue, 17 Aug 2021 04:53:58 GMT
X-Nf-Request-Id: 01FDA1SC3FW610K35CSR8CMRMR
Location: https://tinygo.org/
Server: Netlify
Age: 27733

Redirecting to https://tinygo.org/
-------- 19 --------
panic: runtime error: out of memory
```

ysoldak:
Works of Nano RP2040, crashes with out of memory on Nano 33 IoT

deadprogram:
I would like to merge this, but we cannot just leave the samd21 crashing, right? How should we proceed?

sago35:
@deadprogram 
How about undoing the changes in webclient and tlsclient and adding examples/wifinina/http-get?
I think it would be better to add a comment to examples/wifinina/http-get saying that it does not work with samd21.

deadprogram:
Sounds like a good plan @sago35

deadprogram:
Thanks for the updates @sago35 now merging.

PR #292, opened 2021-08-07
ws2812: generate assembly instead of handwriting it

aykevl:
Previously, a new implementation had to be written for each processor
speed. This is tedious and error-prone, therefore I've rewritten all of
the assembly code using a tool that will calculate exactly how many NOP
instructions are required.

I've also switched from build tags to a switch statement over all
processor speeds.

All in all, this means that:
  - Chips with a supported CPU performance will automatically be
    supported (no build tags necessary).
  - Adding a new chip is as easy as adding the MHz count to the
    generator script and to the switch statement.

Right now this is only for the Cortex-M, but it's certainly feasible to
extend this to other architectures such as the AVR.

Example: to support the stm32f103 chip (at 72MHz), you'd only need to change 1 line and add 2 lines (and run `go generate`).

I tested this on the following boards, both with a WS2811 and WS2812 LED strip (and on-board WS2812 chips if available):

 - BBC micro:bit (16MHz nrf51)
 - Circuit Playground Express (48MHz samd21)
 - ItsyBitsy nRF52840 (64MHz nrf52840)
 - PyBadge (120MHz samd51)

Therefore, all supported clock frequencies should be tested this way.


aykevl:
I have some more improvements coming up, it would be great to have this PR merged first.

ws2812/gen-ws2812-arm.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func main() {
+	f, err := os.Create("ws2812-asm_cortexm.go")
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "could not generate WS2812 assembly code:", err)
+		os.Exit(1)
+	}
+	defer f.Close()
+	f.WriteString(`// +build cortexm
```
deadprogram, line 166: Shouldn't this be `//go:build cortexm` now?

ws2812/ws2812_cortexm.go
```diff
@@ -0,0 +1,34 @@
+// +build cortexm
```
deadprogram, line 1: See previous comment on `go:build` directive.

ws2812/gen-ws2812-arm.go
```diff
@@ -0,0 +1,185 @@
+// +build none
```
deadprogram, line 1: `//go:build none`?

deadprogram:
Actually I think I will just merge, and then correct any formatting issues with the change to Go 1.17 in a separate PR.

PR #293, opened 2021-08-20
net/http: improve http request

sago35:
This PR improves the net/http request.

* Add support for req.URL.RawQuery
* Add support for case-insensitive header names


sago35:
The parts that will be improved in this PR are needed for the implementation of the twitter api.
It will also be needed for many other services' API implementations.

deadprogram:
Looks good, thanks @sago35 now merging.

Issue #276, opened 2021-06-20
HTS221 support


deadprogram:
We need support for the HTS221 capacitive digital sensor for relative humidity and temperature.

It is one of the sensors onboard the Arduino Nano 33 BLE Sense.

deadprogram:
See https://content.arduino.cc/assets/Nano_BLE_Sense_HTS221.pdf

sago35:
The following may be helpful.

https://github.com/arduino-libraries/Arduino_HTS221

alankrantas:
Tried to look into it. I tried to raise a response with the "who am I" register but... the code would simply hang on any I2C operations.

```golang
func main() {

	const HTS221_ADDRESS = 0x5F
	const HTS221_WHO_AM_I_REG = 0x0F

	i2c := machine.I2C1
	i2c.Configure(machine.I2CConfig{
		SCL:       machine.P0_15,  // SCL1
		SDA:       machine.P0_14,  // SDA1
	})

	data := []byte{0}
	err := i2c.ReadRegister(HTS221_ADDRESS, HTS221_WHO_AM_I_REG, data)
	println(data[0], err)

}
```

[schematics](https://content.arduino.cc/assets/NANO33BLE_V2.0_sch.pdf)

If I set to other pins or set nothing I would get an "I2C bus error". I couldn't get "who am I" response from LSM9DS1 either (also stuck). I can ping all onboard devices with the Arduino I2C scanner script.

Is there any way to check if the SCL1/SDA1 pins are working properly in TinyGo?

sago35:
@alankrantas 
I tried it with feather-nrf52840-sense, which has the same nRF52840 microcontroller.
~~It did not work when the I2C initialization part was set to `I2C1`.~~

~~Try changing it to `I2C0` and see if it works.~~
~~If it does, there may be some changes needed in `machine_nrf.go`.~~

Either of the following will work

```go
	// It works (I2C0)
	machine.I2C0.Configure(machine.I2CConfig{
		Frequency: machine.TWI_FREQ_400KHZ,
		SDA:       machine.P0_14,
		SCL:       machine.P0_15,
	})
```

```go
	// It works (I2C1)
	machine.I2C1.Configure(machine.I2CConfig{
		Frequency: machine.TWI_FREQ_400KHZ,
		SDA:       machine.P0_14,
		SCL:       machine.P0_15,
	})
```

sago35:
[...]

PR #295, opened 2021-08-26
Add support for HTS221 capacitive digital sensor for relative humidity and temperature

alankrantas:
For #276.


hts221/hts221.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// New creates a new HTS221 connection. The I2C bus must already be
+// configured.
+//
+// This function only creates the Device object, it does not touch the device.
+//
+// For Nano 33 BLE Sense, use machine.P0_15 (SCL1) and machine.P0_14 (SDA1),
+// and set onNano33BLE as hts221.ON_NANO_33_BLE.
+func New(bus drivers.I2C, deviceType uint8) Device {
```
aykevl: You can also use the `nano_33_ble` build tag for this. I think that's a lot simpler to use.

hts221/hts221.go
```diff
@@ -0,0 +1,208 @@
...
+	// and wait a moment.
+	if d.OnNano33BLE {
+		ENV := machine.Pin(22)
+		ENV.Configure(machine.PinConfig{Mode: machine.PinOutput})
+		ENV.High()
+		R := machine.Pin(32)
+		R.Configure(machine.PinConfig{Mode: machine.PinOutput})
+		R.High()
+		time.Sleep(time.Millisecond * 10)
+	}
```
aykevl: This should be done in the configuration step. A user is not required to call `.Connected()`.

hts221/hts221.go
```diff
@@ -0,0 +1,208 @@
...
+	data := []byte{0}
+	if status {
+		data[0] = 0x84
+	}
+	d.bus.WriteRegister(d.Address, HTS221_CTRL1_REG, data)
+}
+
+// ReadPressure returns the relative humidity in percent * 1000.
+// Returns an error if the device is not turned on.
+func (d *Device) ReadHumidity() (humidity int32, err error) {
```
aykevl, line 51: The BME280 driver has the same interface but returns humidity in percent * 100. I think it would be better to match that API.
Also, the comment says `ReadPressure`, it should be `ReadHumidity`.

hts221/hts221.go
```diff
@@ -0,0 +1,208 @@
...
+	d.bus.ReadRegister(d.Address, HTS221_T1_OUT_REG+1, t1Out[1:])
+
+	h0rH_v := float32(h0rH[0]) / 2.0
+	h1rH_v := float32(h1rH[0]) / 2.0
+	t0degC_v := float32(readUint(t1t0msb[0]&0x03, t0degC[0])) / 8.0
+	t1degC_v := float32(readUint(t1t0msb[0]&0x0C>>2, t1degC[0])) / 8.0
+	h0t0Out_v := float32(readInt(h0t0Out[1], h0t0Out[0]))
+	h1t0Out_v := float32(readInt(h1t0Out[1], h1t0Out[0]))
+	t0Out_v := float32(readInt(t0Out[1], t0Out[0]))
+	t1Out_v := float32(readInt(t1Out[1], t1Out[0]))
```
aykevl, line 130: It is usually better to only use integers, because floating point can be slow. However, in this case it doesn't matter much I think.

aykevl (commented):
This generally looks good to me, just a few comments.

hts221/hts221.go
alankrantas: The trouble is that without pull high these pins first, the I2C bus simply hangs.

Would it be ok if I put it in ```New()``` (and I create two versions of them using build tags)?

alankrantas: OK, just updated (including other drivers for Nano 33 BLE).

deadprogram:
Hello @alankrantas thanks for working on this. Seems like it could be merged if the commits were squashed and the merge conflict in the `Makefile` is resolved.

Do you want to do this yourself or would you like me to do this?

Thank you!

Issue #278, opened 2021-06-20
LPS22HB support


deadprogram:
We need support for the LPS22HB MEMS nano pressure sensor: 260-1260 hPa absolute
 digital output barometer.

It too is one of the onboard sensors for the Arduino Nano 33 BLE Sense.

See datasheet at https://content.arduino.cc/assets/Nano_BLE_Sense_lps22hb.pdf

sago35:
The following may be helpful.

https://github.com/arduino-libraries/Arduino_LPS22HB

deadprogram:
This was already released, so closing.

PR #297, opened 2021-08-28
Add support for LPS22HB MEMS nano pressure sensor

alankrantas:
For #278.


lps22hb/lps22hb.go
```diff
@@ -0,0 +1,73 @@
...
+
+// ReadPressure returns the pressure in milli pascals (mPa).
+func (d *Device) ReadPressure() (pressure int32, err error) {
+	d.waitForOneShot()
+
+	// read data
+	data := []byte{0, 0, 0}
+	d.bus.ReadRegister(d.Address, LPS22HB_PRESS_OUT_REG, data[:1])
+	d.bus.ReadRegister(d.Address, LPS22HB_PRESS_OUT_REG+1, data[1:2])
+	d.bus.ReadRegister(d.Address, LPS22HB_PRESS_OUT_REG+2, data[2:])
```
aykevl, line 39: You can probably simplify this to:

```suggestion
	d.bus.ReadRegister(d.Address, LPS22HB_PRESS_OUT_REG, data)
```

lps22hb/lps22hb.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// ReadTemperature returns the temperature in celsius milli degrees (°C/1000).
+func (d *Device) ReadTemperature() (temperature int32, err error) {
+	d.waitForOneShot()
+
+	// read data
+	data := []byte{0, 0}
+	d.bus.ReadRegister(d.Address, LPS22HB_TEMP_OUT_REG, data[:1])
+	d.bus.ReadRegister(d.Address, LPS22HB_TEMP_OUT_REG+1, data[1:])
```
aykevl, line 52: Same here.

```suggestion
	d.bus.ReadRegister(d.Address, LPS22HB_TEMP_OUT_REG, data)
```

lps22hb/lps22hb.go
```diff
@@ -0,0 +1,73 @@
...
+	tValue := float32(int16(uint16(data[1])<<8|uint16(data[0]))) / 100.0
+
+	return int32(tValue * 1000), nil
```
aykevl, line 55: I think this can be simplified to the following:

```suggestion
	tValue := int16(uint16(data[1])<<8|uint16(data[0]))

	return int32(tValue) * 10, nil
```

This avoids floats altogether.

lps22hb/lps22hb_nano_33_ble.go
```diff
@@ -0,0 +1,28 @@
...
+func New(bus drivers.I2C) Device {
+	// turn on internal power pin (machine.P0_22) and I2C1 pullups power pin (machine.P1_00)
+	// and wait a moment.
+	ENV := machine.P0_22
+	ENV.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	ENV.High()
+	R := machine.P1_00
+	R.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	R.High()
+	time.Sleep(time.Millisecond * 10)
```
aykevl, line 26: I'm slightly uncomfortable with adding board specific code in a driver. This feels like a layering violation to me. It seems to me this is something an application should do (or some sort of board package, which we currently don't provide for TinyGo). @deadprogram any opinion here?

aykevl (commented):
In general looks reasonable to me, with some comments.

lps22hb/lps22hb.go
alankrantas: I know, I just thought this would be easier to understand for other maintainers.

lps22hb/lps22hb_nano_33_ble.go
alankrantas: I have no opinions really, but my only concern is that if you implement this in board package, there would have to have means to turn on/off the internal power at will, since Nano 33 BLE is very popular for IoT/TinyML projects and may need to save battery power.

EDIT:

Looking back on the example code, putting these pin operations in ```Configure()``` would add the risk of I2C bus hanging if the user calls ```Connected()``` first.

In Arduino's drivers they did something like this, and I suppose ```Wire1``` would turn on these pins itself:

```c++
#ifdef ARDUINO_ARDUINO_NANO33BLE
LPS22HBClass BARO(Wire1);
#else
LPS22HBClass BARO(Wire);
#endif
```

lps22hb/lps22hb.go
alankrantas: I think I have tried this on one of the onboard sensors and the register addresses just don't add on automatically. I'll have to try later.

lps22hb/lps22hb_nano_33_ble.go
deadprogram: Hmm, looking at this specifically, along with the other drivers that are onboard the Nano 33 BLE, seems like we would like a balance between convenience and architectural integrity.

As far as the `New()` function, we'd probably prefer to keep that board agnostic. It would be better to use the `Configure()` function for this, since that would facilitate testing similarly to how other I2C devices are being tested in this package.

Having "magic pins" like `machine.P0_22` and `machine.P1_00` seems non-intuitive. I'd rather see something added to the main board definition such as: 

```go
const (
	I2C1_PullupPower Pin = P1_00
)
```

At least if a special `Configure()` protected by a build tag really does have to be created, it is easier to see what it is doing and why.

What do you think?

PR #298, opened 2021-08-30
drivers: add mpu6050 additional functionality

soypat:
Create a new device type: `DeviceStored`. It is a "better" alternative to `Device` for getting acceleration+gyroscope measurements together, less bus usage, faster sample rate and less chance of a failed read, which is a common problem on the MPU6050 with high sampling frequency.

**Benefits**
* One read on i2c bus per gyroscope+accelerometer+temperature (and in some cases, also magnetometer) measurements. Less overhead of getting measurements.
* Smart choice for a IMU interface since you may want to pass the IMU handle around but don't want other libraries re-reading i2c bus. MPU6050 has sampling rate issues and reads may fail at high sample rates.

**Disadvantages**
* data structure consumes 14-20 more bytes
* Must call `Get` or `Update` before calling a measurement function though this could be remedied by embedding the type within a sample rate limiter.

I'd also raise an idea I have in mind with this PR: creating an interface called `ARS/AHRS` or something of the sort:

```go
// ARS Attitude reference system for complementary filter and other basic filters.
type ARS interface {
    Get() error
    Acceleration() (ax,ay,az int32) // micro gravities
    Rotation() (gx, gy, gz int32) // microradians would be ideal for these, right now it's in microdegrees
}

// AHRS Attitude-Heading reference system used by sensor fusion libraries which 
// implement filters such as Madgwick or Mahony.
type AHRS interface {
    ARS
    North() (mx, my, mz int32) // micro teslas
}
```

Working with the Raspberry Pi Pico for bus baud above 40kHz.

### DIFF
--- a/mpu6050/io.go
+++ b/mpu6050/io.go
@@ -1,0 +1,15 @@
+package mpu6050
+
+// Read reads a single register with minimal heap allocations
+func (d *DeviceStored) read(reg uint8) (byte, error) {
+	d.rreg[0] = reg
+	err := d.bus.Tx(uint16(d.Address), d.rreg[:1], d.buf[:1])
+	return d.buf[0], err
+}
+
+// Write writes a single register.
+func (d *DeviceStored) write(reg uint8, data byte) (err error) {
+	d.buf[0] = reg
+	d.buf[1] = data
+	return d.bus.Tx(d.Address, d.buf[:2], nil)
+}
--- a/mpu6050/mpu6050.go
+++ b/mpu6050/mpu6050.go
@@ -20,7 +20,7 @@ type Device struct {
 //
 // This function only creates the Device object, it does not touch the device.
 func New(bus drivers.I2C) Device {
-	return Device{bus, Address}
+	return Device{bus: bus, Address: Address}
 }
 
 // Connected returns whether a MPU6050 has been found.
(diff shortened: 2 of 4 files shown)


aykevl:
Let's first reach a conclusion in #299 before merging this.

Regarding the code: I think it would be fine to merge the two (`Device` and `DeviceStored`). That avoids duplication and I think the extra RAM consumption is not a big problem.

VirgilioRodrigues:
Hi, sample per second?

soypat:
> Hi, sample per second?

Depends on your microcontroller. If using a raspberry pi pico you will be limited by the MPU6050's max sample rate, by far.

VirgilioRodrigues:
I'm looking for solutions for acquiring vibration mechanisms, using Mpu 6050 and 9250 and 9250... I'm finding it quite difficult.... I have to isolate the MPU usage for the accelerometer function, find the connection too....

soypat:
@VirgilioRodrigues What issues are you having? Is the sampling rate insufficient? 

Keep in mind in nominal conditions you'll rarely be getting more than 1kHz sampling frequency. I2C baud, supplied voltage and configuration can further limit this value to well below 1kHz. for best results use high I2C baud (400kHz should be more than enough) and configure the sampling frequency to be as high as possible.

 I'm not sure if current driver sets it at highest, this PR had the intention of letting users modify the sampling frequency.

PR #303, opened 2021-09-01
image: add support for image/jpeg and image/png

sago35:
This PR adds support for image/jpeg and image/png.
The original image package uses too much memory to run on a microcontroller.
I reduced the memory usage as follows and registered it in the driver as TinyGo version.

* image/jpeg
    * image.Image structure is no longer returned
    * Callback is called every 16x16 pix
    * There was a 32KB make() in the compress/zlib package, so I made it a fixed array
* image/png
    * image.Image structure is no longer returned
    * Callback is called every 1 x width pixel.

If the memory size is not a problem, the original image package can be used, so I don't think this should be registered in tinygo-org/tinygo.


sago35:
It works as shown in the video below.
I have confirmed that it works with the following boards.

* atsamd51
    * wioterminal
    * pyportal
* nrf52840
    * feather-nrf52840
* rp2040
    * feather-rp2040
* stm32f405
    * feather-stm32f405 (After running it for a while, it runs out of memory and gives an error)

https://twitter.com/sago35tk/status/1432997940738490374?s=20

sago35:
To run `image/jpeg`, at least 32KB of RAM is required, so it will not work on samd21.
I think `image/png` will work on samd21, but I have not checked.

deadprogram:
@sago35 could you please add a README.md to the main `image` package explaining how to use this? I was a little confused on how to get started.

sago35:
@deadprogram I've added it.
If you find English difficult to understand, please let me know how to write it.

https://github.com/tinygo-org/drivers/blob/1406a8c9c47de368f48c390e0cc4a1d8998a5fad/image/README.md

sago35:
The following changes may be necessary for the success of the CI.
I'll try to rebase it.

e70beede1c848bb67571f2b51639582882145fe6

deadprogram:
I wonder if perhaps `examples/ili9341/slideshow/convert2bin/convert2bin.go` should be located at `cmd/convert2bin/convert2bin.go` instead? Also I did not immediate understand that it was a sort of minimal implementation of an `embed` like functionality.

sago35:
Move convert2bin to . /cmd/convert2bin.
Please let me know if there are any other parts of the documentation that are missing.

deadprogram:
@sago35 do you want to squash any commits before merging?

sago35:
Squash merge, please.

deadprogram:
Thanks @sago35

Issue #302, opened 2021-09-01
target pall on


shayulei:
tinygo build -o ./dht11.hex -size=short -target=d1mini .
# tinygo.org/x/drivers/dht
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/constants.go:105:10: undeclared name: counter
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/constants.go:112:29: undeclared name: counter
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/thermometer.go:158:47: undeclared name: counter
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/thermometer.go:170:40: undeclared name: counter
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/util.go:18:49: undeclared name: counter
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/constants.go:115:9: undeclared name: counter
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/thermometer.go:127:21: undeclared name: counter
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/thermometer.go:160:10: UART1 not declared by package machine
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/thermometer.go:161:16: UART1 not declared by package machine
../../../../../../../../root/Documents/GoprojectLinux/pkg/mod/tinygo.org/x/drivers@v0.17.1/dht/util.go:19:9: undeclared name: counter

sago35:
Same problem as #301.
I think #304 solves it a bit.

* d1mini has less memory, so it is better not to use the `fmt` package
* The UART1 part is a board-dependent code and needs to be modified. You will need to consider how to fix it separately. If interrupts are really a problem, functions like `arm.DisableInterrupts()` are needed, but there may not be any for d1mini right now.
[...]

Issue #301, opened 2021-08-31
Multiple Drivers fail to compile on Raspberry Pico


SoftTacos:
Hi there, two things before I get to my post. First, I'm new to the realm of microcontrollers, so I could be(probably am) missing something here. Second, I just want to express gratitude on the existence of TinyGo, is fantastic and usually easier to just build things myself on my pico in TinyGo.
I'm running tinygo from my windows10 machine with drivers v0.17.1, tinygo v0.19.0 and have been able to get code that I wrote from scratch to work just fine. I'm guessing these errors are because something isn't supported yet for the pico, however I don't know what I don't know here and want to make sure I'm not doing anything pants-on-head.

The first compile error I got is with the [DHT drivers](https://github.com/tinygo-org/drivers/blob/release/examples/dht/main.go):
```
d := dht.New(machine.GP14, dht.DHT22)
d.Measurements()
```
I ran: `tinygo flash -target=pico ./`
And got:
```
# tinygo.org/x/drivers/dht
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.17.1\dht\constants.go:105:10: undeclared name: counter
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.17.1\dht\constants.go:112:29: undeclared name: counter
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.17.1\dht\thermometer.go:158:47: undeclared name: counter
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.17.1\dht\thermometer.go:170:40: undeclared name: counter
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.17.1\dht\util.go:18:49: undeclared name: counter
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.17.1\dht\constants.go:113:18: CPUFrequency not declared by package machine
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.17.1\dht\constants.go:115:9: undeclared name: counter
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.17.1\dht\thermometer.go:127:21: undeclared name: counter
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.17.1\dht\util.go:19:9: undeclared name: counter
```
[...]

PR #304, opened 2021-09-01
dht: improve build tag in lowfreq.go

sago35:
This PR resolves some of the issues in #301 and #302.


deadprogram:
Thanks @sago35 now merging.

aykevl:
I wonder, why does the `counter` type vary between chips? Is it to save a little bit of RAM?

Issue #277, opened 2021-06-20
APDS-9960 support


deadprogram:
We need support for the APDS-9960 Digital Proximity, Ambient Light, RGB and Gesture Sensor

It is onboard the Arduino Nano 33 BLE Sense.

See https://content.arduino.cc/assets/Nano_BLE_Sense_av02-4191en_ds_apds-9960.pdf

sago35:
The following may be helpful.

https://github.com/adafruit/Adafruit_APDS9960

deadprogram:
This has already been released, so now closing. Thanks again @alankrantas

PR #308, opened 2021-09-05
Add support for APDS-9960, Digital Proximity, Ambient Light, RGB and Gesture Sensor

alankrantas:
For #277.

This one is tricky since there are a lot of functionalities packed in this little thing. For now I simply use the Arduino version's default settings (pulse length, pulse count, gain, etc.) so that the gesture detection would work.


apds9960/apds9960_nano_33_ble.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"tinygo.org/x/drivers"
+)
+
+// New creates a new APDS-9960 connection. The I2C bus must already be
+// configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func New(bus drivers.I2C) Device {
+	// turn on internal power pin (machine.P0_22) and I2C1 pullups power pin (machine.P1_00)
```
ysoldak, line 18: Comment does not match the code?

apds9960/apds9960_nano_33_ble.go
```diff
@@ -0,0 +1,28 @@
...
+	// and wait a moment.
+	ENV := machine.P0_15
```
ysoldak: I wonder, would it be good idea to extract these special pins to tinygo repo, in `board_nano-33-ble.go`?
Also, this initialisation sequence seem to be generic for this sensor, only difference from board to board is pins? If so, why not move it into `apds9960.Configure`? Much the same as wifinina driver does, for example.

ysoldak (commented):
Thank you, for the PR!
I have Nano 33 BLE Sense board and a project in mind that involves proximity sensor. How timely!

apds9960/apds9960_nano_33_ble.go
alankrantas: Which part? "it does not touch the device"? In a sense it is, because this only turns on the power supply.

ysoldak: No, I mean the pin numbers

alankrantas: Ok, I guess I mixed up the internal I2C pins.

ysoldak:
Verified on my Nano 33 BLE Sense board, works!
Thank you for implementing this.

Issue #290, opened 2021-08-05
examples/wifinina/tlsclient failed


sago35:
If the URL is changed to the following, no error will be displayed.

* www.google.com
* github.com

The environment is as follows.

* windows10
* pyportal + NINA_W102-1.7.4.bin
* tinygo-org/drivers dev (9234d7d9ffee39a7eb83239a90aed15f288655cf)
* tinygo e834d7887143fc8da542d3d29907e8db7a5b9bb2 (2021/07/16)

```
$ tinygo flash --target pyportal --size short --programmer jlink ./examples/wifinina/tlsclient/
```

```
Connecting to XXXXXX
Connection status: No SSID Available
Connection status: No SSID Available
Connection status: No SSID Available
Connection status: No SSID Available
Connection status: No SSID Available
Connected.
192.168.1.123

---------------
Dialing TCP connection
Connection failed: wifinina error: 0x01
```

scottfeldman:
Just a note for now: this is fixed in netdev

deadprogram:
Closing since no longer valid. Thanks everyone!

PR #311, opened 2021-09-13
Wifinina: transparently maintain wifi connection

ysoldak:
This encapsulates the tedious task of maintaining connection with an access point.
Now you don't have to have that boilerplate in your application code.

Especially useful for devices that can be moved and/or not always have stable WiFi connection.
Supports multiple access points.

Plays nicely with `http.Get` and friends, see refactored example.

P.S. I've tried hard maintain backwards compatibility. Otherwise the API could, probably, be better.

### DIFF
--- a/examples/wifinina/http-get/main.go
+++ b/examples/wifinina/http-get/main.go
@@ -27,23 +27,21 @@ import (
 // access point info
 const ssid = ""
 const pass = ""
+const timeout = time.Minute
 
-// IP address of the server aka "hub". Replace with your own info.
 // Can specify a URL starting with http or https
-const url = "http://tinygo.org/"
+const url = "https://raw.githubusercontent.com/tinygo-org/tinygo/release/LICENSE"
 
-// these are the default pins for the Arduino Nano33 IoT.
-// change these to connect to a different UART or pins for the ESP8266/ESP32
 var (
-
-	// these are the default pins for the Arduino Nano33 IoT.
+	// these are the default pins for the Arduino Nano connected boards.
+	// change these to connect to a different UART or pins for the ESP8266/ESP32
 	spi = machine.NINA_SPI
 
 	// this is the ESP chip that has the WIFININA firmware flashed on it
-	adaptor *wifinina.Device
+	device *wifinina.Device
 )
 
-var buf [0x400]byte
+var buf [0x1000]byte
 
 var lastRequestTime time.Time
 var conn net.Conn
@@ -57,22 +55,27 @@ func setup() {
 		SCK:       machine.NINA_SCK,
 	})
 
-	adaptor = wifinina.New(spi,
+	device = wifinina.New(spi,
 		machine.NINA_CS,
 		machine.NINA_ACK,
 		machine.NINA_GPIO0,
 		machine.NINA_RESETN)
-	adaptor.Configure()
+	device.Configure().WithAccessPoint(ssid, pass, timeout)
 }
 
 func main() {
+
+	if len(ssid) == 0 || len(pass) == 0 {
+		for {
+			println("Please set ssid and password for this example to work")
+			time.Sleep(10 * time.Second)
+		}
+	}
 
 	setup()
 	http.SetBuf(buf[:])
 
 	waitSerial()
-
-	connectToAP()
 
 	// You can send and receive cookies in the following way
 	// 	import "tinygo.org/x/drivers/net/http/cookiejar"
@@ -128,34 +131,4 @@ func waitSerial() {
 	for !machine.Serial.DTR() {
 		time.Sleep(100 * time.Millisecond)
 	}
-}
-
-// connect to access point
-func connectToAP() {
-	if len(ssid) == 0 || len(pass) == 0 {
-		for {
-			println("Connection failed: Either ssid or password not set")
-			time.Sleep(10 * time.Second)
-		}
-	}
-	time.Sleep(2 * time.Second)
-	message("Connecting to " + ssid)
-	adaptor.SetPassphrase(ssid, pass)
-	for st, _ := adaptor.GetConnectionStatus(); st != wifinina.StatusConnected; {
-		message("Connection status: " + st.String())
-		time.Sleep(1 * time.Second)
-		st, _ = adaptor.GetConnectionStatus()
-	}
-	message("Connected.")
-	time.Sleep(2 * time.Second)
-	ip, _, _, err := adaptor.GetIP()
-	for ; err != nil; ip, _, _, err = adaptor.GetIP() {
-		message(err.Error())
-		time.Sleep(1 * time.Second)
-	}
-	message(ip.String())
-}
-
-func message(msg string) {
-	println(msg, "\r")
 }
--- a/wifinina/tcp.go
+++ b/wifinina/tcp.go
@@ -12,8 +12,33 @@ const (
 	ReadBufferSize = 128
 )
 
-func (d *Device) NewDriver() net.DeviceDriver {
-	return &Driver{dev: d, sock: NoSocketAvail}
+// Configure resets wifinina device, creates network driver and tells network stack to use it.
+// Wifi connection can either be handled esplicitly in user code or transparently by using WithAccessPoint.
+func (d *Device) Configure() *Driver {
+	d.Reset()
+	drv := &Driver{dev: d, sock: NoSocketAvail}
+	net.UseDriver(drv)
+	return drv
+}
+
+// WithAccessPoint instructs the driver to establish and maintain access point connection transparently for the user.
+// Multiple access points can be configured.
+// In such case and when multiple access points available, active access point is selected in random.
+// Each access point has an individual timeout after which connection attempt deemed failed and reconnect shall be tried.
+func (drv *Driver) WithAccessPoint(ssid, password string, timeout time.Duration) *Driver {
+	if drv.accessPoints == nil {
+		drv.accessPoints = make([]accessPoint, 1)
+	}
+	drv.accessPoints = append(drv.accessPoints, accessPoint{ssid, password, timeout})
+	return drv
+}
+
+// WithTimeout configures global timeout for transparent access point connection handling.
+// Once this timeout elapses, no more connection and scan retries attempted and network operation returns with timeout error.
+// When timeout not set, driver tries reconnect to access points forever.
+func (drv *Driver) WithTimeout(timeout time.Duration) *Driver {
+	drv.timeout = timeout
+	return drv
 }
 
 type Driver struct {
@@ -24,6 +49,9 @@ type Driver struct {
 	proto uint8
 	ip    uint32
 	port  uint16
+
+	accessPoints []accessPoint
+	timeout      time.Duration
 }
 
 type readBuffer struct {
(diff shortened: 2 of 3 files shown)


examples/wifinina/http-get/main.go
```diff
@@ -57,23 +55,28 @@ func setup() {
...
 	})
 
-	adaptor = wifinina.New(spi,
+	device = wifinina.New(spi,
 		machine.NINA_CS,
 		machine.NINA_ACK,
 		machine.NINA_GPIO0,
 		machine.NINA_RESETN)
-	adaptor.Configure()
+	device.Configure().WithAccessPoint(ssid, pass, timeout)
```
sago35, line 63: It might be better to use Configure() to match the other drivers in tinygo.org/x/drivers, but maybe that's just my preference.
If the reason for using fluent-style is clear, it is good to proceed as is.

examples/wifinina/http-get/main.go
```diff
@@ -27,23 +27,21 @@ import (
 // access point info
 const ssid = ""
 const pass = ""
+const timeout = time.Minute
 
-// IP address of the server aka "hub". Replace with your own info.
 // Can specify a URL starting with http or https
-const url = "http://tinygo.org/"
+const url = "https://raw.githubusercontent.com/tinygo-org/tinygo/release/LICENSE"
```
sago35, line 33: I think this change is very good.
I would prefer https, but I know there is a #290 problem with https.

wifinina/tcp.go
```diff
@@ -12,8 +12,33 @@ const (
...
-func (d *Device) NewDriver() net.DeviceDriver {
-	return &Driver{dev: d, sock: NoSocketAvail}
+// Configure resets wifinina device, creates network driver and tells network stack to use it.
+// Wifi connection can either be handled esplicitly in user code or transparently by using WithAccessPoint.
+func (d *Device) Configure() *Driver {
+	d.Reset()
+	drv := &Driver{dev: d, sock: NoSocketAvail}
+	net.UseDriver(drv)
+	return drv
+}
```
sago35, line 22: How about connecting to AccessPoint by passing the argument `Config{}` to this `Configure()`?

wifinina/tcp.go
```diff
@@ -32,7 +60,97 @@ type readBuffer struct {
...
+		return nil
+	}
+	for i := 0; i < int(count); i++ {
+		ssid := drv.dev.GetNetworkSSID(i)
+		for _, ap := range drv.accessPoints {
+			if ap.ssid == ssid {
+				return &ap
+			}
+		}
+	}
```
sago35, line 146: It is possible to set multiple accessPoints, but the first one found will always be selected.
This is a good solution for embedded systems.

wifinina/tcp.go
```diff
@@ -24,6 +49,9 @@ type Driver struct {
 	proto uint8
 	ip    uint32
 	port  uint16
+
+	accessPoints []accessPoint
```
sago35, line 53: It is a good idea to store the information of the access point. This makes it easy to reconnect, unlike rtl8720dn and older wifinina.
I don't know if it should be stored in wifinina.Driver or in wifinina.Device.

I would like to know if there is a good reason.

wifinina/tcp.go
```diff
@@ -32,7 +60,97 @@ type readBuffer struct {
...
+	}
+
+	for {
+
+		// global timeout elapsed, throw
+		if !stopTime.IsZero() && time.Now().After(stopTime) {
+			return errors.New("global wifi timeout elapsed")
+		}
+
+		st, err := drv.dev.GetConnectionStatus()
```
sago35, line 92: Apart from this PR, it is better to make sure that there is no communication with external devices here.
Since HandleWifi() is called by most methods, it is better to get out of it in as short a time as possible.

sago35 (commented):
@ysoldak 
I think most of the changes are good, but I have a few comments.

examples/wifinina/http-get/main.go
ysoldak: It works on my board, Nano RP2040 Connect. But then, I've updated it to latest nina-fw (1.4.8)

wifinina/tcp.go
ysoldak: I've kept API unchanged intentionally.
Since many examples use `Configure` w/o params and all user code out there in the wild too.
Introduce `ConfigureWithAccessPoint` to be backwards compatible?
Settled on fluent-style (`WithAccessPoint`) to satisfy both needs. Also future-proof if we ever add more "extra" functionality.

wifinina/tcp.go
ysoldak: I was thinking of comparing RSSI, but that can be next little feature.

wifinina/tcp.go
ysoldak: It's a bit of definitions mix here.
So, my reasoning goes like this: 
- `wifinina.Device` is about talking to the chip, lower level. It is a "driver" for this chip, in spirit of this repo.
- `wifinina.Driver` is about networking, more like adapter actually, making `wifinina.Device` usable in TinyGo, with `net`, `net/tls` and `net/http` packages.

In this light, access point information belongs to higher level, means `wifinina.Driver`.

examples/wifinina/http-get/main.go
sago35: I had only tried it with 1.4.5. I'll try it later. Thanks for the information.

wifinina/tcp.go
sago35: From a compatibility point of view, it would certainly be better to add WithAccessPoint.
Do you have any opinions? @deadprogram @bgould

wifinina/tcp.go
ysoldak: Not sure what you propose.
`HandleWifi`s job is to ensure the device is connected to AP. It is blocking by definition, not returning until condition is satisfied, since no sense to proceed with HTTP if we have no connection.

Just describing my thinking, but open for discussion and any improvements, obviously.

wifinina/tcp.go
sago35: I understood.
I think it is good as it is now.

wifinina/tcp.go
sago35: `GetConnectionStatus()`, when called, communicates with SPI and returns the result.
Since `HandleWifi()` is called by many functions, it is better to be able to know the ConnectionStatus without SPI communication.

```
func (d *Device) GetConnectionStatus() (ConnectionStatus, error) {
	status, err := d.getUint8(d.req0(CmdGetConnStatus))   // <---- SPI
	return ConnectionStatus(status), err
}
```

ysoldak: 🤷 It's only firmware in NINA chip knows for sure, we have to ask it. I see no alternative solution.

ysoldak: I can think of optimistic approach: assume everything is OK, handle error by re-connecting to AP and repeating the failed command?

ysoldak: Spent some time trying figure out any way to avoid SPI in `HandleWifi`. I don't see a way.
Thoughts:
- Every place we have `HandleWifi` function, there is always subsequent SPI call[s]. So it's not like we touch SPI when it can be completely avoided.
- HTTP communication is slow. SPI is not the bottleneck here.

ysoldak:
Any suggestions for better name?
`HandleWifi` is best I could come with. Alternative could be `EnsureWifi`. But they both so-so.
`ConnectWifi`? But we not always _connect_, just verify there is active connection.

examples/wifinina/http-get/main.go
deadprogram: I am not a fan of fluent interfaces generally. In this case, it introduces a new API pattern that is not in use elsewhere in this driver. IMO a more "Go-like" AP probably would be:

```go
device.Configure()
device.UseAutoReconnect(true)
device.ConnectToAP(ssid, pass, timeout)
```

EDIT: the function name was supposed to be `ConnectToAP` not `API`.

wifinina/tcp.go
deadprogram: See my previous comment.

wifinina/tcp.go
```diff
@@ -32,7 +60,97 @@ type readBuffer struct {
...
 }
 
+type accessPoint struct {
+	ssid     string
+	password string
+	timeout  time.Duration
+}
+
+// HandleWifi ensures device is connected to an access point (if any) and is called before any network operation
+func (drv *Driver) HandleWifi() error {
```
deadprogram, line 70: Why not name this `EnsureConnected()`?

Also, the first thing it should do is:

```go
if !drv.autoReconnect {
   return nil
}
```

This way only if the user has set `device.UseAutoReconnect(true)` that this is used.

examples/wifinina/http-get/main.go
ysoldak: With this approach access point info leaks to `wifinina.Device` (low level SPI communication adapter), while I've tried to keep it in `wifinina.Driver` (implements `net.DeviceDriver`). But I'm not adamant on this. Can, perhaps, move Wifi stuff down to `wifinina.Device` and leave `wifinina.Driver` care about HTTP only.

On splitting.
Splitting it like this makes it unclear what happens when `device.ConnectToAP(ssid, pass, timeout)` is called, but not `device.UseAutoReconnect(true)`.
And vice versa: what is `device.UseAutoReconnect(true)` without `device.ConnectToAP(ssid, pass, timeout)`?

Tricky also (naming-wise) how to support multiple AP:
`device.ConnectToAP(ssid, pass, timeout)` -- will it connect directly? What if I call it multiple times with different params?

Another fact that makes it hard with naming is in wifinina context we have two types of "connections": wifi and http[s]. Feels better not mix them, for example by using "AccessPoint" or "Wifi" terms for wifi level.

Let me try and move Wifi handling to `wifinina.Device` and avoid fluent style. I'll make a separate PR.
API in that case will look something like this:
```
device.Configure()
device.AddAccessPoint(ssid, pass, timeout, autoconnect)
device.AddAccessPoint(ssid2, pass2, timeout2, autoconnect)
device.ConnectToAccessPoint() // optional
```
Call of `device.AddAccessPoint(ssid, pass, timeout, autoconnect)` adds access point to the list of known access points and last parameter enables auto[re]connect. 
Call of `device.ConnectToAccessPoint()` will explicitly connect device to AP (returns fast if already connected). And this is the function we can use in `wifinina.Driver`.

How does this sound?

deadprogram: I would suggest look at the API to the MQTT driver for guidance here on how it should probably work.

Yes, `ConnectToAP()` should attempt to connect right away, and return an error if it fails. Reconnect logic should be to handle reconnections, not the initial connection.

If `device.autoReconnect` defaults to false, then if the connection fails then it has failed, and the various functions would return an error.

If `device.autoReconnect` is set to `true` via a call to `device.UseAutoReconnect(true)` then if the connection fails then it can try the reconnect logic.

As far as multiple AP, that would be best handled by the user's own logic. This is a case where `device.autoReconnect` being set to `false` seems like the least magic.

meyskens:
What is the current status of this PR? Is it still relevant? If so happy to help out apply the needed changes if needed :) (suffering from connection drops, this could help me solve it)

ysoldak:
> What is the current status of this PR? Is it still relevant? If so happy to help out apply the needed changes if needed :) (suffering from connection drops, this could help me solve it)

Hi Maartje!
I believe this PR is so outdated that it's probably easer to create new, in a branch from current dev head than trying to rebase this one.
Please, feel free to try and do that.

deadprogram:
I am in progress on the first steps for some refactoring of the interfaces to all of the WiFi adaptors.

Will open an issue for discussion of that topic soon.

deadprogram:
Closing now that `netdev` has landed.

PR #313, opened 2021-09-16
ws2812: improve timings to be compatible with the WS2811

aykevl:
The WS2811 has a slightly slower data rate, but it is close enough that
a single library can support both LEDs at the same time. In my testing,
this package was already compatible with WS2811 LEDs before this change
when running at 4-5V, but when the voltage dropped to 3V or so the LEDs
started misbehaving. With these improved timings, the LEDs remained of
the correct color even at very low voltages (although blue was starting
to fade due to lack of voltage - that's not a software/protocol issue).

The package may be a bit slower as a result, but it will also be more compatible so I think that's a win.

---

Sidenote: this is the result of me making some nice LED strings (WS2811) running from 3 1.2V AA batteries. They worked well when fully charged, but didn't work so well when the batteries were running empty. This PR fixes that issue.

### DIFF
--- a/ws2812/gen-ws2812-arm.go
+++ b/ws2812/gen-ws2812-arm.go
@@ -25,6 +25,13 @@ import (
 //   (possibly including some branch prediction).
 // It is certainly possible to extend this to other architectures, such as AVR
 // and RISC-V if needed.
+//
+// Here are two important resources. For the timings:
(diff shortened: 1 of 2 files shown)


deadprogram:
:traffic_light::traffic_light::traffic_light::traffic_light:

Thanks @aykevl now merging.

PR #314, opened 2021-09-16
ws2812: add support for 168MHz (e.g. Adafruit Feather STM32F405)

aykevl:
This was part of the same LED string as I describe in #313. I've used the Adafruit Feather STM32F405 because mine had convenient female headers.

Note that this PR has the timing fix for the WS2811 (#313) already applied.

This PR also requires some changes to the TinyGo machine package to be usable on STM32 chips, I'll make a PR soon.


deadprogram:
Since the related PR has been merged into TinyGo, this PR can be merged as well. Thanks @aykevl

PR #309, opened 2021-09-07
all: Go 1.17 fmt corrections

deadprogram:
This PR changes all build tags to match the format `go:build` that is now required by Go 1.17

It also adds a commit to modify the `make fmt-check` to actually check all of the files in each of the driver subdirectories. Oops!

PR #315, opened 2021-09-16
ws2812: fix generated build tags

aykevl:
See https://github.com/tinygo-org/drivers/pull/309 and https://github.com/tinygo-org/drivers/pull/292.

Before this change, the `//go:build` line would be removed. After this change, it is left in place.


deadprogram:
Thanks @aykevl now merging.

PR #317, opened 2021-09-19
wifinina: control nina pins, for example leds

ysoldak:
- Cherry-picked [bgould/wifinina-refactor@91236352070c3c79a1b5954c1573ef6c386b8972](https://github.com/bgould/drivers/commit/91236352070c3c79a1b5954c1573ef6c386b8972) changes
- Added example of controlling RGB LED connected via NINA chip
- Verified on nano-rp2040

Thanks, @bgould

### DIFF
--- a/examples/wifinina/pins/main.go
+++ b/examples/wifinina/pins/main.go
@@ -1,0 +1,91 @@
+//go:build nano_rp2040
+// +build nano_rp2040
+
+// This examples shows how to control RGB LED connected to
+// NINA-W102 chip on Arduino Nano RP2040 Connect board
+// Built-in LED code added for API comparison
+
+package main
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/wifinina"
+)
+
+const (
+	LED = machine.LED
+
+	// Arduino Nano RP2040 Connect board RGB LED pins
+	// See https://docs.arduino.cc/static/3525d638b5c76a2d19588d6b41cd02a0/ABX00053-full-pinout.pdf
+	LED_R wifinina.Pin = 27
+	LED_G wifinina.Pin = 25
+	LED_B wifinina.Pin = 26
+)
+
+var (
+
+	// these are the default pins for the Arduino Nano-RP2040 Connect
+	spi = machine.NINA_SPI
+
+	// this is the ESP chip that has the WIFININA firmware flashed on it
+	device *wifinina.Device
+)
+
+func setup() {
+
+	// Configure SPI for 8Mhz, Mode 0, MSB First
+	spi.Configure(machine.SPIConfig{
+		Frequency: 8 * 1e6,
+		SDO:       machine.NINA_SDO,
+		SDI:       machine.NINA_SDI,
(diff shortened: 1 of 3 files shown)


ysoldak:
https://user-images.githubusercontent.com/1061565/133912457-41db969d-8b30-4747-926f-d1f7ab0b4686.mov

wifinina/wifinina_nano-rp2040.go
```diff
@@ -0,0 +1,13 @@
+//go:build nano_rp2040
+// +build nano_rp2040
+
+package wifinina
+
+// Defines pins RGB LED is connected to on Arduino Nano RP2040 Connect board
+// See https://docs.arduino.cc/static/3525d638b5c76a2d19588d6b41cd02a0/ABX00053-full-pinout.pdf
+
+const (
+	LED_R = 27
```
soypat: I know this is not subject to future change, but should these definitions not be taken from `machine` package, i.e: `LED_R = machine.GP27` ?

ysoldak: Tricky here. You see, these pin numbers are not of RP2040 itself, but NINA-W102 chip.
Perhaps this shall be extracted to main tinygo repo, to `board_nano-rp2040.go` with names like `NINA_LED_R`.
But that shall be thought through properly, to avoid confusion and people trying to use these leds as built-in led, in fact accessing RP2040 pins instead of NINA's
See here https://docs.arduino.cc/static/3525d638b5c76a2d19588d6b41cd02a0/ABX00053-full-pinout.pdf

ysoldak:
@soypat question made me thinking of how we can make working with NINA pins easier and more like regular pins.

Second commit in this PR introduces this change. I like how it ended up, the single drawback is we have to keep global `wifinina.pinDevice` variable to control the peripheral pins.

Let me know what you think. Is consistency of API worth a global var or shall we keep it simple, like in first commit?
Depending on consensus, I'll either rollback last commit (and squash) or just squash them both as is, to make PR ready for merge.

@deadprogram that do you think, API wise?

deadprogram:
Would be good to implement

```go
func (p Pin) Set(value bool) error {
	if pinDevice == nil {
		return ErrPinNoDevice
	}
	return pinDevice.DigitalWrite(uint8(p), value)
}
```

Then your implementation for `High()` becomes something like:

```go
func (p Pin) High() error {
	return p.Set(true)
}
```

Same sort of thing for `Low()`.

ysoldak:
So, I've implemented your suggestion, @deadprogram 
Also, squashed commits.
The build fails though 🤷

deadprogram:
https://github.com/tinygo-org/tinygo/issues/2205 needs to be fixed.

ysoldak:
Green again 🎉

bgould:
I have a the airlift featherwing with me so I will try to get this running in the next few days and test it so that it can be merged.  Thanks @ysoldak

deadprogram:
Looking good, so now merging. Thanks @ysoldak for taking this over from @bgould

PR #319, opened 2021-09-22
i2csoft: add support for software I2C

sago35:
Added I2C software implementation for esp8266, esp32, etc.
It can also be used for atsamd51 and others, but be aware that it is not a very accurate process.

Since the esp32 has hardware I2C, it should really implement hardware I2C instead of SI2C.
However, I thought I would lose motivation if I implemented the hardware I2C driver first, so I implemented the software I2C first.

https://github.com/tinygo-org/tinygo/pull/2125

### DIFF
--- a/i2csoft/i2csoft.go
+++ b/i2csoft/i2csoft.go
@@ -53,10 +53,10 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 	}
 
 	// enable pins
-	i2c.sda.High()
 	i2c.sda.Configure(machine.PinConfig{Mode: machine.PinOutput})
-	i2c.scl.High()
+	i2c.sda.High()
 	i2c.scl.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	i2c.scl.High()
 
 	return nil
 }
--- a/i2csoft/i2csoft.go
+++ b/i2csoft/i2csoft.go
@@ -72,7 +72,6 @@ func (i2c *I2C) SetBaudRate(br uint32) {
 // It clocks out the given address, writes the bytes in w, reads back len(r)
 // bytes and stores them in r, and generates a stop condition on the bus.
 func (i2c *I2C) Tx(addr uint16, w, r []byte) error {
-	var err error
 	i2c.nack = false
 	if len(w) != 0 {
 		// send start/address for write
@@ -88,16 +87,10 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) error {
 
 		// write data
 		for _, b := range w {
-			err = i2c.WriteByte(b)
-			if err != nil {
-				return err
-			}
+			i2c.writeByte(b)
 		}
 
-		err = i2c.signalStop()
-		if err != nil {
-			return err
-		}
+		i2c.signalStop()
 	}
 	if len(r) != 0 {
 		// send start/address for read
@@ -125,17 +118,14 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) error {
 		// Send NACK to end transmission
 		i2c.sendNack()
 
-		err = i2c.signalStop()
-		if err != nil {
-			return err
-		}
+		i2c.signalStop()
 	}
 
 	return nil
 }
 
-// WriteByte writes a single byte to the I2C bus.
-func (i2c *I2C) WriteByte(data byte) error {
+// writeByte writes a single byte to the I2C bus.
+func (i2c *I2C) writeByte(data byte) {
 	// Send data byte
 	i2c.scl.Low()
 	i2c.sda.High()
@@ -168,12 +158,10 @@ func (i2c *I2C) WriteByte(data byte) error {
 	i2c.wait()
 
 	// wait until transmission successful
-
-	return nil
 }
 
 // sendAddress sends the address and start signal
-func (i2c *I2C) sendAddress(address uint16, write bool) error {
+func (i2c *I2C) sendAddress(address uint16, write bool) {
 	data := (address << 1)
 	if !write {
 		data |= 1 // set read flag
@@ -209,11 +197,9 @@ func (i2c *I2C) sendAddress(address uint16, write bool) error {
 	i2c.wait()
 
 	// wait until bus ready
-
-	return nil
 }
 
-func (i2c *I2C) signalStop() error {
+func (i2c *I2C) signalStop() {
 	i2c.scl.Low()
 	i2c.sda.Low()
 	i2c.sda.Configure(machine.PinConfig{Mode: machine.PinOutput})
@@ -225,10 +211,9 @@ func (i2c *I2C) signalStop() error {
 	i2c.sda.High()
 	i2c.wait()
 	i2c.wait()
-	return nil
 }
 
-func (i2c *I2C) signalRead() error {
+func (i2c *I2C) signalRead() {
 	i2c.wait()
 	i2c.wait()
 	i2c.scl.Low()
@@ -239,7 +224,6 @@ func (i2c *I2C) signalRead() error {
 	i2c.scl.High()
 	i2c.wait()
 	i2c.wait()
-	return nil
 }
 
 func (i2c *I2C) readByte() byte {
@@ -259,7 +243,7 @@ func (i2c *I2C) readByte() byte {
 	return data
 }
 
-func (i2c *I2C) sendNack() error {
+func (i2c *I2C) sendNack() {
 	i2c.wait()
 	i2c.wait()
 	i2c.scl.Low()
@@ -270,7 +254,6 @@ func (i2c *I2C) sendNack() error {
 	i2c.scl.High()
 	i2c.wait()
 	i2c.wait()
-	return nil
 }
 
 // WriteRegister transmits first the register and then the data to the
--- a/i2csoft/i2csoft.go
+++ b/i2csoft/i2csoft.go
@@ -1,0 +1,297 @@
+package i2csoft
+
+import (
+	"errors"
+	"machine"
+)
+
+// I2C is an I2C implementation by Software. Since it is implemented by
+// software, it can be used with microcontrollers that do not have I2C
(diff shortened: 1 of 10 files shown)


sago35:
The code has been cleaned up for the first version.
Also, added an argument to New() to make it compatible with other drivers.

i2csoft/i2csoft_atsamd51.go
```diff
@@ -0,0 +1,23 @@
...
+package i2csoft
+
+import (
+	"device/arm"
+)
+
+func (i2c *I2C) wait() {
+	// atsamd51
+	//  1 : about 388kHz
+	// 17 : about 97kHz
```
sago35: The setting values for samd51 (120MHz) and samd21 (48MHz) are now almost identical. Is it possible that samd21 is faster? Or is samd51 slower?

sago35:
Now ready for review.

i2csoft/i2csoft.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+// New returns the i2csoft driver. For the arguments, specify the pins to be
+// used as SCL and SDA. As I2C is implemented in software, any GPIO pin can be
+// specified.
+func New(sclPin, sdaPin machine.Pin) *I2C {
+	return &I2C{
+		scl:      sclPin,
+		sda:      sdaPin,
+		baudrate: 100 * 1e3,
```
aykevl: You can also write this as `100e3`, or even `100_000`. I think that's clearer than `100 * 1e3`.

i2csoft/i2csoft.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// This exists for compatibility with machine.I2CConfig. SCL and SDA must
+	// be set at the same time. Because Pin(0) is sometimes set, it is not
+	// checked for 0.
+	if config.SCL != config.SDA {
+		i2c.scl = config.SCL
+		i2c.sda = config.SDA
+	}
+
+	i2c.SetBaudRate(config.Frequency)
```
aykevl: Now you're setting the baudrate again? First here:

```go
	if config.Frequency != 0 {
		i2c.baudrate = config.Frequency
	}
```

And now here, but unconditionally.

i2csoft/i2csoft.go
```diff
@@ -0,0 +1,297 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// WriteByte writes a single byte to the I2C bus.
+func (i2c *I2C) WriteByte(data byte) error {
```
aykevl: This is an unexported function that cannot return an error, so I think the `error` return can be removed.

i2csoft/i2csoft_atsamd51.go
aykevl: What do these values mean?

i2csoft/i2csoft_atsamd51.go
```diff
@@ -0,0 +1,23 @@
...
+
+	wait := 1
+	if i2c.baudrate < 400*1e3 {
+		wait = 17
+	}
+
+	for i := 0; i < wait; i++ {
+		arm.Asm(`nop`)
+	}
```
aykevl: The loop itself will likely be a bigger overhead than the `nop` instruction itself. You can perhaps make it more consistent with some more nops:

```go
for i := 0; i < wait; i++ {
    arm.Asm(`
        nop
        nop
        nop
        nop`)
}
```

i2csoft/i2csoft_esp32.go
```diff
@@ -0,0 +1,23 @@
...
+	// 10 : about 387kHz
+	// 56 : about 99kHz
+
+	wait := 10
+	if i2c.baudrate < 400*1e3 {
+		wait = 56
+	}
+
+	for i := 0; i < wait; i++ {
+		device.Asm(`nop`)
```
aykevl, line 15: The `nop` instruction is supported in nearly every architecture, so if that helps you can probably use `device.Asm("nop")` in a shared file (this is also supported on ARM).

aykevl (commented):
Thank you for this PR! I've been hoping someone would make it for a while :)

One suggestion: perhaps it's easier to only support 100kHz. That should simplify the timing logic and is probably supported by most/all I2C devices.

> Since the esp32 has hardware I2C, it should really implement hardware I2C instead of SI2C.
However, I thought I would lose motivation if I implemented the hardware I2C driver first, so I implemented the software I2C first.

It is still useful, for example if you need more I2C peripherals than the hardware supports.

i2csoft/i2csoft.go
sago35: Changed to 100e3.

i2csoft/i2csoft.go
sago35: I'm wrong.
I will correct it.

i2csoft/i2csoft.go
sago35: Changed WriteByte to a private function.
Removed the return values of writeByte and sendAddress.

i2csoft/i2csoft_atsamd51.go
sago35: That was my note for wait.
I have corrected the comment.

i2csoft/i2csoft_atsamd51.go
sago35: I don't think overhead is going to be a problem now.
The reason is that it is easier to adjust if the process is completed in as short a time as possible.

sago35:
I have corrected the part you commented on.
Please review again.

sago35:
There was a problem with gofmt, which has been fixed.
I'm still using go1.16, but I didn't notice it because it's not an error in go1.16.

sago35:
It would be great if you could review it. @aykevl

i2csoft/i2csoft.go
```diff
@@ -0,0 +1,280 @@
...
+	if config.SCL != config.SDA {
+		i2c.scl = config.SCL
+		i2c.sda = config.SDA
+	}
+
+	// enable pins
+	i2c.sda.High()
+	i2c.sda.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	i2c.scl.High()
+	i2c.scl.Configure(machine.PinConfig{Mode: machine.PinOutput})
```
aykevl: Note that calling `.High()` may not do what you expect before calling `.Configure()`. The API says that `.Configure()` must be called first: https://tinygo.org/docs/reference/machine/#gpio

aykevl (changes requested):
I think this driver would be much more useful if it didn't contain code for each chip separately.

You can actually write a generic version quite easily. If you want to make the wait time 100kHz, and the chip is 16MHz, that means you need to wait for 160 cycles. So you can calculate the number of cycles quite easily:

```go
cycles := machine.CPUFrequency() / 100e3
```

Then you can wait for the correct amount of cycles. Something like this:

```go
for i := cycles / 8; i != 0; i-- {
    device.Asm("nop")
    device.Asm("nop")
    device.Asm("nop")
    device.Asm("nop")
    device.Asm("nop")
    device.Asm("nop")
}
```

Every common RISC chip will need a decrement and a branch, so will take at least two cycles in loop overhead. I'm also doing the loop "backwards" because that's often more efficient (the compiler might not do this optimization). In the case of a 16MHz chip, 160 cycles are needed. Here, we do 160/8=20 repetitions of this loop, which should be enough time.
Because I2C isn't very strict about timing, I think this will be precise enough.

The big advantage is of course that you don't need to write the same code for every chip separately so it is truly portable (the main goal of this i2csoft driver).

sago35:
@aykevl 
For several targets, I measured with the following code.
The results are as follows, and we cannot make a good wait time from the CPU clock.

For i2c, the clock accuracy is not important, but since it is blocking, it may want to send the data as fast as possible.
I think m4's Frequency is very slow, which is a problem.

| target | clock | I2C Frequency |
| ------ | ----- | --------------- |
| feather-rp2040 | 125MHz | 18kHz |
| feather-m4-can | 120MHz | 9kHz |
| feather-stm32f405 | 168MHz | 19kHz |
| qtpy (m0) | 48Mhz | 17kHz |
| feather-nrf52840-sense | 64MHz | 19kHz |

```go
package i2csoft

import (
	"device"
	"machine"
)

// wait waits for half the time of the SCL operation interval.
func (i2c *I2C) wait() {
	cycles := machine.CPUFrequency() / 100e3

	for i := cycles / 8; i != 0; i-- {
		device.Asm("nop")
		device.Asm("nop")
		device.Asm("nop")
		device.Asm("nop")
		device.Asm("nop")
		device.Asm("nop")
	}
}
```

i2csoft/i2csoft.go
sago35: I've updated the code.

sago35:
I tried changing the number of nop.
Is the overhead of device.Asm() too large?

* feather-rp2040
    * nop x 24 : 7kHz
    * nop x 12 : 12kHz
    * nop x 6 : 18kHz
    * nop x 1 : 34kHz

```go
package i2csoft

import (
	"device"
	"machine"
)

// wait waits for half the time of the SCL operation interval.
func (i2c *I2C) wait() {
	cycles := machine.CPUFrequency() / 100e3

	for i := cycles / 8; i != 0; i-- {
		device.Asm("nop")
	}
}
```

deadprogram:
Are there any other changes pending on this PR? @sago35 @aykevl it seems like most, but not all, of the questions have been addressed or answered. What more is needed here?

sago35:
@deadprogram @aykevl 
The change request from aykevl is to remove the handling of each chip.
However, at this point, I think it is better to keep the per-chip handling.
In particular, the performance of atsamd51 is very slow.
The slower the Frequency, the longer i2csoft's processing will be blocked.

Removing per-chip handling should be considered in another PR.

deadprogram:
> Removing per-chip handling should be considered in another PR.

Works for me!

Thanks for all your efforts on this PR @sago35 and thanks for reviewing @aykevl 

Now merging.

PR #320, opened 2021-09-24
espnet: WIP support for on-chip WiFi on an ESP32C3

aykevl:
Work in progress. Does not work yet.

Some notes:

  - This requires some changes to TinyGo, look at the espnet branch: https://github.com/tinygo-org/tinygo/pull/2138
  - The  next step is probably defining all the functions in `g_wifi_osi_funcs` (see espnet.c). Right now it hangs in `esp_wifi_init_internal`, probably a NULL pointer dereference.
  - This is only for the ESP32-C3. This will require some work to work on other chips from Espressif.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "espnet/esp-idf"]
+	path = espnet/esp-idf
+	url = https://github.com/espressif/esp-idf.git
--- a/espnet/ap.go
+++ b/espnet/ap.go
@@ -28,10 +28,17 @@ type Config struct {
 }
 
 var internalConfig = C.wifi_init_config_t{
-	osi_funcs: &C.g_wifi_osi_funcs,
+	osi_funcs:           &C.g_wifi_osi_funcs,
+	wpa_crypto_funcs:    C.g_wifi_default_wpa_crypto_funcs,
+	static_rx_buf_num:   10,
+	static_tx_buf_num:   10,
+	mgmt_sbuf_num:       6,
+	sta_disconnected_pm: true,
+	magic:               C.WIFI_INIT_CONFIG_MAGIC,
 }
 
 func (wifi ESPWiFi) Configure(config Config) error {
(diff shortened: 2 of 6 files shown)


aykevl:
Update: the function that's called is `recursive_mutex_create_wrapper`, which calls `xSemaphoreCreateRecursiveMutex`. I guess this will need to be added to the FreeRTOS compatibility layer I've started to write in https://github.com/tinygo-org/tinygo/pull/2138.

joelschutz:
Is this dead?

aykevl:
I'm not currently working on it.

aykevl:
Just found https://github.com/espressif/esp-wireless-drivers-3rdparty (thanks @panbanda!) which we should use as a basis for this, instead of esp-idf.

DatanoiseTV:
Are there any news on this? I have lots of "stock" ESP32/S2/S3 boards for testing and also some of my own ones.

sprive:
Slightly connected questions:

1. are there any microcontrollers _at all_ with TinyGo wifi support?
2. What workarounds exist? If someone's seen a good workaround, like using a second microcontroller as a wifi peripheral, or using Bluetooth networking.. would you mind dropping a link (or DM it)? 

It seems like there's a lot of interest, not only in resolving the TinyGo issue, but even in finding workarounds for the interim. Cheers

b0ch3nski:
> 1. are there any microcontrollers _at all_ with TinyGo wifi support?

@sprive There is [Nano RP2040 Connect](https://docs.arduino.cc/hardware/nano-rp2040-connect) which use `wifinina` driver - see [webserver example](https://github.com/tinygo-org/drivers/blob/dev/examples/wifinina/webserver/main.go) and a super cool project - [plantbot](https://github.com/ysoldak/plantbot).

aykevl:
See this PR for a new attempt, using much more suitable libraries: https://github.com/tinygo-org/drivers/pull/650

Issue #310, opened 2021-09-09
Proposal: `Update` method call to read sensor data from a bus


soypat:
## Background
Sensors acquire measurements from their surrounding and store them as bytes in their registers for us to read via a communications bus, such as (but not limited to) i2c and SPI. Reading sensor data using these protocols carries some overhead that should be minimized if possible. This proposal seeks to provide TinyGo with a lean foundation to build sensing libraries upon.

## The `Update` method
The benefits of having an `Update` method to schedule sensor measurements was discussed in PR #298 and #299. 

**Put simply, the `Update` method reads all requested sensor values into memory for later use using `Measurement` methods**. Subsequent calls to update replace measurements.

The goal of this proposal is to clearly separate the I/O from the data. Most ICs work with auto-increment when reading registers which allows reading many sensor values in one transaction.

The benefits are many:

## Advantages
* Better framework for libraries. Without this change if many libraries use the same sensor then they'd all be fighting for the communications bus. On top of that, this bus abuse gets worse if there are many sensors on the same bus. Splitting I/O and data responsibilities lets the user decide if they delegate I/O to the library, if they call it once every `loop()` or if they call it every couple minutes
* Error management separate from "business logic". So far we've been creating these sensor handlers which have data only methods, such as an ADC's [`Get() uint16`](https://github.com/tinygo-org/drivers/blob/release/mcp3008/mcp3008.go), an IMU's [acceleration and rotation](https://github.com/tinygo-org/drivers/blob/release/mpu6050/mpu6050.go#L43) methods. We've been discarding all bus errors doing this and that is objectively not good design. These errors should be handled somewhere in the stack. By using an `Update()` method the error can be handled explicitly and we can still have a clean API on our sensors.
* Less overhead of read/write function call and protocol transaction means better performance, more sensor data per unit time
* Less chance of a failed read or timeout due to interference. The MPU6050 is notorious for this at high request rates.
* This is how most ICs are meant to be used. i.e: Some ADC's work faster in continuous conversion mode, but this means that the channel you desire might not be the one you get when reading the data register so `for` loops are used to request data until the desired channel is read, which is not guaranteed to happen in a fixed amount of time. 

## Disadvantage
* Peripheral structure consumes data equivalent to total sensor data.
* A call to `Update` must be done before calling a measurement function though this could be remedied by embedding the type within a sample rate limiter.

@aykevl Showed an example of what the API could look like

```go
// in the drivers package
type Measurement uint32
const (
[...]

PR #321, opened 2021-09-26
add Sensor interface

soypat:
Hey there peeps, here's part of the `Update` method proposal. This comes with a `Sensor` interface which implements said method. This interface then could be embedded in all subsequent sensor interface types, i.e:
```go
type Sensor interface {
	Update(which Measurement) error
}

type IMU interface {
    Sensor
    AngularVelocity() (gx, gy, gz int32)
}

type Barometer interface {
    Sensor
    Pressure() int32
}
```

These changes were first discussed in [`add AHRS interfaces for vehicle attitude control`](https://github.com/tinygo-org/drivers/pull/299) and then added as a formal proposal in [`Proposal: Update method call to read sensor data from a bus`](https://github.com/tinygo-org/drivers/issues/310). There has been little to no discussion on the matter. I will restate some of the benefits of the `Update` proposal:

1. Probably most importantly, error management: Where before errors were discarded now they are formally dealt with.
2. Less bus pressure
3. Idiomatic development of sensor frameworks. Provides us with high level interfaces useful for testing.

To be clear, I think we should still have the `ReadMeasurement` methods which perform an update followed by measurement read which are usually much simpler to understand and work with for people who are new to embedded systems.

### DIFF
--- a/sensor.go
+++ b/sensor.go
@@ -1,0 +1,31 @@
+package drivers
+
+// Measurement specifies a type of measurement,
+// for example: temperature, acceleration, pressure.
+type Measurement uint32
+
+// Sensor measurements
+const (
+	Voltage Measurement = 1 << iota
+	Temperature
+	Humidity
+	Pressure
+	Distance
+	Acceleration
+	AngularVelocity
+	MagneticField
+	Luminosity
+	Time
+
+	AllMeasurements Measurement = 0xffffffff
+)
+
+// Sensor represents an object capable of making one
+// or more measurements. A sensor will then have methods
+// which read the last updated measurements.
+//
+// Many Sensors may be collected into
+// one Sensor interface to synchronize measurements.
+type Sensor interface {
+	Update(which Measurement) error
+}


aykevl:
I like this proposal!

I'm not entirely sure about the `Audio` measurement type, because audio is a bit different from the others. Unlike things like temperature, humidity, or time, audio requires a large amount of data to move around and is very time sensitive (a delay of 1ms may be far too long). However, that's only one detail.

@deadprogram @conejoninja @ysoldak @sago35 any opinion on this?

ysoldak:
So, for say, Barometer, we gonna have 3 methods in a typical driver implementation:
```
Update(which Measurement) error // update
Pressure() int32                // read
ReadPressure() (int32, error)   // update & read
```

Will it not be confusing?
Or we willing to drop `ReadPressure()`?
Shall we standardise on error emitted then incompatible Measurement type requested?

soypat:
@aykevl I'll remove the Audio type now. We can always add it later.

@ysoldak Maybe we can discard the `ReadMeasurement` methods, it does seem to lend itself to being confusing having two methods with similar name and functionality. There are some benefits to this method:
* Provides a "cleaner" APIs by not having a returned error method as is the case for most drivers today in this repo. i.e `ReadPressure() (microbar int32)`
* It's easier to understand if you are a newcomer to embedded systems
* Less importantly, if one prefers a memory optimized implementation, `ReadMeasurement` method requires no data to be stored before passing it to user.  Nowadays I think storing a couple uint32's should not be a problem on modern controllers.

ysoldak:
I wonder if we could have a feasibility study.

Branch from this PR and try and use this proposed interface in existing drivers.
Do not need to cover all the drivers, but at least several and more diverse -- the better.
Make a new related PR to show how it gonna look like.
Without this ground work it is hard to settle on anything, IMHO. Too many corner cases may pop-up that can be awkward to workaround or fix later.

soypat:
Closing in favor of #345.

PR #322, opened 2021-09-27
Espnet

zdima:
@aykevl added missing functions and added rudimental mutex to continue with call sequence

### DIFF
--- a/espnet/espnet.c
+++ b/espnet/espnet.c
@@ -70,6 +70,7 @@ static void *_mutex_create(void) {
 	return NULL;
 }
 
+// simplified mutex to continue with call stack
 unsigned int mutx = 0;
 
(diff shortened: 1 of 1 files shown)


aykevl:
Thank you!
I'm now working on a full implementation of FreeRTOS recursive mutexes.

zdima:
> Thank you!
> I'm now working on a full implementation of FreeRTOS recursive mutexes.
The next call is memory allocation.
What can be done to implement memory calls?
Tinygo assumes all memory get allocated in go when calling c-go?

aykevl:
Oops, I missed your comment! I've just now implemented memory allocation. You will need to update TinyGo to the latest version of the espnet branch. I force-pushed, so you need to run something like `git checkout espnet && git fetch origin && git reset --hard origin/espnet`.

PR #328, opened 2021-10-12
DHTXX: Use runtime interrupt instead of UART

### LINKED ISSUE #255: DHTXX driver does not work on Arduino Uno
When trying to build the example for the dht driver (https://github.com/tinygo-org/drivers/blob/release/examples/dht/main.go) on Windows I get the following error:

```
> tinygo flash -target arduino -port COM3 main.go
# tinygo.org/x/drivers/dht
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.15.1\dht\thermometer.go:160:10: UART1 not declared by package machine
..\..\go\pkg\mod\tinygo.org\x\drivers@v0.15.1\dht\thermometer.go:161:16: UART1 not declared by package machine
```

Taking a quick look at the code for the driver, it tries to access UART1 https://github.com/tinygo-org/drivers/blob/release/dht/thermometer.go#L160, which according to the `machine` package for arduino is not defined (https://tinygo.org/microcontrollers/machine/arduino/).

I tried the dirty hack of using UART or UART0 as defined in the machine package for arduino, but they don't have a method `Interrupt` as used by the driver.

Cheers!

Elfo404:
Implementing what was being suggested in https://github.com/tinygo-org/drivers/issues/255#issuecomment-852140542.

i quickly tested this on a `itsybitsy-nrf52840` and seems to work fine.

(Should) fix #255

shall I also add a smoke test for this? something like:

```
tinygo build -size short -o ./build/test.hex -target=itsybitsy-nrf52840 ./examples/dht/main.go
```

### DIFF
--- a/examples/dht/main.go
+++ b/examples/dht/main.go
@@ -12,7 +12,7 @@ func main() {
 	dhtSensor := dht.New(pin, dht.DHT11)
 	for {
 		temp, hum, err := dhtSensor.Measurements()
-		if err != nil {
+		if err == nil {
 			fmt.Printf("Temperature: %02d.%d°C, Humidity: %02d.%d%%\n", temp/10, temp%10, hum/10, hum%10)
 		} else {
 			fmt.Printf("Could not take measurements from the sensor: %s\n", err.Error())
--- a/dht/thermometer.go
+++ b/dht/thermometer.go
@@ -8,6 +8,7 @@ package dht // import "tinygo.org/x/drivers/dht"
 
 import (
 	"machine"
+	"runtime/interrupt"
 	"time"
 )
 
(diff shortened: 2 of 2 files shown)


didiermichel:
Hi,
I have tried this with `arduino` target and I got folowing error at compilation time:
```
error: interp: ptrtoint integer size does not equal pointer size
```
Not sure how to analyse this.
it seems related to tinygo-org issue 2389

doniacld:
Hi @Elfo404 
I spent some time struggling with why the DHT example was not printing any data. Do you think you could do a separate PR with your commit fixing the example? Since it is a quick fix and could be merged easily?!
Thanks a lot

examples/dht/main.go
```diff
@@ -12,7 +12,7 @@ func main() {
 	dhtSensor := dht.New(pin, dht.DHT11)
 	for {
 		temp, hum, err := dhtSensor.Measurements()
-		if err != nil {
+		if err == nil {
```
doniacld, line 15: Great catch!

Elfo404:
@doniacld here you go :) https://github.com/tinygo-org/drivers/pull/381

stevegt:
> Not sure how to analyse this. it seems related to tinygo-org issue 2389

@didiermichel Yes, does sound like tinygo-org/tinygo#2389.

PR #330, opened 2021-10-20
ws2812: add support for HiFive1 and ESP32-C3

aykevl:
Like the title says.

ESP32-C3 support should work right away. HIFive1 support depends on https://github.com/tinygo-org/tinygo/pull/2188 and https://github.com/tinygo-org/tinygo/pull/2189.


deadprogram:
The prereqs have been merged, so now merging this PR as well. Thanks @aykevl :traffic_light:

PR #333, opened 2021-10-22
drivers: add HC-SR04 ultrasonic distance sensor

soypat:
PTAL. Tested with raspberry pico. A little wonky but gets the job done.


soypat:
For some reason first time I checked I did not see the existing HCSR04 driver. sorry!

PR #335, opened 2021-10-26
Sensor interface PR proof of concept

soypat:
Proof of concept for #321. WIP
* BMP180 pressure sensor implementation
* MCP3008 ADC implementation
* #298 implements interface for MPU6050 acceleration/angular velocity sensor

### DIFF
--- a/bmp180/bmp180.go
+++ b/bmp180/bmp180.go
@@ -36,6 +36,8 @@ type Device struct {
 	Address                 uint16
 	mode                    OversamplingMode
 	calibrationCoefficients calibrationCoefficients
+	temp                    int32
+	pressure                int32
 }
 
 // New creates a new BMP180 connection. The I2C bus must already be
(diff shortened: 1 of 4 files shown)


bmp180/bmp180.go
```diff
@@ -80,6 +82,58 @@ func (d *Device) Configure() {
...
 
+// Pressure returns the pressure in milli pascal (mPa).
+func (d *Device) Pressure() int32 { return d.pressure }
+
+// Temperature returns the temperature in celsius milli degrees (°C/1000).
+func (d *Device) Temperature() int32 { return d.temp }
+
+func (d *Device) Update(which drivers.Measurement) (err error) {
+	// Temperature is needed for presssure calculation.
+	if which|drivers.Temperature|drivers.Pressure == 0 {
```
ysoldak, line 93: Shall it not return some sort of error when measurement requested is not temperature or pressure?

mcp3008/buffered.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	d.cs.Low()
+	// This supposes the ADC supports autoincrement like for example the mcp3464 ADC (UNTESTED)
+	err := d.bus.Tx(d.tx[:], d.rx[:])
+	d.cs.High()
+	return err
+}
+
+// Channel returns an ADC measurement for a specific channel. Must call Update Beforehand
+func (d *BufferedDev) Channel(ch int) uint16 {
```
ysoldak, line 90: Why this name? Shall be "Voltage"?

mcp3008/buffered.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Configure sets up the device for communication
+func (d *BufferedDev) Configure() {
+	d.cs.Configure(machine.PinConfig{Mode: machine.PinOutput})
+}
+
+// UNTESTED
+func (d *BufferedDev) Update(which drivers.Measurement) error {
+	if which|drivers.Voltage == 0 {
```
ysoldak, line 59: Right now it only reads voltage, can read more things?
If only voltage ever, why have `which drivers.Measurement` at all?

bmp180/bmp180.go
soypat: The reason behind it not being an error is to allow for grouping of Sensors.
```go
allSensors := []Sensor{imu, bmp, tempSens}
updateAll := func (w Measurement) {
   for _, sens := range allSensors {
     sens.Update(w) // illustrative, please handle your errors
   }
}
```
This way you can have precise requests of what sensor measurements you want to update. This is especially useful if the characteristic time of a sensor is very slow compared to other sensors in the group, i.e Temperature sensor vs. voltage sensor.

mcp3008/buffered.go
soypat: So maybe this name should be Get, the reason it is not Voltage is because it does not represent the voltage, merely the bits read by the ADC. I have added the Voltage method separate as a conversion of Channel.

mcp3008/buffered.go
soypat: See comment above

mcp3008/mcp3008.go
```diff
@@ -83,7 +83,7 @@ func (p ADCPin) Get() uint16 {
 	p.d.bus.Tx(p.d.tx, p.d.rx)
 
 	// scale result to 16bit value like other ADCs
-	result := uint16((p.d.rx[1]&0x3))<<8 + uint16(p.d.rx[2])<<6
+	result := uint16((p.d.rx[1]&0x3))<<(8+6) + uint16(p.d.rx[2])<<6
```
soypat, line 86: @deadprogram I believe I found a bug here which I took the liberty to patch. While scaling you were writing the MSB bits over the LSB bits. I've tested this code and it now works as expected by having the largest possible output be `0x3FF<<6` or `1024<<6`

bmp180/bmp180.go
soypat: It might be worth thinking about this a bit more. Maybe it's worth having a `io.EOF` sentinel like error for the case you mention Yurii, for example: `drivers.ErrNoMeasurement`.

mcp3008/mcp3008.go
deadprogram: @soypat do you want to submit specific PR with a fix?

PR #338, opened 2021-11-12
lsm9ds1: minimal implementation

ysoldak:
Depends on [tinygo PR exposing additional pins on nano-33-ble board](https://github.com/tinygo-org/tinygo/pull/2256) to compile example.

@alankrantas Something of interest for you to check?
Seem like we are both working on Nano 33 BLE support at the moment.

IMU communication does not work for me until I2C_PULLUP pin is set high, something I don't see in your examples for other sensors, @alankrantas And that makes me wonder... Any ideas?

I2C pull-up trick was copied from Arduino Mbed code for this board
https://github.com/arduino/ArduinoCore-mbed/blob/master/variants/ARDUINO_NANO33BLE/variant.cpp#L159

### DIFF
--- a/examples/lsm9ds1/main.go
+++ b/examples/lsm9ds1/main.go
@@ -1,0 +1,103 @@
+// LSM9DS1, 9 axis Inertial Measurement Unit (IMU)
+package main
+
+import (
+	"fmt"
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/lsm9ds1"
+)
+
+const (
+	PLOTTER             = false
+	SHOW_ACCELERATION   = true
+	SHOW_ROTATION       = true
+	SHOW_MAGNETIC_FIELD = true
+	SHOW_TEMPERATURE    = true
+)
+
+func main() {
+
+	// I2C configure
+	machine.I2C0.Configure(machine.I2CConfig{})
+
+	// LSM9DS1 setup
+	device := lsm9ds1.New(machine.I2C0)
+	err := device.Configure(lsm9ds1.Configuration{
+		AccelRange:      lsm9ds1.ACCEL_2G,
+		AccelSampleRate: lsm9ds1.ACCEL_SR_119,
+		GyroRange:       lsm9ds1.GYRO_250DPS,
+		GyroSampleRate:  lsm9ds1.GYRO_SR_119,
+		MagRange:        lsm9ds1.MAG_4G,
+		MagSampleRate:   lsm9ds1.MAG_SR_40,
+	})
+	if err != nil {
+		for {
+			println("Failed to configure", err.Error())
+			time.Sleep(time.Second)
+		}
+	}
+
+	for {
+
+		if con, err := device.Connected(); !con || err != nil {
+			println("LSM9DS1 not connected")
+			time.Sleep(time.Second)
+			continue
+		}
+
+		ax, ay, az, _ := device.ReadAcceleration()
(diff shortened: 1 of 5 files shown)


alankrantas:
I hid the pull-ups in the driver like [this](https://github.com/tinygo-org/drivers/blob/dev/apds9960/apds9960_nano_33_ble.go). There are two version of ```New()```, one for build tag ```nano_33_ble``` and one for everything else (since there are separate sensor boards being sold). So the users do not need to do anything on their end.

There were discussion that should we move these functionality to TinyGo itself, but so far there's no better solution. In theory the user can also turn off these pins by themselves in order to save power (and they would have to initialize a new driver instance when they need it again).

I did looked at LSM9DS1 at that time, but the two ```WHO_AM_I``` registers never return corrected values to me and changes from time to time. It seems to be [wired](https://content.arduino.cc/assets/NANO33BLE_V2.0_sch.pdf) just like every other sensors. I can't figured out why. If you can make it work that would certainly be great!

ysoldak:
Cool, good solution. I'll do the same then. Thanks!

Yes, lsm9ds1 on nano 33 ble works for me, no prob.

ysoldak:
Modified and rebased.

@alankrantas I've modified your workaround a bit, doing pin touching in Configure() instead of New(), since New() supposed to "only creates the Device object, it does not touch the device."

BTW, referenced PR in main tinygo repo was merged, we can use named pins in nano 33 ble sensor initialisation routines now.

alankrantas:
If the user calls ```Connected()``` before ```Configure()``` the I2C operation will hang (and so is the rest of the script). This is why I had to initialize the pins in ```New()```.

deadprogram:
I do not think you should call `Connected` before `Configure`. The convention is to call `New` to create the thing without actual hardware communication. As long as `Configure` is *not* called, you can run unit tests written using pure Go. A number of the `i2c` devices are setup this way already.

alankrantas:
Okay, sorry that I misinterpreted ```Connected()``` that way, especially no one ever told me the proper convention before. I always thought that calling ```who_am_i``` registers would be a better first line of checking before doing everything else in ```Configure()```.

Please fix the other drivers as you see fit.

ysoldak:
Thanks, @deadprogram for clarifications!

@alankrantas interesting idea to check `who_am_i` first thing in Configure. Sort of sanity check? It sounds reasonable, not sure if it makes any difference though? Like if device responds with wrong value, shall we error out? Or it does not respond at all? Sounds rather like a hardware issue that should be fixed than try and handle it in software?

alankrantas:
To be fair, ```who_am_i``` check is not strictly necessary if you already expect the user to connect the device correctly (or it is present on the board), and the device/I2C bus are not damaged in anyway. So you can either integrate ```Connected()``` into ```Configure()``` or just get rid of it.

ysoldak:
I believe `Connected()` has its use and shall keep it.
The thing is in my other project, with Nano RP2040, I've observed I2C dies sometimes, and GPIO24 must be touched, first high, then low to make it work again. And to reliably test if I2C is still up, I do use `Connected()` (on lsm6dox) before reading accelerometer.

So I'll just call `Connected()` in `Configure()`, after pins touching, of course.

ysoldak:
Made described changes, plus added error propagation to make it easier to spot bus issues in application code.

deadprogram:
Thanks for everyone's work on this @ysoldak @alankrantas now merging.

PR #339, opened 2021-11-13
mcp3008: fix bitshift bug

soypat:
By request of @deadprogram. PTAL


deadprogram:
Thanks @soypat for the fix! Now merging.

PR #340, opened 2021-11-16
axp192: add support for AXP192 power system management IC

sago35:
This PR adds support for the AXP192 power management IC used in the M5Stack Core2.
The AXP192 is connected to the ESP32 via I2C.
In the M5Stack Core2, the LCD cannot be displayed without using the AXP192.
After this PR is merged, I will create a PR for the LCD driver.

* http://www.x-powers.com/en.php/Info/product_detail/article_id/29
* datasheet : https://github.com/m5stack/M5-Schematic/blob/master/Core/AXP192%20Datasheet_v1.1_en_draft_2211.pdf
* sch : https://m5stack.oss-cn-shenzhen.aliyuncs.com/resource/docs/schematic/Core/CORE2_V1.0_SCH.pdf
* pinout : https://docs.m5stack.com/en/core/core2

ref : https://github.com/tinygo-org/tinygo/pull/2150

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,12 +213,14 @@ endif
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.hex -target=feather-m4 ./examples/i2csoft/adt7410/
 	@md5sum ./build/test.hex
+	tinygo build -size short -o ./build/test.elf -target=wioterminal ./examples/axp192/m5stack-core2-blinky/
+	@md5sum ./build/test.hex
 
 DRIVERS = $(wildcard */)
 NOTESTS = build examples flash semihosting pcd8544 shiftregister st7789 microphone mcp3008 gps microbitmatrix \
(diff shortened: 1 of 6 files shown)


sago35:
Oops, I guess I forgot to add it to smoke-test.
I will add it now.
After adding it to smoke-test, it is ready for review.

sago35:
??
Huh? Is xtensa not available in drivers-repo?

```
tinygo build -size short -o ./build/test.elf -target=m5stack-core2 ./examples/axp192/m5stack-core2-blinky/
error: No available targets are compatible with triple "xtensa"
make: *** [Makefile:214: smoke-test] Error 1

Exited with code exit status 2

CircleCI received exit code 2
```

sago35:
I think this will pass the test.


* Changed smoke-test from m5stack-core2 to wioterminal
* README.md had not been updated, so it was added.

deadprogram:
Thanks for the addition @sago35 now merging.

PR #342, opened 2021-11-21
Add PCA9685 PWM driver

soypat:
Tested and works!


deadprogram:
Thanks @soypat now merging.

PR #344, opened 2021-11-26
Semtech sx126x Lora radio driver

ofauchon:
This is an early implementation of driver for Semtech sx126x radios .

This driver will support :

  * Standalone, SX126x modules, connected to MCUs through regular SPI buses, and NSS/RST/BUSY GPIOS
  * Sx126x radios embedded in SoC (ex : in STM32WL series ) 

When using sx126x in SoCs, you may not have 'regular' SPI BUS and GPIOs (but specific registers instead).
It may require specific initialization sequences too

The driver uses build tag to split the driver code, and low lever stuff: 

* sx126x.go : The  driver itself without low level / hardware code
* spi_stm32wl.go : Specific code for SPI/NSS/Busy/RST/initialization when using embedded SX126X from STM32WLxx SoC
* spi_generic.go : Wrapper for standard SPI and GPIOs access (When using 'regular' sx126x modules)

Although the driver may (almost) work, it's WIP. 

The PR is a Draft to get feedback on the implementation.

### DIFF
--- a/sx126x/spi_stm32wl.go
+++ b/sx126x/spi_stm32wl.go
@@ -1,0 +1,85 @@
+//go:build stm32wlx
+// +build stm32wlx
+
+package sx126x
+
+import (
+	"device/stm32"
+	"errors"
+	"machine"
+	"tinygo.org/x/drivers"
+)
+
+// New creates a new SX126x connection.
+func New(spi drivers.SPI) *Device {
+	c := make(chan RadioEvent, 10)
+	d := Device{
+		spi:            spi,
+		radioEventChan: c,
+	}
+	if d.spi == machine.SPI3 {
+		d.SubGhzInit()
+		d.SetDeviceType(DEVICE_TYPE_SX1262)
+	} else {
+		panic("Driver only support SUBGHZSPI (SPI3) on stm32wlx targets")
+	}
+	return &d
+}
+
+//SpiSetNss Sets the NSS line
+func (d *Device) SpiSetNss(state bool) {
+	if state {
+		stm32.PWR.SUBGHZSPICR.SetBits(stm32.PWR_SUBGHZSPICR_NSS)
+	} else {
+		stm32.PWR.SUBGHZSPICR.ClearBits(stm32.PWR_SUBGHZSPICR_NSS)
+
+	}
+}
+
+// WaitBusy sleep until all busy flags clears
+func (d *Device) WaitBusy() error {
+	count := 100
+	var rfbusyms, rfbusys bool
+	for count > 0 {
+		rfbusyms = stm32.PWR.SR2.HasBits(stm32.PWR_SR2_RFBUSYMS)
+		rfbusys = stm32.PWR.SR2.HasBits(stm32.PWR_SR2_RFBUSYS)
+
+		if !(rfbusyms && rfbusys) {
+			return nil
+		}
+		count--
+	}
+	return errors.New("WaitBusy Timeout")
+}
+
+// SubGhzInit() configures internal SX1262's SPI bus.
+func (d *Device) SubGhzInit() {
+
+	// Enable APB3 Periph clock and delay
+	stm32.RCC.APB3ENR.SetBits(stm32.RCC_APB3ENR_SUBGHZSPIEN)
+	_ = stm32.RCC.APB3ENR.Get()
+
+	// Disable radio reset and wait it's ready
+	stm32.RCC.CSR.ClearBits(stm32.RCC_CSR_RFRST)
+	for stm32.RCC.CSR.HasBits(stm32.RCC_CSR_RFRSTF) {
+	}
+
+	// Set NSS line low
+	stm32.PWR.SUBGHZSPICR.SetBits(stm32.PWR_SUBGHZSPICR_NSS)
+
+	// Enable radio busy wakeup from Standby for CPU
+	stm32.PWR.CR3.SetBits(stm32.PWR_CR3_EWRFBUSY)
+
+	// Clear busy flag
+	stm32.PWR.SCR.Set(stm32.PWR_SCR_CWRFBUSYF)
+
+	// Enable SUBGHZ Spi
+	// - /8 Prescaler
+	// - Software Slave Management (NSS)
+	// - FIFO Threshold and 8bit size
+	stm32.SPI3.CR1.ClearBits(stm32.SPI_CR1_SPE)
+	stm32.SPI3.CR1.Set(stm32.SPI_CR1_MSTR | stm32.SPI_CR1_SSI | (0b010 << 3) | stm32.SPI_CR1_SSM)
+	stm32.SPI3.CR2.Set(stm32.SPI_CR2_FRXTH | (0b111 << 8))
+	stm32.SPI3.CR1.SetBits(stm32.SPI_CR1_SPE)
+
+}
--- a/sx126x/sx126x.go
+++ b/sx126x/sx126x.go
@@ -1,0 +1,699 @@
+// Package sx126x provides a driver for SX126x LoRa transceivers.
+// Inspired from https://github.com/Lora-net/sx126x_driver/
+
+package sx126x
+
+import (
+	"errors"
+	"time"
+
+	"tinygo.org/x/drivers"
+)
+
+// SX126X radio transceiver RF_IN and RF_OUT may be connected
+// to RF Switch. This interface allows the creation of struct
+// that can drive the RF Switch (Used in Lora RX and Lora Tx)
+type RFSwitch interface {
+	InitRFSwitch()
+	SetRfSwitchMode(mode int) error
+}
+
+const (
+	DEVICE_TYPE_SX1261 = iota
+	DEVICE_TYPE_SX1262 = iota
+	DEVICE_TYPE_SX1268 = iota
+)
+
+const (
+	RFSWITCH_RX    = iota
+	RFSWITCH_TX_LP = iota
+	RFSWITCH_TX_HP = iota
+)
+
+const (
+	RadioEventRxDone    = iota
+	RadioEventTxDone    = iota
+	RadioEventTimeout   = iota
+	RadioEventWatchdog  = iota
+	RadioEventCrcError  = iota
+	RadioEventUnhandled = iota
+)
+
+// RadioEvent are used for communicating in the radio Event Channel
+type RadioEvent struct {
+	EventType int
+	IRQStatus uint16
+	EventData []byte
+}
+
+const (
+	PERIOD_PER_SEC  = (uint32)(1000000 / 15.625) // SX1261 DS 13.1.4
+	SPI_BUFFER_SIZE = 256
+)
+
+// Device wraps an SPI connection to a SX127x device.
+type Device struct {
+	spi            drivers.SPI     // SPI bus for module communication
+	radioEventChan chan RadioEvent // Channel for Receiving events
+	loraConf       LoraConfig      // Current Lora configuration
+	rfswitch       RFSwitch        // RF Switch, if any
+	deepSleep      bool            // Internal Sleep state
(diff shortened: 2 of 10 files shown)


ofauchon:
Hi. 
Could you please tell me if the implementation (with build tags) is the correct way to deal with 
the differents hardware scenarios ? 

Thx 

Olivier

deadprogram:
I have a couple of Arduino MKR 1300 boards on hand. They say they have this for the onboard LoraWAN module: https://content.arduino.cc/assets/mkrwan1310-murata_lora_module-type_abz.pdf

The datasheet says this module uses a SX1276 chip. Will this driver work with that?

ofauchon:
Hi @deadprogram,

SX126X and SX127X have differents SPI protocols.  So each one have its own driver.

I also wrote a driver for SX127x too : https://github.com/ofauchon/tinygo-drivers/tree/lora-sx127x/lora/sx127x

But at the moment, I'm giving focus on SX126x (and sx127x driver developpement is paused)

If you want to give a try, have a look to the  example : 

https://github.com/ofauchon/tinygo-drivers/blob/lora-sx127x/examples/lora/sx127x/pingpong.go

Or at least you could try to detect the SX127x radio module on the SPI with the following piece of code: 

https://github.com/ofauchon/tinygo-drivers/blob/515610326985ddc25c32b06c6c74655413c3cafa/lora/sx127x/sx127x.go#L126-L130

NB: sx126x is under active developement, and many improvements will have to be backported to SX127x: 
(Use DIO interrupts instead of IRQ Status register polling, Radio Event channnel refactoring ...) 

Feel free ton contact me on slack if you need more informations. 

Olivier

ofauchon:
Here is a first version of the SX126x driver. 
I could test it with Generic Node Sensor Edition and nucleo-wl55jc devices.

sx126x/sx126x.go
```diff
@@ -0,0 +1,690 @@
...
+	RadioEventCrcError  = iota
+	RadioEventUnhandled = iota
+
+	PERIOD_PER_SEC = (uint32)(1000000 / 15.625) // 1sec
+
+)
+
+// Device wraps an SPI connection to a SX127x device.
+type Device struct {
+	spi            machine.SPI     // SPI bus for module communication
```
deadprogram: Can you use `drivers.SPI` here?

sx126x/spi_stm32wl.go
```diff
@@ -0,0 +1,106 @@
...
+func (d *Device) SpiRx() uint8 {
+	r, _ := d.spi.Transfer(0xFF)
+	return r
+}
+
+//SpiSetNss Sets the NSS line
+func (d *Device) SpiSetNss(state bool) {
+	if d.spi != machine.SPI3 {
+		for {
+			// Only STM32WL's internal sx126x radio is yet implemented
```
deadprogram: Is going into an endless loop what you want here??

sx126x/spi_stm32wl.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+// WaitBusy sleep until all busy flags clears
+func (d *Device) WaitBusy() error {
+	count := 100
+	var rfbusyms, rfbusys bool
+	if d.spi != machine.SPI3 {
+		for {
+			// Only STM32WL's internal sx126x radio is yet implemented
```
deadprogram: Same comment about infinite loops as above.

sx126x/spi_stm32wl.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+	for count > 0 {
+		rfbusyms = stm32.PWR.SR2.HasBits(stm32.PWR_SR2_RFBUSYMS)
+		rfbusys = stm32.PWR.SR2.HasBits(stm32.PWR_SR2_RFBUSYS)
+
+		if !(rfbusyms && rfbusys) {
+			return nil
+		}
+		//time.Sleep(time.Millisecond)
```
deadprogram: Remove commented line?

sx126x/sx126x.go
```diff
@@ -0,0 +1,690 @@
...
+func (d *Device) ReadRegister(addr, size uint16) ([]uint8, error) {
+	d.CheckDeviceReady()
+	d.SpiSetNss(false)
+	// Send command
+	d.SpiTx(SX126X_CMD_READ_REGISTER)
+	d.SpiTx(uint8((addr & 0xFF00) >> 8))
+	d.SpiTx(uint8(addr & 0x00FF))
+	d.SpiTx(uint8(0x00)) // This byte is for status (unused yet)
+
+	var ret []uint8
```
deadprogram: You can define this as `var ret [size]unit8` to avoid heap allocations, and then instead of the loop use something like this:

```go
d.spi.Tx(nil, ret[:])
```

sx126x/sx126x.go
```diff
@@ -0,0 +1,690 @@
...
+	p[5] = uint8(dio2Mask & 0xFF)
+	p[6] = uint8((dio3Mask >> 8) & 0xFF)
+	p[7] = uint8(dio3Mask & 0xFF)
+	d.ExecSetCommand(SX126X_CMD_SET_DIO_IRQ_PARAMS, p[:])
+}
+
+// GetIrqStatus returns IRQ status
+func (d *Device) GetIrqStatus() (irqStatus uint16) {
+	r := d.ExecGetCommand(SX126X_CMD_GET_IRQ_STATUS, 2)
+	//println("IRQ STATUS ", r[0], r[1])
```
deadprogram: Remove comment?

sx126x/sx126x.go
```diff
@@ -0,0 +1,690 @@
...
+
+	// Switch to standby prior to configuration changes
+	d.SetStandby()
+
+	// Clear errors, disable radio interrupts for the moment
+	d.ClearDeviceErrors()
+	d.ClearIrqStatus(SX126X_IRQ_ALL)
+	d.SetDioIrqParams(0x00, 0x00, 0x00, 0x00)
+
+	// radio.CalibrateAll()
```
deadprogram: Remove commented?

deadprogram:
I made some comments, you can extrapolate to other parts of the code that use the same patterns.

Also, please note the merge conflict.

sx126x/spi_stm32wl.go
ofauchon: You are right, this is not very "clean"....

But you fall here in a rare and specific use case: 

You want to use an external SPI SX126x radio on a STM32WL SoC (which already have a SX1262 module on the same die) -> Want to implement a multi-receiver, or a radio bridge ?

Ok, I agree. 
The best solution would be to implement this, but I'm not sure I'll have the time to do that in the next weeks. 

At the time I wrote the code, I identified three ways to address the problem: 

- Silent failure: Not very kind for who will try to run the driver in this setup
- Raising an error: I found it overkill to add error handling in such kind of basic functions.
- Endless loop: At least, it will stop the execution, and it'll be easy to track down the origin with a debugger.

How would you have dealt with this situation ? 

Thanks 

Olivier

deadprogram: If this is something that can be detected at compile time then `panic` is probably the correct answer.

sx126x/sx126x.go
ofauchon:
```
var ret [size]uint8 fails with 
```

... fails with : 
```
array length size (variable of type uint16) must be constant
```
I could also declare it as a static array with some max size (eg: ret[MAX_READ_SIZE]uint8), but would that make sense ?

sx126x/spi_stm32wl.go
ofauchon: panic is the good answer... 

I added the check/panic  in the New constuctor.

sx126x/spi_stm32wl.go
ofauchon: I added a panic() in the constructor if SPI is not SPI3.

sx126x/sx126x.go
ofauchon: Done. 
Checked for other useless comments in the whole driver

sx126x/sx126x.go
ofauchon: ok

sx126x/sx126x.go
ofauchon: done

sx126x/sx126x.go
deadprogram: > I could also declare it as a static array with some max size (eg: ret[MAX_READ_SIZE]uint8), but would that make sense ?

Yes that would be better. The idea is to avoid all of the heap allocations by reusing an array as a slice hence avoiding use of `make` when not really needed.

ofauchon: Thanks @deadprogram 

Do you think I could use such commands:  
```
tinygo build -print-allocs=. -target gnse -o test.hex  2>&1 | grep heap  | grep sx126x
/home/olivier/go/src/tinygo.org/x/drivers/sx126x/sx126x.go:293:14: object allocated on the heap: escapes at line 294
/home/olivier/go/src/tinygo.org/x/drivers/sx126x/spi_stm32wl.go:20:22: object allocated on the heap: escapes at line 20
/home/olivier/go/src/tinygo.org/x/drivers/sx126x/spi_stm32wl.go:16:2: object allocated on the heap: escapes at line 26
```

... to improve my code ? 
Because it doesn't report heap allocation on the line you spotted earlier. 
```
253	var ret []uint8
```

Can this be an potential issue at runtime ?

Thanks.

deadprogram: That is a question best answered by @aykevl

sx126x/sx126x.go
```diff
@@ -0,0 +1,714 @@
...
+func (d *Device) ExecSetCommand(cmd uint8, buf []uint8) {
+	d.CheckDeviceReady()
+	if cmd == SX126X_CMD_SET_SLEEP {
+		d.deepSleep = true
+	} else {
+		d.deepSleep = false
+	}
+	d.SpiSetNss(false)
+	// Send command and params
+	d.SpiTx(cmd)
```
deadprogram: Why not use this

```go
d.spi.Tx(append(cmd, buf), nil)
```

instead of this block of lines?

ofauchon: Ok, 
I could removed the SpiTx 'wrapper' I was using. 
I could then use  s.spi.Tx and remove the loops.

deadprogram: I think this change looks a lot cleaner, thank you for making it!

sx126x/sx126x.go
ofauchon: At the end, I defined a spiBuffer[SPI_BUFFER_SIZE]  with SPI_BUFFER_SIZE=256 (in the device struct) . 
256 is the max allowed size for reading datas from the radio.

Please tell me if it's good or not  (My mind is not yet very clear on stack/heap usage) 

Thx

sx126x/sx126x.go
```diff
@@ -0,0 +1,700 @@
...
+	}
+}
+
+// ExecGetCommand queries the peripheral the peripheral
+func (d *Device) ExecGetCommand(cmd uint8, size uint8) []uint8 {
+	d.CheckDeviceReady()
+	d.SpiSetNss(false)
+	// Send the command and flush first status byte (as not used)
+	d.spi.Tx([]uint8{cmd, 0x00}, nil)
+	ret := make([]uint8, size)
```
deadprogram: Could this remove the `ret := make([]uint8, size)` and instead be just:

```go
d.spi.Tx(nil, d.spiBuffer[:size])
```

At the end of the function you would want to `return d.spiBuffer[:size]`.

ofauchon: Oh, I forgot this one... :-(
I fixed it, the way you suggests.
No more make() in the driver. 

I made some tests with nucleo-wl55jc, and it works properly.

deadprogram:
Thanks for making all the changes @ofauchon and for such a powerful contribution! Now merging.

PR #346, opened 2021-12-07
wifinina: avoid busy wait

ysoldak:
Use `time.Sleep()` instead of busy-wait in wifinina driver.

Feels like an improvement in context of cooperative runtime: we now let other goroutines work while waiting for response from NINA chip.

Busy-wait approach was supposedly coded in times of unreliable `time.Sleep()` implementation.


deadprogram:
Thank you @ysoldak now merging.

PR #350, opened 2021-12-10
XPT2046 Touch driver

spearson78:


sago35:
Thank you for your contribution.
Please add Examples.

sago35:
I would like you to implement the following interface.

https://github.com/tinygo-org/drivers/blob/90a13a697bc51b41b878ffc9b0fb5ff201ef0731/touch/point.go#L4-L6

spearson78:
I've added an example and migrated to the ReadTouchPoint interface.

I removed the internal scaling to display size to be consistent with the fourwire implementation however I currently return 12 bit values. Should I scale them to 16 bit to match fourwire?

xpt2046/xpt2046.go
```diff
@@ -0,0 +1,175 @@
+// Package xpt2046 implements a driver for the XPT2046 resistive touch controller as packaged on the TFT_320QVT board
+//
+// Datasheet: http://grobotronics.com/images/datasheets/xpt2046-datasheed.pdf
```
sago35:
```suggestion
// Datasheet: http://grobotronics.com/images/datasheets/xpt2046-datasheet.pdf
```

xpt2046/xpt2046.go
```diff
@@ -0,0 +1,175 @@
...
+func (d *Device) ReadTouchPoint() touch.Point {
+
+	tx := uint32(0)
+	ty := uint32(0)
+	tz := uint32(0)
+	sampleCount := uint8(0)
+
+	d.t_cs.Low()
+
+	for ; sampleCount < d.precision && d.Touched(); sampleCount++ {
```
sago35, line 111: If d.precision is 0, `readRaw()` will not be executed.
The initial value of d.precision should be a value greater than or equal to 1.

xpt2046/xpt2046.go
```diff
@@ -0,0 +1,175 @@
...
+	t_clk  machine.Pin
+	t_cs   machine.Pin
+	t_din  machine.Pin
+	t_dout machine.Pin
+	t_irq  machine.Pin
+
+	precision uint8
+}
+
+func New(t_clk, t_cs, t_din, t_dout, t_irq machine.Pin, precision uint8) Device {
```
sago35: I think it would be better to remove the precision setting from New.
It would be better to use Configure(config *Config) instead, where precision can be set.

xpt2046/xpt2046.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if sampleCount > 0 {
+		x := int(tx / uint32(sampleCount))
+		y := int(ty / uint32(sampleCount))
+		z := int(tz / uint32(sampleCount))
+		return touch.Point{
+			X: x,
+			Y: y,
+			Z: z,
+		}
```
sago35, line 127: I think it's not so good if each driver returns different values, so I think it's better to scale it to 16bit as you commented.

I'm also working on a touch-driver called FT6636.
The FT6636 also returns a 12-bit value from the device, but I' m going to scale it.

sago35 (commented):
Thanks for the update.
For the most part I think it is very nice.

Please add it to README.md.
Also, please add smoketest to the Makefile.

xpt2046/xpt2046.go
spearson78: Sorry this was a search replace error.

xpt2046/xpt2046.go
spearson78: I set a default of 10 if 0 is passed.

xpt2046/xpt2046.go
spearson78: Done

xpt2046/xpt2046.go
spearson78: I added the scaling and updated the example.

spearson78:
I've added a default value of 10 for precision if the user sets it to 0 and I now scale the values to 16 bits
I moved the precision to a new Config struct. Should I move the entire pin config to the Configure method?
I updated the Readme and make smoke-test targets.

I'd like to say thank you for guiding me through the requirements for contributions to this project. This is my first commit to tinygo and its been very enjoyable so far :)

sago35:
It seems that there are few cases where the config struct contains pin information.
I think the current source code is fine.

sago35 (approved):
Looks good to me.

I'm a little concerned because I think `tz` is probably 24-bit-range.
However, this is probably not a problem for z.

sago35:
Thank you for your contribution.
Now marging.

PR #352, opened 2021-12-18
Fix LSM303AGR example

alankrantas:
Minor fix of the example code to fix convention, simplify the output and provide some more configure information.


deadprogram:
Thanks @alankrantas now merging.

PR #354, opened 2021-12-24
Fix hd44780i2c's new line

alankrantas:
Found this little bug while try to build a mini game on it. When the cursor is at the last column of any row, the cursor would break to new line _first_. This should be done after printing the character.

Also added some extra descriptions info for the ```SetCursor``` method.


deadprogram:
Thanks for the fix @alankrantas now merging.

PR #357, opened 2022-01-12
wifi/espat, rtl8720dn, wifinina: use common net.Adapter interface

deadprogram:
This PR moves towards implementing a standard minimum common interface for the supported WiFi adaptors:

- espat
- rtl8720dn
- wifinina


deadprogram:
The idea is to consolidate as many of the examples for the WIFi Drivers as possible so they can share code. As part of that process we can identify any differences/gaps.

examples/wifinina/http-get/main.go
```diff
@@ -132,22 +132,18 @@ func waitSerial() {
 
 // connect to access point
 func connectToAP() {
-	if len(ssid) == 0 || len(pass) == 0 {
+	time.Sleep(2 * time.Second)
+	println("Connecting to " + ssid)
```
sago35, line 136: It may be better to use message().
The original newline code was CRLF.

examples/wifinina/http-get/main.go
```diff
@@ -132,22 +132,18 @@ func waitSerial() {
...
-	time.Sleep(2 * time.Second)
-	message("Connecting to " + ssid)
-	adaptor.SetPassphrase(ssid, pass)
-	for st, _ := adaptor.GetConnectionStatus(); st != wifinina.StatusConnected; {
-		message("Connection status: " + st.String())
-		time.Sleep(1 * time.Second)
-		st, _ = adaptor.GetConnectionStatus()
-	}
-	message("Connected.")
-	time.Sleep(2 * time.Second)
```
sago35, line 150: This `time.Sleep(2 * time.Second)` is missing.
However, it is probably not a problem.

sago35 (commented):
@deadprogram 
I think it's very good.
As I commented, I think it would be better to have a unified idea whether `CRLF` or `LF` should be used.

wifi.go
```diff
@@ -0,0 +1,16 @@
...
+
+var (
+	ErrWiFiMissingSSID    = errors.New("missing SSID")
+	ErrWiFiConnectTimeout = errors.New("WiFi connect timeout")
+)
+
+type WiFiDriver interface {
+	ConnectToAccessPoint(ssid, pass string, timeout time.Duration) error
+	Disconnect() error
+}
```
sago35: It seems that the following can also be included in the interface.
At the moment, espat has a slightly different interface, but I think it can be unified.

```
GetIP() (ip, subnet, gateway IPAddress, err error)
```

https://github.com/tinygo-org/drivers/blob/90a13a697bc51b41b878ffc9b0fb5ff201ef0731/wifinina/wifinina.go#L513-L521
https://github.com/tinygo-org/drivers/blob/90a13a697bc51b41b878ffc9b0fb5ff201ef0731/espat/wifi.go#L61-L65
https://github.com/tinygo-org/drivers/blob/90a13a697bc51b41b878ffc9b0fb5ff201ef0731/rtl8720dn/wifi.go#L60-L72

deadprogram:
@sago35 I made some significant changes, please take a look. The biggest one is that I have redefined `net.DeviceDriver` as `net.Adapter` and simplified the usage from the other Drivers to be more typically idiomatic for Go interfaces.

There is a lot more that can be done, such as following up on your suggestion to unify the usage of `IPAddress`. However perhaps this is already enough for this first PR and we can work on it further in followup PRs?

sago35 (approved):
I think it's very good.

sago35:
@deadprogram It looks like we will be able to code the same way with espat, wifinina and rtl8720dn.
I think it's a very good way to proceed.

deadprogram:
I've added one more commit to this PR with a unified tcpclient example that can compile on all supported wifi adaptors:

```
$ tinygo build -size short -o ./build/test.hex -target=wioterminal -tags=rtl8720dn ./examples/wifi/tcpclient/                                                                
  56624    1448    8672 |   58072   10120                                                                                                                                    

$ tinygo build -size short -o ./build/test.hex -target=pyportal -tags=wifinina ./examples/wifi/tcpclient/                                                                    
  14828    1376    6360 |   16204    7736                                                                                                                                    

$ tinygo build -size short -o ./build/test.hex -target=arduino-nano33 -tags=espat ./examples/wifi/tcpclient/                                                                 
  14128    1392    4448 |   15520    5840
```

deadprogram:
Thanks for helping review @sago35 now merging.

PR #365, opened 2022-01-24
wifinina: correct sendParamStr to handle empty strings

### LINKED ISSUE #364: wifinina: panic when trying to connect to AP without password
When trying to use the WiFiNINA driver, if you attempt to connect to an unsecured access point, you get the following error:

```
panic: runtime error: index out of range
```

It appears this is because the firmware expects an empty string param to have length of zero, and then not actually send any data for the null string. I will submit a PR with a fix that appears to address this shortly.

deadprogram:
This PR fixes #364 by modifing the WiFiNINA driver function `sendParamStr()` to correctly handle empty strings, such as when connecting to an unsecured access point.


sago35 (approved):
Looks good to me.

deadprogram:
Oops, forgot to include this in the latest release. Now merging.

PR #367, opened 2022-01-27
ws2812: add support for m5stamp-c3

sago35:
In this PR, add support for m5stamp-c3.
Implemented the case where machine.LED does not exist in others_wo_led.go, as in m5stamp-c3 and qtpy.

Tested with m5stamp-c3 and qtpy.


deadprogram:
Thank you @sago35 :traffic_light::traffic_light::traffic_light::traffic_light:

PR #368, opened 2022-01-29
lsmXXX (IMU units): Error handling, etc

ysoldak:
This PR 
- unifies implementations of lsmXXX family of IMU sensors,
- adds error handling for all methods to all drivers,
- implements better memory management (fixed size array, sliced on demand),
- makes all New() methods return pointers
- adjusts examples

### DIFF
--- a/examples/lsm303agr/main.go
+++ b/examples/lsm303agr/main.go
@@ -13,7 +13,13 @@ func main() {
 	machine.I2C0.Configure(machine.I2CConfig{})
 
 	sensor := lsm303agr.New(machine.I2C0)
-	sensor.Configure(lsm303agr.Configuration{}) //default settings
+	err := sensor.Configure(lsm303agr.Configuration{}) //default settings
+	if err != nil {
+		for {
+			println("Failed to configure", err.Error())
+			time.Sleep(time.Second)
+		}
+	}
 
 	// you can specify the following options to adjust accuracy, sensor range or save power.
 	// see https://github.com/tinygo-org/drivers/blob/release/lsm303agr/registers.go for details:
@@ -28,22 +34,24 @@ func main() {
 		})
 	*/
 
-	if !sensor.Connected() {
-		println("LSM303AGR/MAG not connected!")
-		return
-	}
-
 	for {
+
+		if !sensor.Connected() {
+			println("LSM303AGR/MAG not connected!")
+			time.Sleep(time.Second)
+			continue
+		}
+
 		// accel_x, accel_y, accel_z := sensor.ReadAcceleration()
 		// println("ACCEL_X:", accel_x/100000, " ACCEL_Y:", accel_y/100000, " ACCEL_Z:", accel_z/100000)
 
 		// mag_x, mag_y, mag_z := sensor.ReadMagneticField()
 		// println("MAG_X:", mag_x/100000, " MAG_Y:", mag_y/100000, " MAG_Z:", mag_z/100000)
 
-		pitch, roll := sensor.ReadPitchRoll()
+		pitch, roll, _ := sensor.ReadPitchRoll()
 		println("Pitch:", float32(pitch)/100000, " Roll:", float32(roll)/100000)
 
-		heading := sensor.ReadCompass()
+		heading, _ := sensor.ReadCompass()
 		println("Heading:", float32(heading)/100000, "degrees")
 
 		temp, _ := sensor.ReadTemperature()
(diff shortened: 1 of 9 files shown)


lsm303agr/lsm303agr.go
```diff
@@ -34,25 +36,45 @@ type Configuration struct {
...
+		bus:          bus,
+		AccelAddress: ACCEL_ADDRESS,
+		MagAddress:   MAG_ADDRESS,
+	}
 }
 
 // Connected returns whether both sensor on LSM303AGR has been found.
 // It does two "who am I" requests and checks the responses.
-func (d *Device) Connected() bool {
+func (d *Device) Connected() (connected bool, err error) {
```
aykevl: Not sure about this change. There are a lot of drivers that have the `Connected()` method, and they all only return a `bool`. And maybe that's okay: if there is an I2C error, the device is probably not connected.
Anybody else have an opinion in this?

If we're going to change the signature of the `Connected()` method, I think we should change it for all drivers at once for consistency.

(On the other hand, it would be nice to return an I2C error in `Configure()` below if the connection fails).

lsm303agr/lsm303agr.go
```diff
@@ -90,36 +112,46 @@ func (d *Device) Configure(cfg Configuration) {
...
-func (d *Device) ReadAcceleration() (x int32, y int32, z int32) {
-
-	data1, data2, data3, data4, data5, data6 := []byte{0}, []byte{0}, []byte{0}, []byte{0}, []byte{0}, []byte{0}
-	d.bus.ReadRegister(uint8(d.AccelAddress), ACCEL_OUT_X_H_A, data1)
-	d.bus.ReadRegister(uint8(d.AccelAddress), ACCEL_OUT_X_L_A, data2)
-	d.bus.ReadRegister(uint8(d.AccelAddress), ACCEL_OUT_Y_H_A, data3)
-	d.bus.ReadRegister(uint8(d.AccelAddress), ACCEL_OUT_Y_L_A, data4)
-	d.bus.ReadRegister(uint8(d.AccelAddress), ACCEL_OUT_Z_H_A, data5)
-	d.bus.ReadRegister(uint8(d.AccelAddress), ACCEL_OUT_Z_L_A, data6)
+func (d *Device) ReadAcceleration() (x, y, z int32, err error) {
```
aykevl, line 139: Matches other drivers, so :+1: from me.

lsm303agr/lsm303agr.go
```diff
@@ -90,36 +112,46 @@ func (d *Device) Configure(cfg Configuration) {
...
+	data := d.buf[:6]
+	err = d.bus.ReadRegister(uint8(d.AccelAddress), ACCEL_OUT_X_L_A, data)
```
aykevl, line 141: :+1:, this should be much faster.

aykevl (commented):
Generally looks good to me, but see below.

lsm303agr/lsm303agr.go
ysoldak: I see. The reasoning behind two return values is it lets user distinguish between a bus error and sensor not found. (In addition to being consistent with other methods, that have error too.)

Alternative would be to have one return value, a boolean, and return false on bus error?
I'm fine with that, shall I change?

ysoldak: @aykevl @deadprogram shall I remove "error" return value from `Connected()` method to help this PR?
And then we can have a follow-up to update `Connected()` signature (to include error) across all drivers?

deadprogram: Sounds like a good plan.

examples/lsm303agr/main.go
```diff
@@ -28,22 +34,24 @@ func main() {
...
-
 	for {
+
+		if con, err := sensor.Connected(); !con || err != nil {
+			println("LSM303AGR/MAG not connected!")
+			time.Sleep(time.Second)
+			continue
+		}
+
 		// accel_x, accel_y, accel_z := sensor.ReadAcceleration()
```
deadprogram, line 45: Are these supposed to be commented out here?

ysoldak: Have no idea, not my code, I've just made it work with refactorings.
I suppose, since it's an example, this shows how to read raw acceleration, instead of Pitch and Roll.

lsm303agr/lsm303agr.go
ysoldak: Error was removed from all Connect() methods

ysoldak:
I've removed error return from Connected() methods, PTAL. @deadprogram @aykevl

deadprogram:
Thanks for working on this @ysoldak and also for all the revisions. Now merging.

PR #369, opened 2022-01-29
LPS22HB Pin rename

ysoldak:
Syncs with main tinygo repo, see https://github.com/tinygo-org/tinygo/pull/2579


deadprogram:
Thanks for the quick fix @ysoldak now merging.

PR #370, opened 2022-01-30
drivers: add driver for IS31FL3731 matrix LED driver

antonfisher:
This PR adds a driver for IS31FL3731 matrix LED driver.
Datasheet: https://www.lumissil.com/assets/pdf/core/IS31FL3731_DS.pdf
The first version is basic, but functional.

**What's implemented:**
- picture mode ("auto frame play mode" and "audio frame play mode" are not supported in this version of the driver)
- drawing by XY coordinates or raw pixel's index on LED layout
- frames (0-7) and switching between them.

**Supported boards:**
- any custom LED matrix layout built on IS31FL3731 -- using `DrawPixelIndex(...)` function
- [Adafruit 15x7 CharliePlex LED Matrix FeatherWing (CharlieWing)](https://www.adafruit.com/product/3163) -- using `DrawPixelXY(...)` or `DrawPixelIndex(...)` function (I've tested on this one with NRF52840 controller).

**Boards can be added in the future versions:**
- [Adafruit 16x9 Charlieplexed PWM LED Matrix Driver - IS31FL3731](https://www.adafruit.com/product/2946) -- I've ordered it, waiting to test.

**Tests:**
- `fmt-check` pass
- `unit-test` pass
- `smoke-test` pass for the driver:
image
Except the AVR part fails because I do not have `avr-gcc` installed.

**Example code works like this:**
led-driver-0
led-driver-1


is31fl3731/is31fl3731.go
```diff
@@ -0,0 +1,303 @@
...
+
+// New creates a raw driver w/o any preset board layout
+func New(bus drivers.I2C, address uint8) (d Device, err error) {
+	d = Device{
+		Address: address,
+		bus:     bus,
+		board:   boardRaw,
+	}
+
+	err = d.configure()
```
ysoldak: In other drivers, we don't touch devices in `New()` methods. `Configure()` method must be called separately, usually with configuration parameters.
This means `configure()` must be changed to public (->`Configure()`)

is31fl3731/is31fl3731.go
```diff
@@ -0,0 +1,303 @@
...
+
+	return d, err
+}
+
+// NewAdafruitCharlieWing15x7 creates a new driver with Adafruit 15x7
+// CharliePlex LED Matrix FeatherWing (CharlieWing) layout
+func NewAdafruitCharlieWing15x7(bus drivers.I2C, address uint8) (d Device, err error) {
```
ysoldak: This seem to be very specific, I'd extract all code related to CharlieWing to its own file, like `is31fl3731_acw15x7.go`

examples/is31fl3731/main.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"time"
+
+	"machine"
+
+	"tinygo.org/x/drivers/is31fl3731"
+)
+
+// I2CAddress -- address of led matrix
+var I2CAddress uint8 = 0x74
```
ysoldak: How random is this address? Can it be moved to the driver itself?

antonfisher: For standalone is31fl3731 the addresses can be:
- 0x74 (AD pin connected to GND)
- 0x75 (AD pin connected to SCL)
- 0x76 (AD pin connected to SDA)
- 0x77 (AD pin connected to VCC)

For Adafruit Charlie Wing 15x7 its either 0x74 or 0x77. I guess I can make four constants, but it's up to the particular board to define address. WDYT?

is31fl3731/is31fl3731.go
antonfisher: Changed to `Configure()` 👍

is31fl3731/is31fl3731.go
antonfisher: Moved all CharlieWing code to a separate file (struct) 👍

examples/is31fl3731/main.go
ysoldak: I'd say have them 4 constants in the driver "registers.go" as per praxis and seen in other drivers.
Preferably with comments (exactly as you did here in comment).
And let user of the driver choose which address to use.

ysoldak: Disclaimer, I'm not a maintainer, but submitted several drivers already to this repo and sort of know the drill.
I like your PR and just want to help it :)

antonfisher: got it, thanks for the help! :)

antonfisher: Added address constants 👍

deadprogram:
Thank you @antonfisher for working on this, and thanks @ysoldak for proving all of the feedback/reviews.

I am going to squash/merge now.

PR #372, opened 2022-02-12
buzzer: add pwm version buzzer

sago35:
In this PR, the PWM version buzzer will be added.

The difference in sound when using Wio Terminal is as follows.
https://twitter.com/sago35tk/status/1488848571533500416?s=20&t=aN3Y_Qsd0aHKMxuwVQn-Vg


sago35:
Make it depend on PWM interface instead of machine.TCC.
The machine.TCC is probably only present in samd.

deadprogram:
Small change, perhaps better to name the example here `buzzer-pwm` instead of `buzzer2`? Seems more meaningful.

sago35:
@deadprogram 
You are right.
If the review is OK, squash merge when you merge.
Please let me know if you have any other remarks.

aykevl:
We already have something very similar to this, the tone driver. How is this different?

sago35:
Oops, I didn't know there was such a thing as a tone driver.

sago35:
I modified the source code so that CI would succeed.
However, this PR is unnecessary and can be closed.

PR #379, opened 2022-02-19
stm32: add new device package

deadprogram:
This PR switches all references to `device/stm32` over to the `tinygo.org/x/device` package repo location.


deadprogram:
Got past the original error, but now hit a different one:

```
tinygo build -size short -o ./build/test.hex -target=nucleo-wl55jc ./examples/sx126x/lora_rxtx/
error: Linking globals named 'WWDG_IRQHandler': symbol multiply defined!
```

I am able to reproduce this on my local machine as well with the latest `dev` build.

@aykevl can you please help out here?

sago35:
I think export has been executed multiple times.
Is there something that needs to be changed in tinygo?

https://github.com/tinygo-org/device/blob/main/stm32/stm32wle5.go
line 197

```go
//export WWDG_IRQHandler
func interruptWWDG() {
	callHandlers(IRQ_WWDG)
}
```

aykevl:
@deadprogram my guess is that you still have a `device/stm32` import somewhere. You could remove src/device/stm32 to check for this.

deadprogram:
I had removed that directory from my local TinyGo (`TINYGOROOT`) install even before that error.

aykevl:
Huh. Weird. It is defined in two packages: `tinygo.org/x/device/stm32` and `vendor/tinygo.org/x/device/stm32`. That's not something that should happen.

aykevl:
This might be intended behavior of the Go toolchain (`go list`). In which case I would really need to rethink how to implement this feature.

PR #381, opened 2022-02-19
dht: fix error check in example

Elfo404:
Follow up to https://github.com/tinygo-org/drivers/pull/328#issuecomment-1046024264

Fixes an error check in dht examples that prevents readings from being printed when there's no error.

cc. @doniacld


deadprogram:
Thanks @Elfo404 for the correction! Now merging.

Issue #363, opened 2022-01-20
image/jpeg and image/png tests fail to build?


dkegel-fastly:
I just cloned the repo and tried
```
cd image/jpeg
tinygo test
```

This failed with 
```
reader_test.go:92:2: wrong number of return values (want 2, got 1)
```

It looks like an internal disagreement about the return values of Decode()... @sago35, could you comment?

sago35:
@dkegel-fastly 
I changed the interface from Go's image package.

https://github.com/tinygo-org/drivers/tree/release/examples/ili9341/slideshow
https://github.com/tinygo-org/drivers/blob/90a13a697bc51b41b878ffc9b0fb5ff201ef0731/examples/ili9341/slideshow/main.go#L66-L76

However, it failed because I did not change the tests.

I think we can either maintain the test or remove it.

dkegel-fastly:
I'd vote for fixing the test and turning it on in CI, on the theory that it helps prevent bitrot.

deadprogram:
> I changed the interface from Go's image package.

Is that strictly needed? I have not looked at the specifics in this case, just asking.

sago35:
> Is that strictly needed? I have not looked at the specifics in this case, just asking.

Due to memory size limitations, it was not possible to return `image.Image`.
I changed the function signature to make the user aware of the mistake, but in hindsight I think it would have been better to just return nil.

* Go's image/png : `func Decode(r io.Reader) (image.Image, error)`
* tinygo.org/x/image/png : `func Decode(r io.Reader) error`


> I'd vote for fixing the test and turning it on in CI, on the theory that it helps prevent bitrot.

I'll take a look.

PR #382, opened 2022-02-20
image: fix interface

sago35:
This PR fixes the interface that was discussed in #363.

The test is not OK yet, but it is the same as the Go standard test result, so we won't deal with it here.

```
sago3@B450GT3 MINGW64 ~/dev/src/tinygo.org/x/drivers (fix-interface)
$ cd image/png/ && tinygo test
FAIL    tinygo.org/x/drivers/image/png  0.156s
FAIL

sago3@B450GT3 MINGW64 ~/dev/src/tinygo.org/x/drivers/image/png (fix-interface)
$ cd ../jpeg/ && tinygo test
ld.lld: error: undefined symbol: time.startTimer
>>> referenced by C:\Go\src\time\sleep.go:96
>>>               C:\Users\sago3\AppData\Local\Temp\tinygo2750643149\main.o:(tinygo.org/x/drivers/image/jpeg.TestLargeImageWithShortData)
failed to run tool: ld.lld
error: failed to link C:\Users\sago3\AppData\Local\Temp\tinygo2750643149\main.exe: exit status 1
FAIL
```

### DIFF
--- a/examples/ili9341/slideshow/main.go
+++ b/examples/ili9341/slideshow/main.go
@@ -72,7 +72,8 @@ func drawPng(display *ili9341.Device) error {
 		}
 	})
 
-	return png.Decode(p)
+	_, err := png.Decode(p)
+	return err
 }
 
 func drawJpeg(display *ili9341.Device) error {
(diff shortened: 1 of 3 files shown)


deadprogram:
Thanks for the fix here @sago35 now merging.

PR #383, opened 2022-02-21
Add 'irremote' package as a basic infra-red receiver driver (#384)

### LINKED ISSUE #384: Add an infra-red receiver driver
Add a driver to support IR receivers like AX-1838HS

I opened PR #383 as a starting point. It supports only NEC extended protocol so far.

neildavis:
Supports only NEC (extended address) protocol including repeat codes so far

(Partially) fixes #384

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -229,13 +229,15 @@ endif
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.uf2 -target=pico ./examples/ssd1289/main.go
 	@md5sum ./build/test.uf2
+	tinygo build -size short -o ./build/test.hex -target=pico ./examples/irremote/main.go
+	@md5sum ./build/test.hex
 
 DRIVERS = $(wildcard */)
 NOTESTS = build examples flash semihosting pcd8544 shiftregister st7789 microphone mcp3008 gps microbitmatrix \
 		hcsr04 ssd1331 ws2812 thermistor apa102 easystepper ssd1351 ili9341 wifinina shifter hub75 \
 		hd44780 buzzer ssd1306 espat l9110x st7735 bmi160 l293x keypad4x4 max72xx p1am tone tm1637 \
 		pcf8563 mcp2515 servo sdcard rtl8720dn image cmd i2csoft hts221 lps22hb apds9960 axp192 xpt2046 \
-		ft6336 sx126x ssd1289
+		ft6336 sx126x ssd1289 irremote
 TESTS = $(filter-out $(addsuffix /%,$(NOTESTS)),$(DRIVERS))
 
 unit-test:
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ The following 78 devices are supported.
 | [software I2C driver](https://www.ti.com/lit/an/slva704/slva704.pdf) | GPIO |
 | [ILI9341 TFT color display](https://cdn-shop.adafruit.com/datasheets/ILI9341.pdf) | SPI |
 | [INA260 Volt/Amp/Power meter](https://www.ti.com/lit/ds/symlink/ina260.pdf) | I2C |
+| [Infrared remote control](https://en.wikipedia.org/wiki/Consumer_IR) | GPIO |
 | [IS31FL3731 matrix LED driver](https://www.lumissil.com/assets/pdf/core/IS31FL3731_DS.pdf) | I2C |
 | [4x4 Membrane Keypad](https://cdn.sparkfun.com/assets/f/f/a/5/0/DS-16038.pdf) | GPIO |
 | [L293x motor driver](https://www.ti.com/lit/ds/symlink/l293d.pdf) | GPIO/PWM |
--- a/examples/irremote/main.go
+++ b/examples/irremote/main.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/irremote"
+)
+
+var irCmdButtons = map[uint16]string{
+	0xA2: "POWER",
+	0xE2: "FUNC/STOP",
+	0x62: "VOL+",
+	0x22: "FAST BACK",
+	0x02: "PAUSE",
+	0xC2: "FAST FORWARD",
+	0xE0: "DOWN",
+	0xA8: "VOL-",
+	0x90: "UP",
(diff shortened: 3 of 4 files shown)


sago35:
Why isn't CircleCI working?

neildavis:
> Why isn't CircleCI working?

I don't know. I was hoping someone here could help with that? I also asked on Slack in `#tinygo-dev` but did not get an answer there.

Is there something I need to do to get CCI to run? I set it up on my fork but that did not seem to make any difference. I assumed maybe it only runs for `tinygo-org` members?

If there is action required by contributors to make this work it would be helpful to add that to the `CONTRIBUTING` doc

deadprogram:
I do not know why it is not running. Some new feature of Github not quite working, perhaps?

neildavis:
> I do not know why it is not running. Some new feature of Github not quite working, perhaps?

@deadprogram Could you try toggling the setting in CircleCi please?
[Project Settings -> Advanced -> Build forked pull requests](https://app.circleci.com/settings/project/github/tinygo-org/drivers/advanced)
(not sure if the link will work, I can't see it due to permissions but crafted it from my fork & replacing the org name)
Assuming it's 'On' already, maybe toggling it off/on will help?
Thanks

deadprogram:
It was working previously. Not sure why it stopped, probably related to the GH protection settings that are now automatic and required. CircleCI does not have that setting. So, sounds like a support request to either GH or CircleCI or both.

irremote/irremote.go
```diff
@@ -0,0 +1,173 @@
...
+// https://techdocs.altium.com/display/FPGA/NEC+Infrared+Transmission+Protocol
+// https://simple-circuit.com/arduino-nec-remote-control-decoder/
+
+// Client callback function for IR commands
+// 'code' is the raw NEC format code. 'addr' & 'cmd' are the NEC decoded address and command codes respectively
+// See the NEC protocol reference links above for more information
+type IRCallback func(code uint32, addr uint16, cmd uint8, repeat bool)
+
+// Public API for an IR receiver device
+type IRReceiverDevice interface {
```
deadprogram: Is there another device that implements this interface? Just wondering if it is better to wait until there is another before using interface here?

irremote/irremote.go
```diff
@@ -0,0 +1,173 @@
...
+	// Configure pins for the IR receiver module
+	Configure()
+	// Start/Stop receiving IR callbacks (pass nil to stop)
+	Callback(cb IRCallback)
```
deadprogram: Perhaps better to name this function something like `HandleDataComplete()`?

irremote/irremote.go
neildavis: Yes, you're probably right. My thinking was that extending to cover other protocols (e.g. Sony SIRC, Phillips RC5/RC6 etc.) may result in additional devices in the future, but that's probably premature for now, especially on a pre-1.0 release. 

On a similar note I considered renaming the `New()` factory/constructor to `NewNEC()` for the same reason, thinking maybe it should be explicit and also reserving the generic `New()` for a potential 'automatic' protocol sensing device in the future? Again, I guess that's premature, although it does raise the question of whether we should strive to maintain backwards source compatibility in iterative versions prior to a 1.0 release? My opinion would be 'no' but happy to go with the consensus here. Any guidance?

irremote/irremote.go
neildavis: I agree that this is badly named and needs changing, although I'm not too keen on `HandleDataComplete()` (sorry!)  Personally I'd prefer to avoid the terminology of 'complete' in this case. My justification for this is that multiple IR commands/codes occur in discrete packets over time with no real concept of 'completing' or 'terminating' until you set the callback to `nil` to tell the driver you're not interested in them anymore. i.e. it's not a continous stream that has a natural completion. Neither is it a 'one-shot' async call that 'completes' Not sure I'm describring that too well but hopefully you can understand my thoughts?

It's a 'setter' function - you are setting a callback function property, passing `nil` to clear it. It's analagous to `machine.SetInterrupt()` and actually called from within an interrupt handler - so IMHO it should be named something of the form `SetXxxYyy()` How about  `SetCommandHandler()`, `SetCommandCallback()`,  or just plain `SetCallback()`??

On a related note I was questioning the use of using callback functions from an interrupt handler. I'm not sure if this is a good idea? However, the only alternative I can think of is to use a channel to 'transpose' the callback back to the client thread/goroutine. But use of channels would enforce the use of a scheduler which isn't great for the more constrained devices like AVR, so probably better to leave that up to the client? Any insight/advice here would be gratefully received! Thanks!

neildavis:
@deadprogram I refactored a little in light of your comments to remove the `interface` and also tidy up the API to closer reflect the commonly used [Arduino library](https://github.com/Arduino-IRremote/Arduino-IRremote) - mainly using a `struct` for the callback data rather than a long parameter list. 

I think this is ready for [re]review now, so please take a look when you get a chance. Thanks!

neildavis:
@deadprogram I refactored a little in light of your comments to remove the `interface` and also tidy up the API to closer reflect the commonly used [Arduino library](https://github.com/Arduino-IRremote/Arduino-IRremote) - mainly using a `struct` for the callback data rather than a long parameter list. 

I think this is ready for [re]review now, so please take a look when you get a chance. Thanks!

irremote/irremote.go
deadprogram: > whether we should strive to maintain backwards source compatibility in iterative versions prior to a 1.0 release

yes, unless we have a good reason to change. In which case better to change everything in a single release if possible.

irremote/irremote.go
deadprogram: This is usually beyond the scope of what most drivers should do. However interrupt handlers can and should use channels. "It depends" is always a sure answer.

Anyhow the word "Handler" is more commonly used in Go such as `net/http`. I like `SetCommandHandler()` the most here. :smile_cat:

neildavis:
@deadprogram I changed the callback terminology to `SetCommandHandler()` as discussed above. Repeat codes are also now implemented as a small addition to the protocol, and I added a line to `README.md` for the device. I'm hoping this is ready to merge now? Please take a look when you get a chance. Thanks :)

deadprogram:
Let's merge now so any future work can be in smaller more atomic increments.

Thank you very much for working on this @neildavis

Issue #358, opened 2022-01-14
DHT22 measure wrong values


fiurthorn:
using DHT22 from https://www.amazon.de/gp/product/B08NP8NR1J/ref=ppx_yo_dt_b_asin_title_o01_s03?ie=UTF8&psc=1.
gives wrong values:

- about 4808% humidity
- -1433°C

```
humidity := dht.NewWithPolicy(machine.GP15, dht.DHT22, dht.UpdatePolicy{})
t2, h, _ := humidity.Measurements()
temp2 = fmt.Sprintf("%d", t2/10)
hu = fmt.Sprintf("%d", h/10)
tinyfont.WriteLine(&display, &freeserif.Regular9pt7b, 40, 15, hu+" %", white)
tinyfont.WriteLine(&display, &freeserif.Regular9pt7b, 40, 30, temp2+"  C", white)
```

conejoninja:
Could we close this issue? seems to be fixed by #390

PR #390, opened 2022-02-26
dht: fix humidity and temperature extraction for DHT22 (#358)

doniacld:
## Context
When reading values from the DHT22 sensor with an ArduinoNano IoT33, the obtained values are not realistic.
e.g.:
```bash
Temperature:  537.6°C, Humidity: 4992.1%
```

## Resolution

Closes issue #358 

The buffer of the temperature value was padded using `LittleEndian` instead of `BigEndian`.
The 2 bytes buffers of the humidity value were just reverted.

Output
```
Temperature: 22.9°C, Humidity: 49.2%
```

NB: To make it works I had to disable the interruption in `receiveSignals`.
```go
state := interrupt.Disable()
defer interrupt.Restore(state)
```

## Tests

I wrote a simple unit test on the `extractData` method using the expected values from the [datasheet of the DHT22]( 
https://cdn-shop.adafruit.com/datasheets/Digital+humidity+and+temperature+sensor+AM2302.pdf).
I am not sure about the UT convention, any feedback is welcome.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,12 +231,12 @@ endif
 DRIVERS = $(wildcard */)
 NOTESTS = build examples flash semihosting pcd8544 shiftregister st7789 microphone mcp3008 gps microbitmatrix \
 		hcsr04 ssd1331 ws2812 thermistor apa102 easystepper ssd1351 ili9341 wifinina shifter hub75 \
-		hd44780 buzzer ssd1306 espat l9110x st7735 bmi160 l293x dht keypad4x4 max72xx p1am tone tm1637 \
+		hd44780 buzzer ssd1306 espat l9110x st7735 bmi160 l293x keypad4x4 max72xx p1am tone tm1637 \
 		pcf8563 mcp2515 servo sdcard rtl8720dn image cmd i2csoft hts221 lps22hb apds9960 axp192 xpt2046 \
 		ft6336 sx126x ssd1289
 TESTS = $(filter-out $(addsuffix /%,$(NOTESTS)),$(DRIVERS))
 
 unit-test:
-	@go test -v $(addprefix ./,$(TESTS)) 
+	@go test -v $(addprefix ./,$(TESTS))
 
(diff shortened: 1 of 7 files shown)


bgould:
Thank you for your contribution!

These looks good to me, though I don't have this sensor so I can't test it myself.

Also I'm not sure but maybe `dht` can/should be removed from here? https://github.com/tinygo-org/drivers/blob/a15f2167cc7e46864d82e08904a3ddaa3e610d5e/Makefile#L230-L234

EDIT: I see why `dht` is probably in the NOTESTS section... a lot of those files reference the `machine` package directly.  To include the running the unit test from the Makefile would like require a little bit of moving things around to make sure we aren't referencing `machine` in the files with testable functions, coupled with putting build tags like `// +build tinygo` in files that import `machine`

doniacld:
> // +build tinygo

Hi @bgould,
I have been reworking my PR to make the UT works by creating a dedicated file for `device`.
From my perspective, the [constants.go](https://github.com/tinygo-org/drivers/blob/release/dht/constants.go) file should be reworked since it contains heterogeneous content but it could be done in a second iteration.

Running `make unit-test`

```bash
=== RUN   TestDeviceType_extractData
=== RUN   TestDeviceType_extractData/DHT22
--- PASS: TestDeviceType_extractData (0.00s)
    --- PASS: TestDeviceType_extractData/DHT22 (0.00s)
PASS
ok      tinygo.org/x/drivers/dht        0.470s
```

bgould:
Thank you, I had the same thought that some of the stuff in constants.go maybe doesn't belong in there.  The refactor looks good to me, just looking into the CI failure (which is unrelated to your PR) to see if that should be fixed before merging.

doniacld:
Thanks, @bgould let me know about the issue, I saw it was reverted in tiny-go.

deadprogram:
Thanks very much for the code contribution @doniacld and thanks @bgould for helping review.

Now merging.

PR #391, opened 2022-03-01
ws2812: support high-MHz ARMv6M chips like the RP2040

aykevl:
The possible branch distance is a lot shorter on ARMv6M (Cortex-M and Cortex-M0+) for conditional branches. Therefore, convert this long conditional branch into an unconditional branch.

This probably makes the code a little bit slower but because it is in the low period of the WS2812 signal it shouldn't matter for the protocol. And it avoids difficult workarounds specifically for the RP2040.


deadprogram:
Thanks @aykevl

PR #392, opened 2022-03-02
Add support for '8-step mode' to easystepper

neildavis:
Currently, easystepper driver supports only '4-step mode' (12-23-34-41) when energizing the coils
This does not work with all motors & drivers (e.g. 28BJY-48 & ULN2003)

This PR adds support for 8-step mode (1-12-2-23-3-34-4-41)

Source compatibility has been maintained at the expense of a (now) redundant public API
(Move() func's are retained but call the new MoveStepsMode() APIs using 4-step mode param)
since Go does not support default values for parameters.

This PR also encapsulates the API using interfaces instead of exposing the internal
Device structs to the caller.

The example has also been updated to make use of both 'Move' APIs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -227,6 +227,8 @@ endif
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.uf2 -target=pico ./examples/ssd1289/main.go
 	@md5sum ./build/test.uf2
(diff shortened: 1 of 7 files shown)


neildavis:
Sorry, selected the wrong branch and couldn't find a way to edit it. So closed this one and opened #394

PR #394, opened 2022-03-02
Add support for '8-step mode' to easystepper (#393)

### LINKED ISSUE #393: easystepper should support '8-step' mode
Currently, easystepper driver supports only '4-step mode' (12-23-34-41) when energizing the coils
This does not work with all motors & drivers (e.g. 28BJY-48 & ULN2003)

Support for 8-step mode (1-12-2-23-3-34-4-41) should be added.

#394 resolves this

neildavis:
Currently, easystepper driver supports only '4-step mode' (12-23-34-41) when energizing the coils
This does not work with all motors & drivers (e.g. 28BJY-48 & ULN2003)

This commit adds support for 8-step mode (1-12-2-23-3-34-4-41)

Note: this commit breaks backward source compatibility for the factory/constructor functions.

Fixes #393

### DIFF
--- a/easystepper/easystepper.go
+++ b/easystepper/easystepper.go
@@ -2,28 +2,76 @@
 package easystepper // import "tinygo.org/x/drivers/easystepper"
 
 import (
+	"errors"
 	"machine"
 	"time"
 )
 
+// StepMode determines the coil sequence used to perform a single step
+type StepMode uint8
+
+// Valid values for StepMode
+const (
+	// ModeFour uses a 'four step' coil sequence (12-23-34-41). This is the default (zero-value) mode
+	ModeFour StepMode = iota
+	// ModeEight uses an 'eight step' coil sequence (1-12-2-23-3-34-4-41)
+	ModeEight
+)
+
+// stepCount is a helper function to return the number of steps in a StepMode sequence
+func (sm StepMode) stepCount() uint {
+	switch sm {
+	default:
+		fallthrough
+	case ModeFour:
+		return 4
+	case ModeEight:
+		return 8
+	}
+}
+
+// DeviceConfig contains the configuration data for a single easystepper driver
+type DeviceConfig struct {
+	// Pin1 ... Pin4 determines the pins to configure and use for the device
+	Pin1, Pin2, Pin3, Pin4 machine.Pin
+	// StepCount is the number of steps required to perform a full revolution of the stepper motor
+	StepCount uint
+	// RPM determines the speed of the stepper motor in 'Revolutions per Minute'
+	RPM uint
+	// Mode determines the coil sequence used to perform a single step
+	Mode StepMode
+}
+
+// DualDeviceConfig contains the configuration data for a dual easystepper driver
+type DualDeviceConfig struct {
+	DeviceConfig
+	// Pin5 ... Pin8 determines the pins to configure and use for the second device
+	Pin5, Pin6, Pin7, Pin8 machine.Pin
+}
+
 // Device holds the pins and the delay between steps
 type Device struct {
 	pins       [4]machine.Pin
-	stepDelay  int32
+	stepDelay  time.Duration
 	stepNumber uint8
+	stepMode   StepMode
 }
 
 // DualDevice holds information for controlling 2 motors
 type DualDevice struct {
-	devices [2]Device
+	devices [2]*Device
 }
 
-// New returns a new easystepper driver given 4 pins, number of steps and rpm
-func New(pin1, pin2, pin3, pin4 machine.Pin, steps int32, rpm int32) Device {
-	return Device{
-		pins:      [4]machine.Pin{pin1, pin2, pin3, pin4},
-		stepDelay: 60000000 / (steps * rpm),
+// New returns a new single easystepper driver given a DeviceConfig
+func New(config DeviceConfig) (*Device, error) {
+	if config.StepCount == 0 || config.RPM == 0 {
+		return nil, errors.New("config.StepCount and config.RPM must be > 0")
 	}
+	return &Device{
+		pins:      [4]machine.Pin{config.Pin1, config.Pin2, config.Pin3, config.Pin4},
+		stepDelay: time.Second * 60 / time.Duration((config.StepCount * config.RPM)),
+		stepMode:  config.Mode,
+	}, nil
 }
 
 // Configure configures the pins of the Device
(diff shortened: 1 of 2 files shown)


soypat:
Why have two new interfaces been defined? Why can't they be plain structs?

neildavis:
> Why have two new interfaces been defined? Why can't they be plain structs?

Probably a misguided attempt on my behalf at encapsulation, [generality](https://go.dev/doc/effective_go#generality) and a bias for interface-based programming.

I have seen both approaches in this project, e.g. [keypad4x4](https://github.com/tinygo-org/drivers/blob/release/keypad4x4/keypad4x4.go#L11-L34) also takes this approach. 

However, on reflection and looking deeper I see this is less common and in this case without a need for polymorphism I agree it is overkill. I will push an update to remove the interfaces and revert to returning the Device structs.

Thanks for the review.

neildavis:
Can anyone please advise on how to get the CircleCI checks to run? Thanks.

neildavis:
@soypat I removed the interfaces and rebased against dev. Could you re-review please? Thanks.

easystepper/easystepper.go
```diff
@@ -6,11 +6,21 @@ import (
 	"time"
 )
 
+// StepMode determines the number of coil configurations for a single step
+type StepMode uint8
+
+// Valid values for StepMode
+const (
+	FourStepMode  = 4 // 12-23-34-41
```
soypat: These should be of StepMode type! I'd also go with a prefix naming style for Mode/type/kind enums like so:
```go
ModeFour  StepMode = 4 // 12-23-34-41
ModeEight StepMode = 8 // 1-12-2-23-3-34-4-41
```
Having the `Mode` prefix will make them appear together in intellisense and documentation.

easystepper/easystepper.go
```diff
@@ -19,10 +29,17 @@ type DualDevice struct {
 }
 
 // New returns a new easystepper driver given 4 pins, number of steps and rpm
+// This driver will use the 4-Step sequence mode (12-23-34-41) to energize the coils
 func New(pin1, pin2, pin3, pin4 machine.Pin, steps int32, rpm int32) Device {
+	return NewWithMode(pin1, pin2, pin3, pin4, steps, rpm, FourStepMode)
+}
+
+// NewWithMode returns a new easystepper driver given 4 pins, number of steps, rpm and a step mode
+func NewWithMode(pin1, pin2, pin3, pin4 machine.Pin, steps int32, rpm int32, mode StepMode) Device {
```
soypat: Can we define a `StepConfig` type? i.e.
```go
type StepConfig struct {
    // documentation
    Pin1, Pin2, Pin3, Pin4 machine.Pin
    // documentation
    RPM int32
    Mode StepMode
}
type DualStepConfig struct {
     StepConfig
     Pin5, Pin6, Pin7, Pin8 machine.Pin
}
```
We do this for PWM types in the machine package with less arguments. I feel it is justified since it would really clean up the function signatures, which are so very long right now.

easystepper/easystepper.go
```diff
@@ -119,6 +146,15 @@ func (d *DualDevice) Off() {
 
 // stepMotor changes the pins' state to the correct step
 func (d *Device) stepMotor(step uint8) {
+	if d.stepMode == FourStepMode {
```
soypat: Switch statement might be nicer here but I wouldn't worry too much :)

soypat (changes requested):
Review incoming. Sorry I didn't pick up on these changes beforehand!

easystepper/easystepper.go
neildavis: @soypat I like this idea! I'm in favour of improving APIs wherever possible. Just to clarify, do you suggest changing the signature of the existing `New(...)` func too, or only using these config structs for the 8-step enhancement? I was clearly trying to maintain backwards source compatibility, but I'm not sure if this is a requirement and/or even desirable on a pre 1.0 release? What do you think?

easystepper/easystepper.go
neildavis: Personally, I prefer to use a `switch` only when the number of cases > 2 but that's probably my long ingrained bias from my earlier days using C/C++ where it was proven more efficient. I'd hope modern compiler optimization makes this a style choice :)

neildavis:
@soypat Thanks for the review. I've pushed changes to reflect the suggestions you made. I took the liberty of breaking backwards source-compatibility when using a config struct. Hopefully this isn't an issue for a pre-1.0 release?

Using a config struct introduced a few other issues to deal with, like handling default zero-values in the struct when not specified by key. So I refactored a little to include ensuring a valid default step mode and also introducing some error handling in the constructors if the zero-default (non-sensical) step count and speed (RPM) values are provided.

Please take a fresh look when you have time. Thanks!

easystepper/easystepper.go
```diff
@@ -2,28 +2,76 @@
...
+		return 8
+	}
+}
+
+// DeviceConfig contains the configuration data for a single easystepper driver
+type DeviceConfig struct {
+	// Pin1 ... Pin4 determines the pins to configure and use for the device
+	Pin1, Pin2, Pin3, Pin4 machine.Pin
+	// StepCount is the number of steps required to perform a full revolution of the stepper motor
+	StepCount uint
```
soypat, line 38: It makes sense  to set StepCount and RPM to uint types since it makes no physical sense for it to be a negative number, but this can hide user input errors like bad calculations, i.e: calculating RPM from radians per second or the like. It's usually a good idea to use the `int` type for all integer inputs except in specific cases. This is my personal opinion and comes from experience using other Go APIs I've seen in the wild.

soypat:
@deadprogram and @sago35 it'd be good to get second pair of eyes on this. I hope I didn't overstep the line in suggesting breaking changes.

easystepper/easystepper.go
neildavis: Hmm, that's a good point. 

~~I'm also thinking it should maybe be explicit in size too, i.e. `int32`? I'm not sure of the rules for a 'machine word' in TinyGo, but if `int` & `uint` are only 8-bits on 8-bit MCUs (e.g. AVR) then it should be explicit to provide for a larger range than 127!~~

OK, so thanks to niaow on Gophers slack #tinygo I now understand that `int` & `uint` are minimum 32-bit, so a signed 32-bit var should be sufficient here for both `StepCount` and `RPM` I'll make this change and push after any further comments/changes.

deadprogram:
@conejoninja did the original work on easystepper. Any opinions on this?

soypat:
Two weeks have passed- I say break the API. The following function would be a pain point for small monitor users if extended for 8 pin steppers.
```go
func NewWithMode(pin1, pin2, pin3, pin4 machine.Pin, steps int32, rpm int32, mode StepMode) Device
```

deadprogram:
Overdue merge proceeding...

PR #396, opened 2022-03-10
ili9341: add support for atsame5x

sago35:
This PR adds support for atsame5x.


deadprogram:

PR #397, opened 2022-03-10
ili9341: added Feather board support to InitDisplay()

sago35:
This PR adds the following feather board support to the ili9341 example.
DC, SS_PIN, etc. need to be connected separately, but those who have a feather board may find it helpful.

* feather-m0
* feather-m4
* feather-m4-can
* feather-nrf52840
* feather-nrf52840-sense
* feather-stm32f405
* feather-rp2040


deadprogram:
Thanks @sago35 now merging.

PR #399, opened 2022-03-24
Add l3gd20 gyroscope driver

soypat:
Gyroscope driver with example.

### DIFF
--- a/examples/l3gd20/main.go
+++ b/examples/l3gd20/main.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/l3gd20"
+)
+
+func main() {
+	const (
+		// Default address on most breakout boards.
+		pcaAddr = 0x40
+	)
+
+	bus := machine.I2C0
+
+	err := bus.Configure(machine.I2CConfig{})
+	if err != nil {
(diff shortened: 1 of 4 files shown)


examples/l3gd20/main.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		x, y, z = gyro.AngularVelocity()
+		println(x, y, z)
+		time.Sleep(500 * time.Millisecond)
+	}
+}
+
+// ScanI2CDev finds I2C devices on the bus and rreturns them inside
+// a slice. If slice is nil then no devices were found.
+func ScanI2CDev(bus *machine.I2C) (addrs []uint8) {
```
deadprogram: Cool function, but is not being called and also probably doesn't belong in this example.

Separate note, I worked on this a little last year, but just found the snippet and made a gist now: https://gist.github.com/deadprogram/6e1f3ecdb79cf223db2fdaeb4534bfe6

deadprogram: So perhaps this function can be removed from here?

soypat: Yes! I'll fix this ASAP

soypat: Done!

deadprogram:
Thanks @soypat now merging.

PR #403, opened 2022-04-19
vl53l1x: Fix switch-case semantics

ostaebler:
The source code has been ported from C. It seems, the different
semantics of switch-case in Go has not been ported properly.
To fix this, remove 'break' statements and introduce 'fallthrough' where
necessary.

Signed-off-by: Oliver Stäbler <oliver.staebler@bytesatwork.ch>

### DIFF
--- a/vl53l1x/vl53l1x.go
+++ b/vl53l1x/vl53l1x.go
@@ -157,7 +157,6 @@ func (d *Device) SetDistanceMode(mode DistanceMode) bool {
 		d.writeReg(SD_CONFIG_WOI_SD1, 0x05)
 		d.writeReg(SD_CONFIG_INITIAL_PHASE_SD0, 6)
 		d.writeReg(SD_CONFIG_INITIAL_PHASE_SD1, 6)
-		break
 	case MEDIUM:
 		// timing config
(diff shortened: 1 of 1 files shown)


vl53l1x/vl53l1x.go
```diff
@@ -334,51 +331,44 @@ func (d *Device) getRangingData() {
 	d.rangingData.mm = uint16((uint32(d.results.mmCrosstalkSD0)*2011 + 0x0400) / 0x0800)
 	switch d.results.status {
 	case 17: // MULTCLIPFAIL
```
deadprogram: These lines could be reduced to

`case 1, 2, 3, 17:`

instead of using `fallthrough`

EDIT: also better to put in order.

ostaebler: I've been discussing exactly this with my colleague, but we opted for keeping the comments and going with 'fallthrough'.
How about this _middle ground_?

deadprogram: Seems good to me. Thanks @ostaebler

deadprogram (approved):

PR #405, opened 2022-04-24
sdcard: support thingplus-rp2040

oflebbe:
The instances of SPI interfaces SPI0, SPI1, ... are
pointers to SPI types on rp2040 machines, rather plain
values like other machines.

This needs restructuring the sdcard API to take a pointer
to the SPI interface rather the SPI type itself.

A different issue is the Serial.DTR signal to be used in the example code.
The current UART code does not implement the modem control signals. Here is a
workaround for not having this functionality. Maybe it would be simpler to
simply remove waiting for serial line to come alive ?

### DIFF
--- a/examples/sdcard/console/feather-m4.go
+++ b/examples/sdcard/console/feather-m4.go
@@ -8,7 +8,7 @@ import (
 )
 
 func init() {
-	spi = machine.SPI0
+	spi = &machine.SPI0
 	sckPin = machine.SPI0_SCK_PIN
 	sdoPin = machine.SPI0_SDO_PIN
 	sdiPin = machine.SPI0_SDI_PIN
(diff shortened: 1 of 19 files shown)


deadprogram:
> This needs restructuring the sdcard API to take a pointer to the SPI interface rather the SPI type itself.

This is a good thing anyhow. Most of the other SPI devices are using the `drivers.SPI` interface. The only reason the SD care does not (yet?) is because supporting `Configure()` was not quite as simple, and the SDCard driver needs that to reconfigure the interface to kick up to high-speed.

So in short, this seems like a fine change to make.

> Maybe it would be simpler to simply remove waiting for serial line to come alive ?

I am in favor of removing that code as well.

oflebbe:
Updated PR

sago35 (approved):
@oflebbe @deadprogram 
Looks good to me.

sago35 (approved):
@oflebbe 
LGTM

deadprogram:
Thank you for the contribution @oflebbe and for the review @sago35 now merging.

PR #406, opened 2022-04-30
scd4x: implementation for CO2 sensor

deadprogram:
This PR adds the scd4x co2 sensor.

https://www.adafruit.com/product/5187

Needs to be tested on real hardware, hence draft.

EDIT: now tested on real hardware and working.


deadprogram:
Thanks for the corrections @conejoninja now merging.

PR #407, opened 2022-04-30
ws2812: support thingplus-rp2040 board

oflebbe:
Added 125 MHz rp2040 timing
  Added unsafe.Pointer for pointer conversion

### DIFF
--- a/examples/ws2812/thingplus-rp2040.go
+++ b/examples/ws2812/thingplus-rp2040.go
@@ -1,0 +1,13 @@
+//go:build thingplus_rp2040
+// +build thingplus_rp2040
+
+package main
+
+import "machine"
+
+// This is the pin assignment for the internal neopixel of the
+// Sparkfun thingplus rp2040.
+// Replace neo and led in the code below to match the pin
+// that you are using if different.
+var neo machine.Pin = machine.GPIO8
+var led = machine.LED
--- a/ws2812/gen-ws2812.go
+++ b/ws2812/gen-ws2812.go
@@ -252,7 +252,7 @@ func writeGoWrapper(f *os.File, arch string, megahertz int) error {
 	fmt.Fprintf(buf, "	portClear, maskClear := d.Pin.PortMaskClear()\n")
 	fmt.Fprintf(buf, "\n")
 	fmt.Fprintf(buf, "	mask := interrupt.Disable()\n")
-	fmt.Fprintf(buf, "	C.ws2812_writeByte%d(C.char(c), portSet, portClear, maskSet, maskClear)\n", megahertz)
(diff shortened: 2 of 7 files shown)


examples/ws2812/thingplus-rp2040.go
```diff
@@ -0,0 +1,12 @@
+//go:build thingplus_rp2040
+// +build thingplus_rp2040
+
+package main
+
+import "machine"
+
+// This is the pin assignment for the Digispark only.
```
deadprogram: "Digispark"? :smiley_cat:

ws2812/gen-ws2812.go
```diff
@@ -252,7 +252,7 @@ func writeGoWrapper(f *os.File, arch string, megahertz int) error {
 	fmt.Fprintf(buf, "	portClear, maskClear := d.Pin.PortMaskClear()\n")
 	fmt.Fprintf(buf, "\n")
 	fmt.Fprintf(buf, "	mask := interrupt.Disable()\n")
-	fmt.Fprintf(buf, "	C.ws2812_writeByte%d(C.char(c), portSet, portClear, maskSet, maskClear)\n", megahertz)
+	fmt.Fprintf(buf, "	C.ws2812_writeByte%d(C.char(c), (*uint32)(unsafe.Pointer(portSet)), (*uint32)(unsafe.Pointer(portClear)), maskSet, maskClear)\n", megahertz)
```
aykevl, line 255: RP2040 is the odd one here, it returns a `*volatile.Register32` instead of a `*uint32` like all the others.
I don't remember the reasoning behind the `*uint32` API but I think it would be better to fix it in the machine package instead of working around it here.

examples/ws2812/thingplus-rp2040.go
oflebbe: sparkfun omg

oflebbe: Was a cut & paste error, resolved.

ws2812/gen-ws2812.go
oflebbe: The volatile.Register32 seems to come straight from the SVD file of the rp2040. 
Here is a possible way to align https://github.com/tinygo-org/tinygo/pull/2872/files

deadprogram:
Thank you very much for working on this @oflebbe now merging.

oflebbe:
Thanks for merging.  I am asking myself if I shouldn't revert the change to `gen-ws2812.go` since I aligned the API with this merged https://github.com/tinygo-org/tinygo/pull/2872 PR ? 
Sorry, I was AFK for a few days :)

deadprogram:
Hmm, probably so. I didn't think of that.

PR #409, opened 2022-05-05
vl53l1x: Add functions for setting the device address

baw-michael:
This PR implements functions to get and set the device address similar to its
reference implementation in C:
https://github.com/pololu/vl53l1x-arduino/blob/master/VL53L1X.cpp#L28

The field `Address` of type `Device` could be changed to a private now; but I kept it because it would be a breaking change.


deadprogram:
Looks good to me @baw-michael thank you for the code contribution. Now merging.

PR #410, opened 2022-05-06
lsm6ds3tr: initial implementation

ysoldak:
This is an initial implementation of [lsm6ds3tr](https://www.st.com/en/mems-and-sensors/lsm6ds3tr-c.html) 6-axis IMU driver.

iNEMO 6DoF inertial measurement unit (IMU), for entry level / mid-tier smart phones and Portable PC platforms

lsm6ds3tr is a different version of already supported lsm6ds3.
According to [datasheet](https://www.st.com/resource/en/datasheet/lsm6ds3tr-c.pdf), this IMU does have differences in registers, etc to justify a separate driver.

[XIAO BLE Sense](https://www.seeedstudio.com/Seeed-XIAO-BLE-Sense-nRF52840-p-5253.html) boards (tinygo target "xiao_ble") have this IMU installed.

### DIFF
--- a/examples/lsm6ds3tr/main.go
+++ b/examples/lsm6ds3tr/main.go
@@ -1,0 +1,37 @@
+// Connects to an LSM6DS3TR I2C a 6 axis Inertial Measurement Unit (IMU)
+package main
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/lsm6ds3tr"
+)
+
+func main() {
+	machine.I2C0.Configure(machine.I2CConfig{})
+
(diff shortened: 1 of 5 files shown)


sago35:
@ysoldak 
I ran the EXAMPLE with XIAO BLE Sense.
It appears to be working correctly.

I have not checked the source code yet.
Please let me know when the changes are fixed.

ysoldak:
@sago35 What do you mean? If PR is ready for review? Yes, it is, please, review the code.
I use this driver in my head tracker project that uses both Acc and Gyro, seem to be working correctly.

deadprogram (approved):
Looking good to me @ysoldak thanks!

PR #414, opened 2022-05-19
wifinina: add mutex to prevent communication race problems

deadprogram:
This PR adds a mutex to the WifiNINA driver to prevent communication race problems when multiple goroutines try to access adaptor at the same time.

### DIFF
--- a/wifinina/wifinina.go
+++ b/wifinina/wifinina.go
@@ -12,6 +12,7 @@ import (
 	"fmt" // used only in debug printouts and is optimized out when debugging is disabled
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"machine"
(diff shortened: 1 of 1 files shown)


ysoldak:
Why don’t put mutex on all public methods? Basically any “concurrent” access to wifinina upsets communication. Like one goroutune does networking, another does RGB leds…

deadprogram:
@ysoldak I tried to only use the mutex where it seemed absolutely needed. Are there any specific functions you think I should cover that I missed, or is this more of a general comment?

ysoldak:
@deadprogram general comment, yes.

deadprogram:
That being the case, I will merge this and then further improvements can continue from there. Thanks @ysoldak

PR #423, opened 2022-07-11
Fix irremote reporting incorrect NEC addresses and command codes (#422)

### LINKED ISSUE #422: irremote reports incorrect NEC addresses & command codes
Found whilst adding support for #421:

Although the 4 constituent bytes of the 32-bit data packet are read in the correct order, and the 'inverse' (logical NOT) verification passes, the bit order in each byte is reversed. They are read MSB->LSB whereas the protocol requires LSB->MSB.

This means that:
1. NEC commands whose binary representation is non-palindromic (i.e. most of them) are incorrectly reported. e.g. `0xA2` (`0b10100010`) is reported as  `0x45` (`0b01000101`)
2. NEC non-extended range addresses in the 8-bit range whose binary representation is non-palindromic  (i.e. most of them) are incorrectly reported in the same way as commands.
3. NEC extended range 16-bit addresses are incorrectly reported.

Resolved by #423, although it's likely to be superceded shortly by PR for #421 which will also include the fix as part of a refactor.

neildavis:
Change from reading MSB->LSB to LSB->MSB
Fixes #422


deadprogram:
@neildavis could you please rebase this PR against `dev` branch so the build gets triggered and we can verify your changes? Thanks!

deadprogram:
Change looks good to me! Now merging thanks @neildavis

PR #427, opened 2022-07-13
Makefile recursively finds unit-tests (#426)

### LINKED ISSUE #426: Makefile: unit-tests target improvements
Currently  `Makefile` assumes all dirs (drivers or not) either have unit tests, or will at least run under `go test` A manual  list (`$(NOTESTS)`) variable  is used to exclude anything problematic (usually code that imports `machine` which doesn't compile under 'vanilla go' used for running `go test`)

This is rather inflexible and has a few problems:
1. It requires manually updating for each new driver that uses the 'machine` package (or non-driver content)
2. It's impossible to nest packages, in particular say `parent/child` where `child` may have unit tests.
3. The output of `make unit-test` contains many superfluous lines like: `?   	tinygo.org/x/drivers/adxl345	[no test files]` where the package doesn't have any tests but doesn't fail either.

I propose an improvement to recursively identify which dirs contain files with names matching `*_test.go` and only run `go test` on those dirs. A manual exclusion list can be retained for convenience.

I've submitted #427 as a candidate to resolve this.

neildavis:
Fixes #426

This commit modifies how the Makefile selects dirs for the unit-test target
Instead of assuming all dirs have tests and manually excluding those that don't,
it now recursively finds all dirs containing files with names matching "*_test.go"

A manual exclusion list is maintained for operational use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,13 +242,14 @@ endif
 	tinygo build -size short -o ./build/test.uf2 -target=circuitplay-express ./examples/makeybutton/main.go
 	@md5sum ./build/test.uf2
 
-DRIVERS = $(wildcard */)
-NOTESTS = build examples flash semihosting pcd8544 shiftregister st7789 microphone mcp3008 gps microbitmatrix \
-		hcsr04 ssd1331 ws2812 thermistor apa102 easystepper ssd1351 ili9341 wifinina shifter hub75 \
-		hd44780 buzzer ssd1306 espat l9110x st7735 bmi160 l293x keypad4x4 max72xx p1am tone tm1637 \
-		pcf8563 mcp2515 servo sdcard rtl8720dn image cmd i2csoft hts221 lps22hb apds9960 axp192 xpt2046 \
-		ft6336 sx126x ssd1289 irremote uc8151 makeybutton
-TESTS = $(filter-out $(addsuffix /%,$(NOTESTS)),$(DRIVERS))
+# rwildcard is a recursive version of $(wildcard) 
+# https://blog.jgc.org/2011/07/gnu-make-recursive-wildcard-function.html
+rwildcard=$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $(subst *,%,$2),$d))
+# Recursively find all *_test.go files from cwd & reduce to unique dir names
+HAS_TESTS = $(sort $(dir $(call rwildcard,,*_test.go)))
+# Exclude anything we explicitly don't want to test for whatever reason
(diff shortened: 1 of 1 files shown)


deadprogram:
This seems like an excellent idea to me. Can you please resolve merge conflict?

neildavis:
> This seems like an excellent idea to me. Can you please resolve merge conflict?

Thanks. Done.

Makefile
```diff
@@ -242,13 +242,14 @@ endif
...
-		hcsr04 ssd1331 ws2812 thermistor apa102 easystepper ssd1351 ili9341 wifinina shifter hub75 \
-		hd44780 buzzer ssd1306 espat l9110x st7735 bmi160 l293x keypad4x4 max72xx p1am tone tm1637 \
-		pcf8563 mcp2515 servo sdcard rtl8720dn image cmd i2csoft hts221 lps22hb apds9960 axp192 xpt2046 \
-		ft6336 sx126x ssd1289 irremote uc8151 makeybutton
-TESTS = $(filter-out $(addsuffix /%,$(NOTESTS)),$(DRIVERS))
+# rwildcard is a recursive version of $(wildcard) 
+# https://blog.jgc.org/2011/07/gnu-make-recursive-wildcard-function.html
+rwildcard=$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $(subst *,%,$2),$d))
+# Recursively find all *_test.go files from cwd & reduce to unique dir names
+HAS_TESTS = $(sort $(dir $(call rwildcard,,*_test.go)))
```
sago35, line 249: Simple question.
Is this part of the result the same as the following command?

```
git ls-files "*_test.go"
```

deadprogram: It appears to give same result for me.

neildavis:
```shell
HAS_TESTS = $(sort $(dir $(shell git ls-files *_test.go)))
```
This would work if you prefer shell-ing out to `git` over native make constructs? Normally I try to avoid this on compatibility grounds, but `git` should be pretty universal, and it is more readable.

I can push & squash this change if required?

deadprogram: There is just so much inside of make, I did not know about until this PR.

Seems like a higher probability that the `git` command might be maintainable by non-Make gnurus. At the same time, using the innate makefile commands is also totally reasonable.

sago35: I too think either would be fine.
So I don't think it is necessary to change the current commit.

sago35 (approved):

deadprogram:
Thanks for working on this @neildavis now merging.

PR #429, opened 2022-07-17
makeybutton: add driver for MakeyMakey-like button

deadprogram:
This PR adds a driver for a MakeyMakey-like button (https://makeymakey.com/)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,13 +239,15 @@ endif
 	@md5sum ./build/test.hex
 	tinygo build -size short -o ./build/test.uf2 -target=pico ./examples/scd4x/main.go
 	@md5sum ./build/test.uf2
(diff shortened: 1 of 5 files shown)


sago35:
How is makeybutton connected to the microcontroller?

conejoninja:
> How is makeybutton connected to the microcontroller?

it isn't?

Makeymakey has its own micro (atmega32u4), by the example provided I guess this is kind of an interface similar to makeymakey. Otherwise it will need an special firmware to be connected as a device ??

deadprogram:
This is not to connect a MakeyMakey. It is to use a pin of whatever MCU to act like a MakeyMakey does by connecting a high Ohm resistor.

deadprogram:
Circuit:
image

conejoninja:
It's ok for me, merging

PR #430, opened 2022-07-18
rtl8720dn: add ./examples/rtl8720dn/version

sago35:
Created an example to read the version of rtl8720dn.

Also, wioterminal rtl8720dn initialization was created in examples/rtl8720dn/wioterminal.go.
Right now, there are many examples/rtl8720dn/*/wioterminal.go, but eventually we would like to unify them into examples/rtl8720dn/wioterminal.go.

The version readout yields the following results

```
RTL8270DN Firmware Version: 2.1.2
```

### DIFF
--- a/examples/rtl8720dn/version/main.go
+++ b/examples/rtl8720dn/version/main.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"fmt"
+	"time"
+
+	"tinygo.org/x/drivers/examples/rtl8720dn"
+)
+
+var (
+	debug = false
+)
+
+func main() {
+	err := run()
+	for err != nil {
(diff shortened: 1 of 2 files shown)


examples/rtl8720dn/wioetrminal.go
```diff
@@ -0,0 +1,79 @@
+//go:build wioterminal
```
deadprogram: Check spelling of the name of this file, please.

sago35: I made a mistake.
Fixed and force-pushed.

deadprogram:
Thanks @sago35 now merging.

PR #433, opened 2022-07-20
epd: add waveshare 2.9in (v1)

kenbell:
This is essentially a cut & paste of the 2.13" device by @conejoninja with updated LUTs and default size.


deadprogram:
Thanks @kenbell merging.

PR #435, opened 2022-07-22
net/http: Fix http.Get() with port specification

sago35:
This PR corrects the behavior of `http.Get()`.
Only http is affected, not https.

Two items were corrected.

1. Change to pass only hostname to net.ParseIP
2. Fix Port setting was always 80

### DIFF
--- a/net/http/tinygo.go
+++ b/net/http/tinygo.go
@@ -38,8 +38,16 @@ func (c *Client) doHTTP(req *Request) (*Response, error) {
 	}
 
 	// make TCP connection
-	ip := net.ParseIP(req.URL.Host)
-	raddr := &net.TCPAddr{IP: ip, Port: 80}
+	ip := net.ParseIP(req.URL.Hostname())
+	port := 80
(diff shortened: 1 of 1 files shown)


sago35:
Before the correction, addresses such as http://hostname:8088/ were not accessible.

For example, the following is a std-Go program that listens for 8088, but could not connect to

```go
package main

import (
	"fmt"
	"log"
	"net/http"
)

var (
	ip   = ""
	port = 8088
	cnt  = 0
)

func handler(w http.ResponseWriter, req *http.Request) {
	cnt++
	fmt.Fprintf(w, "hello %d\n", cnt)
	fmt.Printf("%d %#v\n", cnt, req.UserAgent())
}

func main() {
	http.HandleFunc("/", handler)
	fmt.Printf("listen: %s:%d\n", ip, port)
	log.Fatal(http.ListenAndServe(fmt.Sprintf("%s:%d", ip, port), nil))
}
```

deadprogram:
Thanks for the fix @sago35 now merging.

PR #436, opened 2022-07-23
rtl8720dn: add SetupAndConnectToAccessPoint()

sago35:
In many cases, the task of setting up and connecting the rtl8720dn is tedious.
SetupAndConnectToAccessPoint() and simplified the code.
This makes examples/rtl8720dn/webclient etc. as close to the same source code as possible as the PC.


sago35:
Perhaps it would be useful to create a similar structure for wifinina and others.
In that case, instead of examples/rtl8720dn, it might be better to use rtl8720dn/setup.

scottfeldman:
I believe this PR is subsumed by the greater netdev work in drivers:dev.  Netdev gives drivers a common entry point to setup and connect to Wifi.  Apps use that API, regardless of which network driver is being used.  Consequently, the examples/net examples apps aren't specific to one network driver or the other.

PR #439, opened 2022-07-24
net/http: support for cookies when https

sago35:
Until now, cookies were enabled only for HTTP.
Fix implemented and verified to work correctly.

https://httpbin.org/#/Cookies/get_cookies

```go
	// cookie
	jar, err := cookiejar.New(nil)
	if err != nil {
		log.Fatal(err)
	}
	client := &http.Client{Jar: jar}
	http.DefaultClient = client

	url = "https://httpbin.org/cookies/set/name/there"
	res, err := http.Get(url)
	if err != nil {
		log.Fatal(err)
	}
```

and

```go
		url = "https://httpbin.org/cookies"
		res, err := http.Get(url)
		if err != nil {
			log.Fatal(err)
		}
```

### DIFF
--- a/net/http/tinygo.go
+++ b/net/http/tinygo.go
@@ -20,6 +20,11 @@ func SetBuf(b []byte) {
 }
 
 func (c *Client) Do(req *Request) (*Response, error) {
+	if c.Jar != nil {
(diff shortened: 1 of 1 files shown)


deadprogram (approved):

deadprogram:

PR #443, opened 2022-07-27
makeybutton: some fixes and refactoring

deadprogram:
This PR fixes the makeybutton by reducing the size of the buffer for better responsiveness. It also add a `Configure()` method to bring it in line with the typical device patterns.

### DIFF
--- a/makeybutton/button.go
+++ b/makeybutton/button.go
@@ -6,7 +6,10 @@
 //
 package makeybutton
 
-import "machine"
+import (
+	"machine"
+	"time"
+)
 
 // ButtonState represents the state of a MakeyButton.
 type ButtonState int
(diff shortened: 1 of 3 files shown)


makeybutton/button.go
```diff
@@ -47,6 +44,15 @@ func NewButton(pin machine.Pin) *Button {
 	}
 }
 
+// Configure configures the Makey Button pin to have the correct settings to detect touches.
+func (b *Button) Configure() error {
+	b.pin.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
```
conejoninja, line 54: Is this needed if the next line changes the mode too ?

deadprogram: It is needed to set the pin into the correct floating mode. A little weird, but needed.

makeybutton/button.go
```diff
@@ -61,7 +67,7 @@ func (b *Button) Get() ButtonEvent {
 	b.readings.Put(pressed)
 
 	switch {
-	case pressed && avg > 0:
+	case pressed && avg > -1*bufferSize-2:
```
conejoninja: bufferSize is 6, avg could take a value between -6 and +6 
-1*bufferSize-2 is -8
avg > -1*bufferSize-2 willl always be true¿?

makeybutton/button.go
sago35: If a comment or other information is not added, it will be accidentally erased during future changes.
So, I would suggest adding a comment.

makeybutton/button.go
deadprogram: Oops! Corrected.

makeybutton/button.go
deadprogram: Added.

conejoninja:
It looks good to me with the fixes proposed, is there any more work to be done or it could be merged?

deadprogram:
I say merge! :)

conejoninja:
thanks! merging

PR #444, opened 2022-07-29
bme280: improve config support

kenbell:
This breaks back compat by adding a parameter to the "Config" method, inline with other devices.

The extra config allows the device to be used in 'Forced' mode for more power-efficient use-cases.

### DIFF
--- a/bme280/bme280.go
+++ b/bme280/bme280.go
@@ -8,6 +8,7 @@ package bme280
 
 import (
 	"math"
+	"time"
 
 	"tinygo.org/x/drivers"
 )
@@ -34,11 +35,27 @@ type calibrationCoefficients struct {
 	h6 int8
 }
 
+type Oversampling byte
+type Mode byte
+type FilterCoefficient byte
+type Period byte
+
+// Config contains settings for filtering, sampling, and modes of operation
+type Config struct {
+	Pressure    Oversampling
+	Temperature Oversampling
+	Humidity    Oversampling
+	Period      Period
+	Mode        Mode
+	IIR         FilterCoefficient
+}
+
 // Device wraps an I2C connection to a BME280 device.
 type Device struct {
 	bus                     drivers.I2C
 	Address                 uint16
 	calibrationCoefficients calibrationCoefficients
+	Config                  Config
 }
 
 // New creates a new BME280 connection. The I2C bus must already be
@@ -52,9 +69,34 @@ func New(bus drivers.I2C) Device {
 	}
 }
 
-// Configure sets up the device for communication and
-// read the calibration coefficientes.
+// ConfigureWithSettings sets up the device for communication and
+// read the calibration coefficients.
+//
+// The default configuration is the Indoor Navigation settings
+// from the BME280 datasheet.
 func (d *Device) Configure() {
+	d.ConfigureWithSettings(Config{})
+}
+
+// ConfigureWithSettings sets up the device for communication and
+// read the calibration coefficients.
+//
+// The default configuration if config is left at defaults is
+// the Indoor Navigation settings from the BME280 datasheet.
+func (d *Device) ConfigureWithSettings(config Config) {
+	d.Config = config
+
+	// If config is not initialized, use Indoor Navigation defaults.
+	if d.Config == (Config{}) {
+		d.Config = Config{
+			Mode:        ModeNormal,
+			Period:      Period0_5ms,
+			Temperature: Sampling2X,
+			Humidity:    Sampling1X,
+			Pressure:    Sampling16X,
+			IIR:         Coeff16,
+		}
+	}
 
 	var data [24]byte
 	err := d.bus.ReadRegister(uint8(d.Address), REG_CALIBRATION, data[:])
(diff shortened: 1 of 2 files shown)


deadprogram:
Perhaps this could be added as `ConfigureWithSettings(config Config)` to avoid a breaking change?

deadprogram:
Thank you very much for making my requested changes @kenbell now merging!

deadprogram (approved):

bme280/bme280.go
```diff
@@ -52,9 +69,34 @@ func New(bus drivers.I2C) Device {
...
 func (d *Device) Configure() {
+	d.ConfigureWithSettings(Config{})
+}
+
+// ConfigureWithSettings sets up the device for communication and
+// read the calibration coefficients.
+//
+// The default configuration if config is left at defaults is
+// the Indoor Navigation settings from the BME280 datasheet.
+func (d *Device) ConfigureWithSettings(config Config) {
```
conejoninja, line 86: Why have both Configure & ConfigureWithSettings ? Shouldn't it be better to just call Configure(config Config) ?

bme280/bme280.go
```diff
@@ -52,9 +69,34 @@ func New(bus drivers.I2C) Device {
...
+	d.Config = config
+
+	// If config is not initialized, use Indoor Navigation defaults.
+	if d.Config == (Config{}) {
```
conejoninja, line 90: What happens if the user only set some of the values of the configuration? They will default to 0. I think it's better to check every parameter, like other drivers do.

Also, is there any need for d.Config to be exported?

conejoninja (changes requested):

bme280/bme280.go
kenbell: I had it that way and @deadprogram asked to separate out to avoid breaking back-compat since existing code had 'Configure' with no parameters.

conejoninja: Right, just see the reply, sorry for the question.

bme280/bme280.go
kenbell: I'm not sure I follow, comparing against the 'default' struct is the same as comparing all fields?

Since the default values in the struct are valid values for the fields, I don't think there's anything more to do?

conejoninja: Maybe I'm not understanding it right, this is the configuration if left empty, right?:

```go
Config{
			Mode:        ModeNormal,
			Period:      Period0_5ms,
			Temperature: Sampling2X,
			Humidity:    Sampling1X,
			Pressure:    Sampling16X,
			IIR:         Coeff16,
		}
```

Now, If I just want to set one parameter:

```go
sensor.ConfigureWithSettings(bme280.Config{Humidity: bme280.Sampling2X})
```

I get the following configuration (if I'm not wrong):

```go
d.Config = Config{
			Mode:        ModeSleep, // 0
			Period:      Period0_5ms, // 0 
			Temperature: SamplingOff, // 0
			Humidity:    Sampling2X, // value set by the user
			Pressure:    SamplingOff, // 0
			IIR:         Coeff0, // 0
		}
```

What I expected/wanted: 

```go
Config{
			Mode:        ModeNormal, 
			Period:      Period0_5ms,
			Temperature: Sampling2X,
			Humidity:    Sampling2X, // value set by the user
			Pressure:    Sampling16X,
			IIR:         Coeff16,
		}
```


For me it's a bit strange behavior, I'm forced to set all the parameters instead of only the ones that differs from the "default/none" configuration. Or maybe I'm missing something.

kenbell: Oh, I see - the problem i hit is that `SamplingOff` (Zero) is a valid value, if Configure logic was...
```
if config.Pressure == SamplingOff {
   d.config.Pressure = Sampling2X
}
```

There would be no way to disable sampling of pressure - setting it to 'off' would always be changed to 2x.

One alternative could be to make all the parameters pointers and check against nil?   But that would make it tedious to initialize the config structure?

The other alternative is to make zero an invalid value by bumping all the consts by 1 - which complicates the driver logic in that `0` in the device register would now be `1` in the config struct.  That would also make the register constants less descriptive and couldn't be re-used by app code reading the registers directly.

I'm open for other ways to do defaults...

conejoninja: Ok, it's not the behavior I'll expected but I guess it's good enough.

I only could think of defining doing something like 
```go
const (
	SamplingOff Oversampling = iota+1
	Sampling1X
	Sampling2X
	Sampling4X
	Sampling8X
	Sampling16X
)
```

But then you'll have to do : 
```go
d.Config.Temperature = config.Temperature-1
d.Config.Mode = mode-1
...
```

Maybe a bit too much, so better to leave it as it is? What do you think @deadprogram

kenbell: FYI - looking at periph.io, they've got for a different model for 'default values' (see https://github.com/google/periph/commit/0e5d546d9ab1f630fc2a1e43cdfb06e7b7b8665d)

I don't think we should adopt just for one driver - but it seems a cleaner model in some ways, and could help solve the problem in machine package where default I2C pins create a back-dependency between boards and chipsets.

kenbell: Also created a proposal to adopt something similar to the periph.io model across TinyGo: https://github.com/tinygo-org/tinygo/issues/3152

deadprogram:
Let's merge and continue from there. Thanks @kenbell

PR #445, opened 2022-08-01
rtl8720dn: add Rpc_tcpip_adapter_init_with_timeout()

sago35:
For example, if the firmware of the RTL8720 is out of date, no response will be returned.
This PR defines functions that can time out in such cases.

The reason for not updating Rpc_tcpip_adapter_init directly is that Rpc_tcpip_adapter_init is automatically generated.


scottfeldman:
An alternative to this PR would be to check the fw version before calling tcpip_adapter_init(), and error out if fw version is too old.  I confirmed we can get fw version before calling tcpip_adapter_init().  Current fw version I have on my wioterminal is:

`RTL8720 firmware version : 2.1.2`

PR #447, opened 2022-08-04
bts7960: Add new device package

kostyay:
Added support for bts 7960 motor driver. 
It is based on the C++ code from this library: https://github.com/luisllamasbinaburo/Arduino-BTS7960


bts7960/bts7960.go
```diff
@@ -0,0 +1,77 @@
...
+	lEn            machine.Pin
+	rEn            machine.Pin
+	lPwm           machine.Pin
+	rPwm           machine.Pin
+	rPwmCh, lPwmCh uint8
+	pwm            machine.PWM
+}
+
+// New returns a new motor driver.
+func New(lEn, rEn, lPwm, rPwm machine.Pin, pwm machine.PWM) *Device {
```
aykevl: Can you define and use a PWM interface instead of using `machine.PWM`? It is called differently on other chips.

For example, with the following interface:

```go
type PWM interface {
    Channel(pin machine.Pin) (channel uint8, err error)
    Set(channel uint8, value uint32)
    Top() uint32
}
```

```suggestion
func New(lEn, rEn, lPwm, rPwm machine.Pin, pwm PWM) *Device {
```

bts7960/bts7960.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// Stop stops the motor
+func (d *Device) Stop() {
+	d.pwm.Set(d.lPwmCh, 0)
+	d.pwm.Set(d.rPwmCh, 0)
+}
+
+// Left turns motor left.
+func (d *Device) Left(speed uint32) {
```
aykevl, line 75: Hmm, what about `Clockwise` / `Counterclockwise`? No strong opinion from me, but it might be easier to understand. (I don't know this motor though so I don't know which terminology makes sense).

aykevl (commented):
Can you also add an example and a smoke test? Each driver has an example in the examples directory and a smoke test in the Makefile.

bts7960/bts7960.go
kostyay: The controller has a L and and R pins so I prefer to keep it this way
image

kostyay:
Fixed ur comments

deadprogram:
First of all, thanks for working on this.

SO if you look at both the L293x and the BTS7960, it can be noted that they are both H-bridges and as a result have essentially the same sort of way that you connect to them and use them.

As the Wikipedia page says:

> These circuits are often used in [robotics](https://en.wikipedia.org/wiki/Robotics) and other applications to allow DC motors to run forwards or backwards.

The only difference is that the BTS7960 has a connection to enable separately the forward direction from the reverse direction. 

I would suggest basically copy the structure/format of the L293x driver. Of course rename the pins, and then implement the same API in the manner that the BTS7960 requires. I propose that we define `Motoring` interface:

```go
type Motoring interface {
    Forward()
    ForwardWithSpeed(speed uint32)
    Backward()
    BackwardWithSpeed(speed uint32)
    Stop()
}
```

In the case of the `Forward()` and `Backward()` methods, a device with PWM would use full speed.

In the case of the `ForwardWithSpeed()` and `BackwardWithSpeed()` methods, a device without PWM would ignore the params and use full speed.

This will make it possible to use either the BTS7960 or the L293X with a minimal amount of code changes.

Last comment, the L293X driver does not try to configure the PWM for the user:
https://github.com/tinygo-org/drivers/blob/release/l293x/l293x.go#L88-L89

This is part because not every board has the same PWM implementation, even though they expose the same interface. Just seems like it simplifies things for users.

What do you think? If it seems like a good idea, I can go back to the L293x and make sure it is doing the same thing.

kostyay:
Hey
Sounds good, I will do it :)

PR #448, opened 2022-08-06
rtl8720dn: add UDP close function

sago35:
I noticed that close was not implemented when I tried to use it as a webserver after NTP.
By adding this, UDP can be closed.


deadprogram:
Thank you for the addition @sago35 now merging.

PR #451, opened 2022-08-18
Trim quotes from IP returned by espat's GetDNS()

Elara6331:
The `espat` driver is unable to dial any socket because the IP returned by `GetDNS()` has quotes around it, and then `ConnectTCPSocket()` adds quotes as well, and you end up with a string that looks like `""0.0.0.0""`.

This PR fixes the `GetDNS()` function by removing the quotes at the ends of the string before returning it.


deadprogram:
Thank you for the fix @Arsen6331 now merging.

PR #453, opened 2022-08-28
rtl8720dn: device wrapper refactoring

deadprogram:
This PR is to start a conversation about refactoring the rtl8720dn driver to be a little more like the other wifi drivers.

It also moves the Wioterminal specific hardware setup into a separate function that is part of the driver code instead of the example.

@sago35 see what you think about this approach. It is very similar to what you have already done, mostly just moves things around a bit to hopefully make it a little easier to use.


rtl8720dn/adapter.go
```diff
@@ -1,25 +1,52 @@
...
+func New(uart *machine.UART) *Driver {
+	return &Driver{
+		RTL8720DN: &RTL8720DN {
+			port:  uart,
+			seq:   1,
+			sema:  make(chan bool, 1),
+			debug: false,
+		},
+	}
+}
```
sago35: The CHIP_PU pin is required to initialize the rtl8720dn even if it is not a Wio Terminal.
So I think the following is correct for the New function.

```
func New(uart *machine.UART, en machine.Port) *Driver
```

image

```

File: rtl8720dn/wioterminal.go [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+	machine.RTL8720D_CHIP_PU.High()
+	time.Sleep(1000 * time.Millisecond)
+
+	UART_RTL8720DN = &UARTx{
+		UART: &machine.UART{
+			Buffer: machine.NewRingBuffer(),
+			Bus:    sam.SERCOM0_USART_INT,
+			SERCOM: 0,
+		},
+	}
```
sago35: This part is also required for non-WioTerminal.
The important parts are the following.

```
func (u *UARTx) Read(p []byte) (n int, err error) {
	if u.Buffered() == 0 {
		time.Sleep(1 * time.Millisecond)
		return 0, nil
	}
	return u.UART.Read(p)
}
```

The UARTx definition can be moved to `rtl8720dn/rtl8720dn.go` since it is a common process even at non-Wio Terminal.

sago35 (commented):
@deadprogram 
Thanks for refactoring.
My original implementation was poor and was difficult to understand.
CHIP_PU and UARTx are not processes for Wio Terminal but for RTL8720DN-driver.

rtl8720dn/wioterminal.go
deadprogram: Done.

rtl8720dn/adapter.go
deadprogram: Thanks for the excellent clarification @sago35 

Please see my latest commit that I think implements your feedback

deadprogram: @sago35 if you think my changes are correct now, I will proceed to complete modifying the other rtl8720dn examples in the same way.

sago35: @deadprogram 
I will investigate the source code some more.

It looks like it needs to call UART.Configure after manipulating the CHIP_PU.
So I think we have to pass machine.Pin before uart.Configure to initialize it.

sago35: @deadprogram 
I thought about it again.
Two major tasks were undertaken.
If it looks OK, `git cherry-pick` what you need.

1. remove code for wioterminal from driver
2. [UART3 defined in machine/board_wioterminal.go](https://github.com/tinygo-org/tinygo/pull/3110)

https://github.com/tinygo-org/drivers/tree/rtl8720dn-device-wrapper-sago35

deadprogram: I made some changes inspired by, but not exactly the same, as what you did. I still really wanted to keep the default UART config for Wioterminal in the driver to reduce the amount of code needed for applications.

See what you think @sago35

deadprogram: Any other feedback @sago35 or can I now proceed with the rest of the examples?

rtl8720dn/wioterminal.go
```diff
@@ -0,0 +1,35 @@
...
+	UART_RTL8720DN.Interrupt = interrupt.New(sam.IRQ_SERCOM0_2, handleInterrupt)
+
+	return New(UART_RTL8720DN, machine.PB24, machine.PC24, machine.RTL8720D_CHIP_PU)
+}
+
+func handleInterrupt(interrupt.Interrupt) {
+	// should reset IRQ
+	UART_RTL8720DN.Receive(byte((UART_RTL8720DN.Bus.DATA.Get() & 0xFF)))
+	UART_RTL8720DN.Bus.INTFLAG.SetBits(sam.SERCOM_USART_INT_INTFLAG_RXC)
+}
```
sago35: Starting with TinyGo 0.26, only the following should be required.


```go
func NewForWioterminal() *Driver {
	return New(machine.UART3, machine.PB24, machine.PC24, machine.RTL8720D_CHIP_PU)
}
```

refs: https://github.com/tinygo-org/tinygo/pull/3110

sago35 (commented):
@deadprogram  It would be better to use machine.UART3.
Other than that, I think it is very good.

rtl8720dn/wioterminal.go
deadprogram: I was able to remove a bunch of code as a result of better understanding now. Thank you for your help @sago35

sago35:
@deadprogram  I think all that is left is to format it using go1.19 and the new goimports.

https://pkg.go.dev/golang.org/x/tools/cmd/goimports

deadprogram:
Corrected.

sago35:
LGTM

deadprogram:
@sago35 commits squashed and ready for final review, please.

sago35 (commented):
@deadprogram 
I think it is good.
I am glad to see the amount of code reduced.

I made the following commit.
If all is well, import the changes.

* Add adapter.Debug(debug)
* Update examples/wifi/tcpclient/rtl8720dn.go

deadprogram:
OK all changes made. I think finally ready to merge!

examples/rtl8720dn/webclient-tinyterm/main.go
```diff
@@ -81,8 +81,10 @@ func run() error {
 	}
 
 	adaptor := rtl8720dn.New(machine.UART3, machine.PB24, machine.PC24, machine.RTL8720D_CHIP_PU)
+	adaptor.Debug(debug)
 	adaptor.Configure()
 
+	http.UseDriver(adaptor)
```
sago35, line 87: At this time, it is only required when creating a webserver.

examples/rtl8720dn/webclient/main.go
```diff
@@ -39,8 +39,10 @@ func main() {
 
 func run() error {
 	adaptor := rtl8720dn.New(machine.UART3, machine.PB24, machine.PC24, machine.RTL8720D_CHIP_PU)
+	adaptor.Debug(debug)
 	adaptor.Configure()
 
+	http.UseDriver(adaptor)
```
sago35, line 45: At this time, it is only required when creating a webserver.

sago35 (approved):
LGTM.

`http.UseDriver(adaptor)` is not necessary, but I don't think it needs to be deleted.

deadprogram:
OK, merging. Thanks for all your help @sago35

PR #454, opened 2022-08-30
Initial support for VL6180x sensor

conejoninja:


vl6180x/vl6180x.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// startRange starts the readings
+func (d *Device) startRange() {
+	for d.dataReady() {
+	}
+	d.writeReg(SYSRANGE_START, 0x01)
+}
+
+// IsRangeComplete return true when the reading is complete and
```
ysoldak: and... ?

vl6180x/vl6180x.go
```diff
@@ -0,0 +1,271 @@
+// Package vl6180x provides a driver for the VL6180X time-of-flight
```
ysoldak: I was intrigued by this "time-of-flight" description, never encountered before.
So, the full description would be:
>The VL6180X (sometimes called the VL6180) is a Time of Flight distance sensor like no other you've used! The sensor contains a very tiny laser source, and a matching sensor. The VL6180X can detect the "time of flight", or how long the laser light has taken to bounce back to the sensor. Since it uses a very narrow light source, it is good for determining distance of only the surface directly in front of it.
>
>Unlike sonars that bounce ultrasonic waves, the 'cone' of sensing is very narrow. Unlike IR distance sensors that try to measure the amount of light bounced, the VL6180X is much more precise and doesn't have linearity problems or 'double imaging' where you can't tell if an object is very far or very close.

https://www.adafruit.com/product/3316

Does it make sense to add this description here perhaps?

conejoninja: I'm not sure, if someone is looki g forna vl6180x driver they probably know what the sensor does 😂. We also have the vl53l1x (which I used as a template for this one). If we add the description here it might make sense to add it to vl53l1x too.

ysoldak:
Why so many changes in README.md? Seem to be whitespaces (not really visible in diff)?

vl6180x/vl6180x.go
ysoldak: Yeah, I understand. I was thinking more along the lines someone looking for a proximity sensors supported by tinygo, for their next project, for example. And searching through the repo code would show them this driver too.
And just generally feels good to have at least some short description about devices.
But I'm not insisting, just a suggestion.

vl6180x/vl6180x.go
conejoninja: fixed

conejoninja:
I was surprised at first too , but the whitespaces "fixes" this (before this PR): Screenshot from 2022-09-03 08-05-52


It's just a formatting issue, not really important, but it was done automatically I guess.

And this is after the edit: 
Screenshot from 2022-09-03 08-04-33

deadprogram:
@conejoninja do you want to squash commits before merge, or shall I do it for you?

deadprogram:
Now merging, thanks @conejoninja

PR #458, opened 2022-09-14
makeybutton: better algorithm implementation

deadprogram:
This PR modifies the algorithm to better match the original.

### DIFF
--- a/makeybutton/button.go
+++ b/makeybutton/button.go
@@ -4,11 +4,18 @@
 // Inspired by the amazing MakeyMakey
 // https://makeymakey.com/
 //
+// This code is a reinterpretation of
+// https://github.com/sparkfun/MaKeyMaKey/blob/master/firmware/Arduino/makey_makey/makey_makey.ino
 package makeybutton
 
 import (
 	"machine"
 	"time"
+)
+
+var (
+	pressThreshold   int = 1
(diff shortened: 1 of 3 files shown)


ysoldak:
How do you know it matches original better, trial and error or there is code available to compare with?

makeybutton/button.go
```diff
@@ -49,7 +54,7 @@ func NewButton(pin machine.Pin) *Button {
 
 // Configure configures the Makey Button pin to have the correct settings to detect touches.
 func (b *Button) Configure() error {
-	// Note that we have to first turn on the pullup, and then turn off the pullup,
+	// Note that on AVR we have to first turn on the pullup, and then turn off the pullup,
```
ysoldak, line 60: Same holds for non-AVR? Or this driver for AVR only?

deadprogram: It is only "required" for AVR, but does not do anything bad on the other platforms I tested (mostly rp2040).

deadprogram:
See https://github.com/sparkfun/MaKeyMaKey/blob/master/firmware/Arduino/makey_makey/makey_makey.ino for the "original".

I will add that link to the source for clarity.

ysoldak:
And not approve, but LGTM :D

PR #459, opened 2022-09-14
ssd1306 Add getter method on display buffer

alexjomin:
Hi there, 

I think I could be great to have a getter method to retrieve the buffer, so we could iterate over the slice to apply an effect on the whole screen for example. 

Feel free to comment and thank you for the awesome work on TinyGo!

### DIFF
--- a/ssd1306/ssd1306.go
+++ b/ssd1306/ssd1306.go
@@ -233,6 +233,11 @@ func (d *Device) SetBuffer(buffer []byte) error {
 	return nil
 }
 
+// GetBuffer returns the whole buffer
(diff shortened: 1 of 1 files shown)


deadprogram (approved):
Seems reasonable to me. Can we also add this to the other displays?

@conejoninja @sago35 ?

conejoninja:
Yes, this could be added to other displays with buffer (not all have because of size), I think some of them (mostly monochrome) already have a getter, but I'll check them.

alexjomin:
Thank you for the review, let me know if I can help.

deadprogram:
Thanks @alexjomin now merging.

PR #463, opened 2022-09-25
all: correct go fmt

deadprogram:
This PR correctly runs go fmt on the files in this package.


deadprogram:
I just can't wait to commit this cause the broken builds are bothering me :smile_cat:

PR #464, opened 2022-09-26
Update waveshare-epd examples to use new machine pin constants

s-fairchild:
The examples appear to be using constants that are no longer present.
Updating to make use of GPIO constants present in machine package


deadprogram:
Seems like still some errors @s-fairchild

aykevl:
@s-fairchild these examples were written for the microbit. It is expected you need to modify the examples a little bit for different boards.

s-fairchild:
I assumed something like that was happening when the tests failed. Thank you for the information I didn't realize that was the target board.

PR #465, opened 2022-09-30
wifinina: remove stale comment...code is reached

scottfeldman:
main() calls sendBatch() in a loop, and each time sendBatch() does finish and close the connection, so remove the stale comment.

### DIFF
--- a/examples/wifinina/tcpclient/main.go
+++ b/examples/wifinina/tcpclient/main.go
@@ -106,7 +106,6 @@ func sendBatch() {
 		println("error:", err.Error(), "\r")
 	}
 
-	// Right now this code is never reached. Need a way to trigger it...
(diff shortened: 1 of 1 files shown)


examples/wifinina/tcpclient/main.go
```diff
@@ -67,7 +67,7 @@ func main() {
 func sendBatch() {
 
 	// make TCP connection
-	ip := net.ParseIP(serverIP)
+	ip := net.IP([]byte(serverIP))
```
ysoldak, line 70: This is wrong, calls are not equal.

ysoldak (changes requested):

scottfeldman:
I screwed up and pushed an unrelated change to this PR, so deleting this PR and starting over.

PR #467, opened 2022-10-01
qmi8658c: Add support for the QMI8658C sensor

rcbadiale:
Add initial support for the [QMI8658C Accelerometer and Gyroscope](https://www.qstcorp.com/upload/pdf/202202/%EF%BC%88%E5%B7%B2%E4%BC%A0%EF%BC%89QMI8658C%20datasheet%20rev%200.9.pdf) sensor by QST Solution.

This sensor is used in the [WaveShare RP2040 Round LCD 1.28in](https://www.waveshare.com/wiki/RP2040-LCD-1.28) development board.

Add an example based on the aforementioned development board.

Add smoke test to `makefile` and update `README.md`

### DIFF
--- a/qmi8658c/qmi8658c.go
+++ b/qmi8658c/qmi8658c.go
@@ -1,0 +1,186 @@
+// Package qmi8658c provides a driver for the QMI8658C accelerometer and gyroscope
+// made by QST Solutions.
+//
+// Datasheet:
+// https://www.qstcorp.com/upload/pdf/202202/%EF%BC%88%E5%B7%B2%E4%BC%A0%EF%BC%89QMI8658C%20datasheet%20rev%200.9.pdf
+package qmi8656c
+
+import "tinygo.org/x/drivers"
+
+// Device wraps the I2C connection to the QMIC8658 sensor
+type Device struct {
+	bus        drivers.I2C
+	Address    uint16
+	AccLsbDiv  uint16
+	GyroLsbDiv uint16
+}
+
+type Config struct {
+	// SPI Config
+	SPIMode    byte // One of SPI_X_WIRE
+	SPIEndian  byte // One of SPI_XXX_ENDIAN
+	SPIAutoInc byte // One of SPI_NOT_AUTO_INC or SPI_AUTO_INC
+	// Accelerometer
+	AccEnable  byte // One of ACC_ENABLE or ACC_DISABLE
+	AccScale   byte // One of ACC_XG
+	AccRate    byte // One of ACC_XX_YYHZ
+	AccLowPass byte // One of ACC_LOW_PASS_X
+	// Gyro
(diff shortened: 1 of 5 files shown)


qmi8658c/qmi8658c.go
```diff
@@ -0,0 +1,186 @@
...
+		AccRate:     ACC_NORMAL_1000HZ,
+		AccLowPass:  ACC_LOW_PASS_2_62,
+		GyroEnable:  GYRO_FULL_ENABLE,
+		GyroScale:   GYRO_512DPS,
+		GyroRate:    GYRO_1000HZ,
+		GyroLowPass: GYRO_LOW_PASS_2_62,
+	}
+}
+
+// Check if the usar has defined a desired configuration, if not uses the
```
conejoninja: usar* typo

qmi8658c/qmi8658c.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+	}
+	x = raw[0] * 1000 / int32(d.GyroLsbDiv)
+	y = raw[1] * 1000 / int32(d.GyroLsbDiv)
+	z = raw[2] * 1000 / int32(d.GyroLsbDiv)
+	return x, y, z
+}
+
+// Read the temperature from the sensor, the values returned are in Celsius.
+func (d *Device) ReadTemperature() (t float32) {
```
conejoninja: Please, follow the signature  `ReadTemperature() (int32, error) {` as other drivers do. It should return millidegrees:

https://github.com/tinygo-org/drivers/blob/release/bme280/bme280.go#L116


As a rule of thumb, avoid float when possible.

conejoninja (changes requested):
This number needs to be updated too.
I think somewhere in the process lost count, with your driver it should be 89 devices on the list, but please count them again just in case I'm wrong.

qmi8658c/qmi8658c.go
rcbadiale: Fixed the typo, thanks for noticing.

qmi8658c/qmi8658c.go
rcbadiale: Changed the signature as proposed and updated the example file.

My main language is Python and I'm new to Go and TinyGo, so I have to ask, this would be for memory optimization? Since float32 would use a lot more memory than int32?

rcbadiale:
Apparently they are 90 devices, if I can count correctly.

All changes were made, hope it's alright now.

Thank you very much for the review.

conejoninja:
Thanks for your contribution, it's a very well written driver.
Now merged

qmi8658c/qmi8658c.go
conejoninja: I don't know the details, but yes, the boards targeted don't have much memory and don't play well with floats in general.

PR #471, opened 2022-10-10
apds9960: fix compile by removing unused import

ajanata:
The apds9960 package does not compile because it has an unused import. This change simply removes that unused import.


deadprogram:
Thank you for the fix @ajanata now merging!

PR #474, opened 2022-10-19
Added si5351 driver

ghost:
Driver for Si5351 clock module

### DIFF
--- a/si5351/registers.go
+++ b/si5351/registers.go
@@ -1,0 +1,64 @@
+package si5351
+
+// The I2C address which this device listens to.
+const AddressDefault = 0x60     // Assumes ADDR pin is low
+const AddressAlternative = 0x61 // Assumes ADDR pin is high
+
+const (
+	OUTPUT_ENABLE_CONTROL = 3
+
+	CLK0_CONTROL = 16
+	CLK1_CONTROL = 17
+	CLK2_CONTROL = 18
+	CLK3_CONTROL = 19
+	CLK4_CONTROL = 20
+	CLK5_CONTROL = 21
+	CLK6_CONTROL = 22
+	CLK7_CONTROL = 23
+
+	MULTISYNTH0_PARAMETERS_1 = 42
+	MULTISYNTH0_PARAMETERS_3 = 44
+	MULTISYNTH1_PARAMETERS_1 = 50
+	MULTISYNTH1_PARAMETERS_3 = 52
+	MULTISYNTH2_PARAMETERS_1 = 58
+	MULTISYNTH2_PARAMETERS_3 = 60
+
+	SPREAD_SPECTRUM_PARAMETERS = 149
+
+	PLL_RESET = 177
+
+	CRYSTAL_INTERNAL_LOAD_CAPACITANCE = 183
+)
+
+const (
+	CRYSTAL_LOAD_6PF  = (1 << 6)
+	CRYSTAL_LOAD_8PF  = (2 << 6)
+	CRYSTAL_LOAD_10PF = (3 << 6)
+)
+
+const (
+	CRYSTAL_FREQ_25MHZ = 25000000
+	CRYSTAL_FREQ_27MHZ = 27000000
+)
+
+const (
+	PLL_A = iota
+	PLL_B
+)
+
+const (
+	R_DIV_1   = 0
+	R_DIV_2   = 1
+	R_DIV_4   = 2
+	R_DIV_8   = 3
+	R_DIV_16  = 4
+	R_DIV_32  = 5
+	R_DIV_64  = 6
+	R_DIV_128 = 7
+)
+
+const (
+	MULTISYNTH_DIV_4 = 4
+	MULTISYNTH_DIV_6 = 6
+	MULTISYNTH_DIV_8 = 8
+)
--- a/si5351/registers.go
+++ b/si5351/registers.go
@@ -47,14 +47,14 @@ const (
 )
 
 const (
-	R_DIV_1   = 0
-	R_DIV_2   = 1
-	R_DIV_4   = 2
-	R_DIV_8   = 3
-	R_DIV_16  = 4
-	R_DIV_32  = 5
-	R_DIV_64  = 6
-	R_DIV_128 = 7
+	R_DIV_1 = iota
+	R_DIV_2
+	R_DIV_4
+	R_DIV_8
+	R_DIV_16
+	R_DIV_32
+	R_DIV_64
+	R_DIV_128
 )
 
 const (
--- a/si5351/si5351.go
+++ b/si5351/si5351.go
@@ -1,0 +1,439 @@
+package si5351
+
+import (
+	"errors"
+	"math"
+
+	"tinygo.org/x/drivers"
+)
+
+// Device wraps an I2C connection to a SI5351 device.
+type Device struct {
+	bus     drivers.I2C
+	Address uint8
+
+	buf            [8]byte
+	initialised    bool
+	crystalFreq    uint32
+	crystalLoad    uint8
+	pllaConfigured bool
+	pllaFreq       uint32
+	pllbConfigured bool
+	pllbFreq       uint32
+	lastRdivValue  [3]uint8
+}
+
+var errNotInitialised = errors.New("Si5351 not initialised")
+var errInvalidParameter = errors.New("Si5351 invalid parameter")
+
+// New creates a new SI5351 connection. The I2C bus must already be configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func New(bus drivers.I2C) Device {
+	return Device{
+		bus:         bus,
+		Address:     AddressDefault,
+		crystalFreq: CRYSTAL_FREQ_25MHZ,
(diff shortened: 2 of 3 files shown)


deadprogram:
Hello @conotto thank you for working on this driver.

I have changed the branch to `dev` as described here: https://github.com/tinygo-org/drivers/blob/release/CONTRIBUTING.md#how-to-use-our-github-repository

ghost:
Please let me know if anything is missing

si5351/registers.go
```diff
@@ -0,0 +1,64 @@
...
+
+const (
+	R_DIV_1   = 0
+	R_DIV_2   = 1
+	R_DIV_4   = 2
+	R_DIV_8   = 3
+	R_DIV_16  = 4
+	R_DIV_32  = 5
+	R_DIV_64  = 6
+	R_DIV_128 = 7
```
thunderstorm99:
```suggestion
	R_DIV_1 = iota
	R_DIV_2
	R_DIV_4
	R_DIV_8
	R_DIV_16
	R_DIV_32
	R_DIV_64
	R_DIV_128
```
This could be written like so as well, right?

si5351/registers.go
```diff
@@ -0,0 +1,64 @@
...
+	OUTPUT_ENABLE_CONTROL = 3
+
+	CLK0_CONTROL = 16
+	CLK1_CONTROL = 17
+	CLK2_CONTROL = 18
+	CLK3_CONTROL = 19
+	CLK4_CONTROL = 20
+	CLK5_CONTROL = 21
+	CLK6_CONTROL = 22
+	CLK7_CONTROL = 23
```
thunderstorm99, line 17:
```suggestion
	CLK0_CONTROL = iota + 16
	CLK1_CONTROL
	CLK2_CONTROL
	CLK3_CONTROL
	CLK4_CONTROL
	CLK5_CONTROL
	CLK6_CONTROL
	CLK7_CONTROL
```
This could also be re-written using iota

ghost: Should't that be iota + 16 ?

thunderstorm99: Yes, you are 100% right, that was a typo, sorry!

ghost:
Is there anything else required ? Or can this be merged already ?

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+	crystalFreq    uint32
+	crystalLoad    uint8
+	pllaConfigured bool
+	pllaFreq       uint32
+	pllbConfigured bool
+	pllbFreq       uint32
+	lastRdivValue  [3]uint8
+}
+
+var errNotInitialised = errors.New("Si5351 not initialised")
```
rminnich, line 26: I believe that nowadays, the practice is to make these exported:
var Err...

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+		bus:         bus,
+		Address:     AddressDefault,
+		crystalFreq: CRYSTAL_FREQ_25MHZ,
+		crystalLoad: CRYSTAL_LOAD_10PF,
+	}
+}
+
+// Configure sets up the device for communication
+// TODO error handling
+func (d *Device) Configure() {
```
rminnich, line 43: I'm surprised that Configure does not have an error return

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+
+	data := d.buf[:1]
+
+	// Disable all outputs setting CLKx_DIS high
+	data[0] = 0xFF
+	d.bus.WriteRegister(d.Address, OUTPUT_ENABLE_CONTROL, data)
```
rminnich, line 49: Can this get an error? I'm guessing yes?

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+	// Disable spread spectrum output.
+	d.DisableSpreadSpectrum()
+
+	d.initialised = true
+
+}
+
+// Connected returns whether a device at SI5351 address has been found.
+func (d *Device) Connected() bool {
+	err := d.bus.Tx(uint16(d.Address), []byte{}, []byte{0})
```
rminnich, line 72: Can Tx get an error that is not just about things not being found? It seems a bit dangerous to throw the error away, or is that the interface definition? I"m still catching up.

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+	return err == nil
+}
+
+func (d *Device) EnableSpreadSpectrum() (err error) {
+	data := d.buf[:1]
+	err = d.bus.ReadRegister(d.Address, SPREAD_SPECTRUM_PARAMETERS, data)
```
rminnich, line 78: The more common practice, at least at google, would be an unnamed error return and then
if err := d.bus.ReadRegister(...); err != nil {
return err
}
...
return d.bus.WriteRegister
it avoids any possible problems around shadowing (which has caused trouble in practice)

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+	if !(denom > 0) {
+		return errInvalidParameter
+	}
+	// 20-bit limit
+	if !(num <= 0xFFFFF) {
+		return errInvalidParameter
+	}
+	// 20-bit limit
+	if !(denom <= 0xFFFFF) {
+		return errInvalidParameter
```
rminnich, line 165: Here is a case where returning an unexported variable can case trouble: a caller might wish to do something like
if errors.Is(err, si5351.ErrInvalidParametr)
also, it's common practice nowadays  to wrap errors with more information. You have three ways at least to return an invalid parameter here and that can be very frustrating to debug. 
At the very least:
return fmt.Errorf("%#x is < 0xfffff: %w", denom, ErrInvalidParameter)

you may have to use %v of do something else if tinygo is not quite caught up, but this is a more useful error return pattern to use.

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+	data[4] = uint8(p1 & 0x000000FF)
+	data[5] = uint8(((p3 & 0x000F0000) >> 12) | ((p2 & 0x000F0000) >> 16))
+	data[6] = uint8((p2 & 0x0000FF00) >> 8)
+	data[7] = uint8(p2 & 0x000000FF)
+	d.bus.WriteRegister(d.Address, baseaddr, data)
+
+	// Reset both PLLs
+	data = d.buf[:1]
+	data[0] = (1 << 7) | (1 << 5)
+	d.bus.WriteRegister(d.Address, PLL_RESET, data)
```
rminnich, line 218: can this error? Do you want to return an error if it does?

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+//	used, but this isn't currently implemented in the driver.
+func (d *Device) ConfigureMultisynth(output uint8, pll uint8, div uint32, num uint32, denom uint32) (err error) {
+
+	// Basic validation
+	if !d.initialised {
+		return errNotInitialised
+	}
+	// Channel range
+	if !(output < 3) {
+		return errInvalidParameter
```
rminnich, line 294: here's another case where more informative errors might be useful, viz:
return fmt.Errorf("output %d is < 3:%w", ErrInvalidParameter)

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+		p2 = uint32(128*float64(num) - float64(denom)*math.Floor(128*(float64(num)/float64(denom))))
+		p3 = denom
+	}
+
+	// Get the appropriate starting point for the PLL registers
+	baseaddr := uint8(0)
+	switch output {
+	case 0:
+		baseaddr = MULTISYNTH0_PARAMETERS_1
+		break
```
rminnich, line 362: you don't need these breaks. That's only an issue in C.

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+	data[0] = uint8((p3 & 0xFF00) >> 8)
+	data[1] = uint8(p3 & 0xFF)
+	data[2] = uint8(((p1 & 0x30000) >> 16)) | d.lastRdivValue[output]
+	data[3] = uint8((p1 & 0xFF00) >> 8)
+	data[4] = uint8(p1 & 0xFF)
+	data[5] = uint8(((p3 & 0xF0000) >> 12) | ((p2 & 0xF0000) >> 16))
+	data[6] = uint8((p2 & 0xFF00) >> 8)
+	data[7] = uint8(p2 & 0xFF)
+	err = d.bus.WriteRegister(d.Address, baseaddr, data)
+	if err != nil {
```
rminnich, line 382: again, strongly recommend the practice:
if err := d.bus.Write(...); err != nil {
return nil
}

avoid named return parameters is usual.

si5351/si5351.go
```diff
@@ -0,0 +1,443 @@
...
+	}
+	if num == 0 {
+		clkControlReg |= (1 << 6) // Integer mode
+	}
+
+	var register uint8
+	switch output {
+	case 0:
+		register = CLK0_CONTROL
+		break
```
rminnich, line 400: don't need the breaks.

btw, golangci-lint would likely make useful suggestions.

rminnich (commented):
interesting to see this driver, thanks!

Issue #475, opened 2022-10-23
lsm303agr: accelerometer not working as expected with microbit V1.5 and V2.2


breml:
Today I tried the lsm303agr driver from latest release (tinygo.org/x/drivers v0.23.0) and the accelerometer did not provide the expected values. I used the the code from the example as a foundation.

After some debugging, I found the code in https://github.com/tinygo-org/drivers/issues/125#issuecomment-647614939, which worked for me as expected. To me, it looks like the difference is on line https://github.com/tinygo-org/drivers/blob/release/lsm303agr/lsm303agr.go#L140 where the current driver has

```golang
ReadRegister(uint8(d.AccelAddress), ACCEL_OUT_X_L_A, data)
```

and the working code from the before mentioned issue comment uses:

```golang
ReadRegister(LSM303_ADDRESS_ACCEL, LSM303_REGISTER_ACCEL_OUT_X_L_A | 0x80, data)
```

The difference is the `| 0x80`. I had a look at the datasheet (https://www.st.com/resource/en/datasheet/lsm303agr.pdf), but I was not able to figure out, what the effect of setting the most significant bit with `| 0x80` is.

I made an exact copy of the lsm303agr driver in my project and only added `| 0x80`, which fixed the accelerometer for me on both, microbit V1.5 and V2.2.

Additionally, I also don't know, if it is neccessary to set this bit for the reading of the magnetometer and the temperature is necessary as well.

breml:
polite CC: @ysoldak, @alankrantas, @conejoninja

ysoldak:
Hm, interesting. Seem to be my bad, a typo in a [refactoring PR](https://github.com/tinygo-org/drivers/pull/368).

Could you please try and checkout either "dev" branch or "v0.23.0" tag and replace `ACCEL_OUT_X_L_A` with `ACCEL_OUT_X_H_A` on [line 140](https://github.com/tinygo-org/drivers/blob/v0.23.0/lsm303agr/lsm303agr.go#L140)?

Same for magnetometer, line 193: `MAG_OUT_X_L_M ` -> `MAG_OUT_X_H_M `
And temperature, line 223: `OUT_TEMP_L_A ` -> `OUT_TEMP_H_A `

I don't have IMU myself, can't test.

breml:
@ysoldak I quickly tested what you suggested for the accelerometer (line 140) and this does not work.

[...]

PR #478, opened 2022-10-28
lsm303agr: fix I2C address auto increment for multi data read

breml:
fixes: #475

### DIFF
--- a/lsm303agr/lsm303agr.go
+++ b/lsm303agr/lsm303agr.go
@@ -137,7 +137,7 @@ func (d *Device) Configure(cfg Configuration) (err error) {
 // -1000000.
 func (d *Device) ReadAcceleration() (x, y, z int32, err error) {
 	data := d.buf[:6]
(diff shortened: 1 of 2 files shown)


breml:
CC: @ysoldak, @alankrantas

breml:
I confirm, that my test application, which uses the accelerometer works on both, micro:bit v1.5 and V2.2. I didn't test the other two sensors though, because I don't have a meaningful test app available.

alankrantas:
Can confirm that micro:bit v1.5 and v2.0 works (with the official example, which use all three sensors) after applying the patch.

breml:
@deadprogram This PR has been tested successfully on microbit V1.5, V2.0 and V2.2 (see above).
For the reasoning and the details also see #475.

I think, this PR is ready to be merged.

deadprogram:
Thank you very much for the fix @breml and to @alankrantas for testing. Now merging.

PR #479, opened 2022-11-02
vl53l1x: Add getter for the effective SPAD count

baw-michael:
Implement function to get the effective SPAD count.

Signed-off-by: Michael Meister <michael.meister@bytesatwork.ch>


deadprogram:
Thank you very much for the improvement @baw-michael now merging.

PR #480, opened 2022-11-08
wifinina: add support for http server

scottfeldman:
This adds http.ListenAndServe() implementation for wifinina.  Included is an example webserver, basically copied from the rtl8720dn/webserver.

Tested on Arduino Nano RP2040 Connect.  Also tried testing on Arduino Nano33 IoT but test panics with "out of memory".  There's only 32K of SRAM on the Nano33, vs 264K on the rp2040 Connect.

### DIFF
--- a/examples/wifinina/webserver/main.go
+++ b/examples/wifinina/webserver/main.go
@@ -1,0 +1,209 @@
+package main
+
+import (
+	"fmt"
+	"machine"
+	"strconv"
+	"time"
+
+	"tinygo.org/x/drivers/net/http"
+	"tinygo.org/x/drivers/wifinina"
+)
+
+// You can override the settings with the init() in another source code:
+//
+// func init() {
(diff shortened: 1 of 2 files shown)


sago35:
I will check this weekend.

sago35 (approved):
LGTM

Source code verified.
I think it is very good.

I also tried running it on PyPortal.
I was able to confirm that it is working correctly.

deadprogram:
Thank you for the addition @scottfeldman and to @sago35 for review. Now going to squash/merge.

fangyygw:
out of memory in  Arduino-Nano33-iot

scottfeldman:
@fangyygw Yes, arduino-nano33 doesn't have enough SRAM (32KB) to support http server.  Alternate boards that can use wifinina for http server are: nano-rp2040, pyportal, metro-m4-airlift, and matrixportal-m4.

PR #482, opened 2022-11-17
Fix odd if sequence in the example ntpclient

dgryski:


deadprogram:
It was an even number of if now it's odd? Thanks for the fix @dgryski now merging.

PR #485, opened 2022-11-22
fix broken link for lora sx1261

inindev:
updated link for Semtech SX126x Lora
added spaces to line-up the "Interface Type" markdown for the QMI8658C accelerometer and the Waveshare GC9A01 TFT round display


deadprogram:
Thanks @inindev now merging.

PR #486, opened 2022-11-26
buzzer: tone duration float64, no tone during rest

breml:


aykevl:
Can you describe what bugs this fixes?

breml:
Yes sure. The first commit makes all the note durations `float64`, such that these constants can be used with `Tone(hz, duration float64) (err error)`, which expects the duration to be a float.

The second commit makes sure, that the note [`Rest`](https://github.com/tinygo-org/drivers/pull/486/files#diff-8e7ff364f2c1bf2330f3874e17264e5e5ce6f363b3960994e500f788b88cd896R12) works as intended. This is, `Rest` is a pause and and during the pause, there is no sound. The only effect of a `Rest` is to have a defined time of silence. Without the fix, the `Rest` does not have any effect, because the duration is a function of the tone (hz) and if this is `0`, then everything becomes `0` and the function returns more or less imediatly.

deadprogram:
Thanks for the fix here @breml now merging.

PR #488, opened 2022-12-09
Update DHT22 receiveSignal to use runtime/interrupt

### LINKED ISSUE #483: DHT 22 not working with ESP32
Hey folks, 
so I was trying to hook up a DHT22 to my ESP32 Dev Module with the following basic code:

```
package main

import (
   "machine"
   "tinygo.org/x/drivers/dht"
)

func main() {
   sensor := dht.New(machine.SDA_PIN, dht.DHT22)
   _ = sensor
}
```

whatever I did calling `dht.New(machine.SDA_PIN, dht.DHT22)` caused tinygo to fail flashing with the error> `machine.UART1.Interrupt undefined (type *machine.UART has no field or method Interrupt)`. 
Is this a bug, since I've been told that ESP support in tinygo is not quite perfect yet, or is it an issue on my end?

jasdel:
Fixes #483 by updating the DHT22 device to use `runtime/interrupt` instead of `machine.UART1` for interrupt disabling while receiving signals from the device.

### DIFF
--- a/dht/thermometer.go
+++ b/dht/thermometer.go
@@ -11,6 +11,7 @@ package dht // import "tinygo.org/x/drivers/dht"
 
 import (
 	"machine"
+	"runtime/interrupt"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks for the correction @jasdel that makes a lot more sense. Now merging.

PR #492, opened 2022-12-20
build: remove older format build tags

deadprogram:
This PR removes all of the build tags in the older format in similar fashion to https://github.com/tinygo-org/tinygo/pull/3348


bgould:
Onward to the future!  🚀 thank you @deadprogram

PR #496, opened 2022-12-27
net/http: add PostForm()

sago35:
This PR adds `net/http.PostForm()`.
Copy of go 1.19.4.


deadprogram:
Thank you for the enhancement @sago35 now merging.

PR #497, opened 2022-12-27
rtl8720dn: add handling when connect fails

sago35:
This PR adds handling when rtl8720dn fails to connect to a specific TCP destination.

-1 represents many rpc failures, but details are unknown.
In fact, rtl8720dn returns -1 when a connection fails.

### DIFF
--- a/rtl8720dn/netdriver.go
+++ b/rtl8720dn/netdriver.go
@@ -71,9 +71,12 @@ func (d *Driver) ConnectTCPSocket(addr, port string) error {
 	name[6] = byte(ipaddr[2])
 	name[7] = byte(ipaddr[3])
 
-	_, err = d.Rpc_lwip_connect(socket, name, uint32(len(name)))
+	result, err := d.Rpc_lwip_connect(socket, name, uint32(len(name)))
 	if err != nil {
 		return err
+	}
+	if result == -1 {
+		return fmt.Errorf("failed to connect to %d.%d.%d.%d port %s", addr[0], addr[1], addr[2], addr[3], port)
 	}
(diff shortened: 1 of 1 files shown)


rtl8720dn/netdriver.go
```diff
@@ -71,10 +71,13 @@ func (d *Driver) ConnectTCPSocket(addr, port string) error {
 	name[6] = byte(ipaddr[2])
 	name[7] = byte(ipaddr[3])
 
-	_, err = d.Rpc_lwip_connect(socket, name, uint32(len(name)))
+	result, err := d.Rpc_lwip_connect(socket, name, uint32(len(name)))
 	if err != nil {
 		return err
 	}
+	if result == -1 {
+		return fmt.Errorf("failed to connect to %d.%d.%d.%d port %s", addr[0], addr[1], addr[2], addr[3], port)
```
deadprogram, line 79: It would be great to have a function to return the `string` for an IP address without using `fmt` :smile_cat:

scottfeldman:
Note in drivers:dev branch, this is fixed as:

```
result := r.rpc_lwip_connect(int32(sock), name, uint32(len(name)))
if result == -1 {
        return fmt.Errorf("Connect to %s failed", ip)
}
```

deadprogram:
@sago35 is this PR still relevant?

sago35:
No, it's already been updated to the following code, so the issue is resolved. I'll close this PR.

https://github.com/tinygo-org/drivers/blob/v0.27.0/rtl8720dn/rtl8720dn.go#L501-L505

Issue #501, opened 2023-01-07
net/net.go IP.String() returns nonsense


vjanelle:
The String() function for IP just converts the raw bytes to string, and doesn't return an intelligent and human readable value.

deadprogram:
This has now been released, so closing. Thank you!

PR #502, opened 2023-01-07
(#501) make IP.String() method return something sensible

vjanelle:


deadprogram:
Hello @vjanelle thanks for pointing out this problem.

It would be better to solve it in a way that avoid using the `fmt` package, because that package makes heavy use of reflection which greatly increases the memory footprint.

Also, please make all PRs against the `dev` branch. Please see https://github.com/tinygo-org/drivers/blob/release/CONTRIBUTING.md#how-to-use-our-github-repository for more info.

Thank you!

vjanelle:
np!  I saw somewhere else this being done with just concat, I'll try that instead.

vjanelle:
@deadprogram reimplemented, added a test to cover the change

deadprogram:
Thank you very much @vjanelle now merging.

deadprogram:
I did not realize it at the time I reviewed this PR, but it has the unintentional side effect of breaking code such as the Drivers `net/mqtt` packages. The mismatch between the IP `wifinina` package and the `net` package result in invalid IP addresses being used during actual usage.

Since the upcoming refactor of `netdev` changes this for the actual correct Go implementation, I think it is better to just revert this PR instead of trying to fix it.

Sorry I did not do a more complete review/testing at the time @vjanelle and hope it does not inconvenience.

PR #504, opened 2023-01-09
examples: New lorawan basic-demo example

ofauchon:
"basic-demo" is an example application for connecting a Lorawan Network.
It's written to be as simple as possible so you can easily reuse it.
  
All the code shared between examples had been moved in a common package.


examples/lora/lorawan/basic-demo/main.go
```diff
@@ -0,0 +1,122 @@
...
+func parse(data []byte) error {
+	s := string(data)
+
+	if s == "CONNECT" {
+		lorawanDoConnect = true
+	}
+	return nil
+}
+
+// This will be run as a go routine to keep us connected
```
deadprogram: Perhaps I am reading this wrong, but looks like this function would retry until it connects once, but then once connected would not reconnect. Perhaps the comment just needs to explain this.

deadprogram:
I think that perhaps a more bare bones rx-type example might be more useful than another console style application. Something that joins, then creates some periodic sensor readings maybe?

ofauchon:
> I think that perhaps a more bare bones rx-type example might be more useful than another console style application. Something that joins, then creates some periodic sensor readings maybe?

Yes, you're right. 
Let's strip it down

ofauchon:
There was something wrong with DevNounce encoding/endianness.
Fixed in this branch, but the code needs review ...
 
Anyways, Join and Uplinks are working with this PR.
Please give a try . Thx

examples/lora/lorawan/basic-demo/main.go
```diff
@@ -0,0 +1,98 @@
...
+var (
+	radio   lora.Radio
+	session *lorawan.Session
+	otaa    *lorawan.Otaa
+
+	lorawanConnected bool
+)
+
+// loraConnect() will loop until we're connected to Lorawan network
+func loraConnect() {
```
deadprogram: Just wonder why we don't just write this function in a synchronous way, e.g. 

```go
func loraConnect() error {
    start := time.Now()
    var err error
    for time.Since(start) < 30*time.Second {
        println("Starting Lorawan Join sequence")
	err = lorawan.Join(otaa, session)
        if err == nil {
            println("loraConnect: Connected !")
            return nil
        }

        println("loraConnect: Join error:", err)
	println("loraConnect: retrying")
	time.Sleep(time.Second * LORAWAN_RECONNECT_DELAY_SEC)
    }

    err = errors.New("lorawan: unable to join")
    println(err.Error())
    return err
}
```

And then call it like this:

```go
if err := loraConnect(); err != nil {
    failMessage(err)
}
```

In other words, avoid the goroutine entirely.

What do you think?

ofauchon: Hi @deadprogram, 

The loraConnect() was initially design with goroutine to handle all the Lorawan connection lifecycle (machine state for join/accept/uplinks/downlinks, reconnections, retransmissions,  channel hopping, RX windows ... ). 

This way, the developper would only have to start the goroutine and interact with it through IN/OUT channel events.
(This takes advantage of goroutine/channel features, but I could be wrong ...)

That said, the current lorawan code only implements basic features ...
So we can remove the goroutine in this example.

This will make the code simple and easy to understand for newcomers and testers.

Next, I'll be happy to discuss about the correct implementation of the Lorawan machine state (goroutine or not),

I'll push the necessary changes in the next hours. 

Thanks

Issue #489, opened 2022-12-11
request for ds18b20 driver


kellyfitzy:
Hi,

I'm a novice programmer looking for a tinygo ds18b20 driver to use with Raspberry Pi Pico W.   Willing to test.  Have used various ds18b20 golang drivers on Raspberry Pi Nano, etc.  Thanks in advance!

dn-kolesnikov:
I'm done my library for 1-wire
In current moment I’m adding support for ds18b20
In few days I’m going to push it to my GitHub
Tested for RP pico

kellyfitzy:
thanks!

dn-kolesnikov:
> Hi,
> 
> I'm a novice programmer looking for a tinygo ds18b20 driver to use with Raspberry Pi Pico W. Willing to test. Have used various ds18b20 golang drivers on Raspberry Pi Nano, etc. Thanks in advance!

https://github.com/dn-kolesnikov/tinygo-drivers.git

first commit
so far, it only works with a single sensor on the bus.
no docs, no tests. :no_good:
the rest is in the process
see example...

kellyfitzy:
Thanks, I'll check it out.

matveynator:
> > Hi,
> > I'm a novice programmer looking for a tinygo ds18b20 driver to use with Raspberry Pi Pico W. Willing to test. Have used various ds18b20 golang drivers on Raspberry Pi Nano, etc. Thanks in advance!
> 
> https://github.com/dn-kolesnikov/tinygo-drivers.git
> 
> first commit so far, it only works with a single sensor on the bus. no docs, no tests. 🙅 the rest is in the process see example...

Could you pull to tinygo-drivers when ready and stable? Thank you very much!

conejoninja:
Hello @dn-kolesnikov , thanks for your work, as @matveynator said, please make a pull request to the repo so we can properly review it.

I do have some DS18B20 laying around (the waterproof version), and will gladly review/test it.


One thing I notice, it that the ReadTemperature looks like this 
```go
func (d Device) ReadTemperature() (Temperature, error) {
```
It would be great if it has the same signature as other temperature sensor in the repository: temperature should be int32, and it should be milliCelsius instead of Celsius. That makes any temp. sensor interchangeable, and you could replace one with another without much effort.

```go
func (d *Device) ReadTemperature() (temperature int32, err error) {
```

dn-kolesnikov:
[...]

PR #505, opened 2023-01-10
first implementation of 1-wire protocol

conejoninja:
I would like @dn-kolesnikov to review this PR as it was mainly his work, I just adapted it to the proper repo structure.

Once merge, issue #489 could be closed.


deadprogram:
Was just looking at this PR. So first of all seems to me that it would be better to structure the files like this:

`onewire/onewire.go`

And then separately

`ds18b20/ds18b20.go`

This seems more in keeping with the way things like `lora` and `net` are currently.

Future onewire devices (EEPROMs, fuel gauges, etc.) would be added at the root level, like pretty much all of the other drivers are now.

What do you think?

deadprogram:
Also, to be in fully correct naming it should be `1-wire` not `onewire`.

Due to Go naming rules this would turn into just `wire`, correct?

This seems appealing:

```go
package somedevice

import "tinygo.org/x/drivers/1-wire"

func (d *Driver) DoSomething() {
    wire.Reset()
    wire.Write(...)
}
```

What do you think?

conejoninja:
I just follow @dn-kolesnikov naming.

NOTE: There's a popular Arduino library called Wire that handles I²C communication, https://www.arduino.cc/reference/en/language/functions/communication/wire/  It might be confusing?

deadprogram:
The name of the package here would be `1-wire` In any case, I doubt there would be confusion over that particular name in the code.

Regarding the directory organization, I am not really in favor of the proposed structure. How do we proceed?

conejoninja:
the structure you proposed seems ok to me and make sense. I'll make the changes and uodate the PR

conejoninja:
I updated the package as suggested, renamed the folder to `1-wire` , and the package to `wire` (suggested by the IDE was `__wire`), then, we are forced to import it as 
```go
import (
	"machine"
	"time"
	wire "tinygo.org/x/drivers/1-wire"

	"tinygo.org/x/drivers/ds18b20"
)
``` 


It's ok for me.

dn-kolesnikov:
> Was just looking at this PR. So first of all seems to me that it would be better to structure the files like this:
> 
> 
> 
> `onewire/onewire.go`
> 
> 
> 
> And then separately
> 
> 
> 
> `ds18b20/ds18b20.go`
> 
> 
> 
> This seems more in keeping with the way things like `lora` and `net` are currently.
> 
> 
> 
> Future onewire devices (EEPROMs, fuel gauges, etc.) would be added at the root level, like pretty much all of the other drivers are now.
> 
> 
> 
> What do you think?

The Arduino standard library is called OneWire (https://www.arduino.cc/reference/en/libraries/onewire/).

deadprogram:
@dn-kolesnikov has a good point. We should probably name it `onewire`.

examples/ds18b20/main.go
```diff
@@ -0,0 +1,67 @@
...
+			continue
+		}
+		// temperature in celsius milli degrees (°C/1000)
+		println("Temperature (°C/1000): ", t)
+	}
+}
+
+// SliceToHexString converts a slice to Hex string
+// fmt.Printf - compile error on an Arduino Uno boards
+func SliceToHexString(rom []uint8) string {
```
deadprogram: Just out of curiosity, what was the reason to not use `encoding/hex`?

matveynator:
Good day ). As each device are using tiny bits of onewire protocol implementation (as I understood - small memory usage wise) it is good that devices are added root level. "onewire"  name also seems reasonable... I am also ready to help )

examples/ds18b20/main.go
dn-kolesnikov: Just forgot about it... image

dn-kolesnikov:
at the moment I am checking the function of searching for multiple devices on the bus. I had to change the Device structure and add a few additional functions. the next commit to my repository will break compatibility with the current versions of the library. the examples will also be rewritten taking into account the new changes.

conejoninja:
> at the moment I am checking the function of searching for multiple devices on the bus. I had to change the Device structure and add a few additional functions. the next commit to my repository will break compatibility with the current versions of the library. the examples will also be rewritten taking into account the new changes.

Feel free to make the PR directly from your repository. I did it because I needed for a FOSDEM talk, but I started to think it will not be merged on time, no need to rush.

deadprogram:
I was actually getting ready to merge this unless there is some reason I shouldn't?

conejoninja:
> I was actually getting ready to merge this unless there is some reason I shouldn't?

well, @dn-kolesnikov commented the next change will break the compatibility, so I guess it's better to wait

bgould:
@conejoninja Sounds like @dn-kolesnikov has some new version of this mentioned in #489 ... how should we move forward with this, do you want to merge the changes into your branch or just should @dn-kolesnikov start a new one?

dn-kolesnikov (approved):

dn-kolesnikov (approved):

onewire/onewire.go
```diff
@@ -0,0 +1,213 @@
...
+		return errNoPresence
+	}
+	return nil
+}
+
+// WriteBit transmits a bit to 1-Wire bus.
+func (d Device) WriteBit(data uint8) {
+	d.p.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	if data&1 == 1 { // Send '1'
+		time.Sleep(5 * time.Microsecond)
```
bxparks, line 63: Sorry for the late comment, but I wasn't around in January when this PR was started.

I just want to point out that `time.Sleep()` yields to the goroutine scheduler. There is no guarantee that this will sleep for 5 microseconds. It could sleep for 100s of microseconds, or several milliseconds.

Since TinyGo does not provide a non-yielding `runtime.delayMicros()` or `machine.delayMicros()` function, as far as I know, maybe there is no other option right now.

onewire/onewire.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return romIDs[:romIndex:romIndex], nil
+}
+
+// Crc8 compute a Dallas Semiconductor 8 bit CRC.
+func (d Device) Сrc8(buffer []uint8, size int) (crc uint8) {
+	// Dow-CRC using polynomial X^8 + X^5 + X^4 + X^0
+	// Tiny 2x16 entry CRC table created by Arjen Lentz
+	// See http://lentz.com.au/blog/calculating-crc-with-a-tiny-32-entry-lookup-table
+	crc8_table := [...]uint8{
```
bxparks, line 202: Just curious, is the TinyGo compiler smart enough to store this as a constant in flash, or will it create this on the stack? I guess it's only 32 bytes, so it doesn't matter except maybe on AVR processors. In my experience, a `const string` was the only way I could guarantee that an array like this would be stored in flash.

dn-kolesnikov: The smallest of the AVR ATtiny85 microcontrollers supported in Tinygo has 512 bytes of RAM. If 32 bytes are critical for your project, you can use a fast CRC calculation algorithm for avr in assembler. Or use
`
const crc8_table string = "\x00\x5E\xBC\xE2\x61\x3F\xDD\x83\xC2\x9C\x7E\x20\xA3\xFD\x1F\x41\x00\x9D\x23\xBE\x46\xDB\x65\xF8\x8C\x11\xAF\x32\xCA\x57\xE9\x74"
`

onewire/onewire.go
dn-kolesnikov: I have conducted several experiments for the chips I have (rp2040, stm32f103, stm32f411, attiny85, atmega328p). Created a project:
```
package main

import "time"

func main() {
	for {
		time.Sleep(5 * time.Microsecond)
	}
}
```
Got an assembler listing and studied them. In all variants, a similar tick counting mechanism was used to organize the delay.

bxparks: That's interesting.. Have you looked at the SAMD21? I've been struggling with it. `time.Sleep()` seems to have an overhead of 250-400 microseconds. The following program prints: `elapsed millis: 36621` instead of a number close to `500` milliseconds:

```go
package main

import "time"

func main() {
        time.Sleep(2 * time.Second) // wait for serial monitor

        println("Starting...")
        start := time.Now()

        for i := 0; i < 100_000; i++ {
                time.Sleep(5 * time.Microsecond)
        }

        elapsed := time.Since(start)
        println("elapsed millis:", elapsed.Milliseconds())
}
```

conejoninja:
I think the conversation is going a bit off-topic. We could merge this and make improvements later.

bxparks:
> I think the conversation is going a bit off-topic. We could merge this and make improvements later.

Sorry about that, I agree.

deadprogram:
OK based on the last comments I am squash/merging. Thank you very much everyone who worked on this!

PR #506, opened 2023-01-10
Lorawan: Support for uplink transmissions and new basic-demo example

ofauchon:
lora/lorawan/basic-demo is an example code for connecting to Lorawan networks and periodically send uplink messages.

The code is voluntarily basic so it's easy to understand and test.

This PR also provides : 

  - Fix in lora/lorawan driver to support uplink transmission.
  - New common package for all reusable code in lorawan examples 
  - A new LoraConfig() method in lora/radio.go interface to unify lora drivers setup

The code was tested on : 

- Boards: Lora-E5 and Nucleo WL55JC
- Gateway: RPI+DraginoHAT running Lora Packet Forwarder to Chirpstack Gateway


ofauchon:
I added a test for lorawan driver in Makefile too

deadprogram:
Thanks for the improvements @ofauchon now merging.

PR #511, opened 2023-01-18
ateccx08: support for ATECCx08 family of cryptographic processors

deadprogram:
This PR is to provide at least minimal support for the ATECCx08 family of cryptographic processors to create random numbers.

See https://datasheet.octopart.com/ATSAMA5D27-WLSOM1-Microchip-datasheet-149595509.pdf

### DIFF
--- a/ateccx08/ateccx08.go
+++ b/ateccx08/ateccx08.go
@@ -1,0 +1,258 @@
+// Package ateccx08 provides a driver for the ATECCx08 I2C cryptographic co-processor.
+//
+// Datasheet: https://datasheet.octopart.com/ATSAMA5D27-WLSOM1-Microchip-datasheet-149595509.pdf
+package ateccx08 // import "tinygo.org/x/drivers/ateccx08"
+
+import (
+	"errors"
+	"time"
+
+	"tinygo.org/x/drivers"
+)
+
+var (
+	maxCommandTime = (200 + 50) * time.Millisecond
+)
+
+var (
+	ErrWakeup          = errors.New("error on wakeup")
+	ErrInvalidCRCCheck = errors.New("invalid CRC check")
+	ErrLockFailed      = errors.New("locked failed")
+)
+
+type Device struct {
+	bus     drivers.I2C
+	Address uint8
+}
+
+// New returns ATECCx08 device for the provided I2C bus using default address.
+func New(i2c drivers.I2C) *Device {
+	return &Device{
+		bus:     i2c,
+		Address: Address,
+	}
+}
+
+// Configure the ATECCx08 device.
+func (d *Device) Configure() error {
+	return nil
+}
+
+// Connected returns whether ATECCx08 has been found.
+func (d *Device) Connected() bool {
+	if err := d.Wakeup(); err != nil {
+		return false
+	}
+
+	v, err := d.Version()
+	if err != nil {
+		return false
+	}
+
+	return (v == ATECC508 || v == ATECC608)
+}
+
+// Wakeup the ATECC by trying to write something to address 0x00
+func (d *Device) Wakeup() error {
+	d.bus.Tx(uint16(0x0), []byte{0x00}, nil)
+	time.Sleep(1500 * time.Microsecond)
+	d.bus.Tx(uint16(d.Address), []byte{0x00}, nil)
+	time.Sleep(maxCommandTime)
+
+	var status [4]byte
+	if err := d.readResponse(status[:]); err != nil {
+		return err
+	}
+
+	if status[0] != StatusAfterWake {
+		return ErrWakeup
+	}
+
+	return nil
+}
+
+// Sleep puts the ATECC to sleep.
+func (d *Device) Sleep() {
+	d.bus.Tx(uint16(d.Address), []byte{0x01}, nil)
+	time.Sleep(time.Millisecond)
+}
+
+// Idle puts the ATECC in idle mode.
+func (d *Device) Idle() {
+	d.bus.Tx(uint16(d.Address), []byte{0x02}, nil)
+	time.Sleep(time.Millisecond)
+}
+
+type ATECCVersion uint16
+
+func (at ATECCVersion) String() string {
+	switch at {
+	case ATECC508:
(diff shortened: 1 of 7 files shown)


deadprogram:
Now working to be able to generate random numbers. I have added an example that connects the standard Go `crypto/rand` package to use the ATECCx08 processor, works great on boards like the Arduino RP2040 Nano Connect, or Arduino Nano 33 IoT that have an onboard ATECC608.

ateccx08/ateccx08.go
```diff
@@ -0,0 +1,260 @@
...
+
+var (
+	ErrWakeup          = errors.New("error on wakeup")
+	ErrInvalidCRCCheck = errors.New("invalid CRC check")
+	ErrLockFailed      = errors.New("locked failed")
+)
+
+type Device struct {
+	bus     drivers.I2C
+	buf     []byte
```
dgryski: Could you make this a `[2]byte` and remove the allocation in `New()`?

Actually, where is `buf` even used?

deadprogram: That looks like copypasta on my part. Removed, thanks!

ateccx08/ateccx08.go
```diff
@@ -0,0 +1,258 @@
...
+		return false
+	}
+
+	return false
+}
+
+// Lock locks a zone in the device.
+// Note that you cannot unlock a device zone once locked,
+// so make sure you know what you are doing!
+func (d *Device) Lock(zone int) error {
```
ysoldak, line 194: Any link where to read about zone locking and what can go wrong? Covered in datasheet, perhaps?

ysoldak (approved):
Looks good and works for me.

ateccx08/ateccx08.go
ysoldak: > For the ATECC608A-TNGTLS, the Configuration zone has already been locked and the access policies of the Data zone have already been enforced. However, several of the data slots can still be updated through the use of other commands. If so desired, some of these slots can be permanently locked from future updates by using the Slot Locking mode of the Lock command.
> The Slot Lock mode of the Lock command is used to individually lock slots. Any slot that has the Lockable bit set and has not previously been slot-locked can be locked to prevent any further updates. This process is not reversible. Once a slot has been locked, it is permanently locked. In this mode of operation, the Summary CRC and Data fields are ignored.

examples/ateccx08/lock/main.go
```diff
@@ -0,0 +1,53 @@
...
+
+	println(version.String(), "started")
+
+	if !atecc.IsLocked() {
+		for i := 10; i > 0; i-- {
+			println(version.String(), "is not locked. Locking in", i, "seconds...")
+			time.Sleep(time.Second)
+		}
+
+		// locks the Configuration zone... PERMANENTLY!
```
aykevl, line 38: Perhaps it's a good idea to put this a bit more prominently at the top of the file? Examples don't usually make permanent changes to hardware.

examples/ateccx08/random/rand.go
```diff
@@ -0,0 +1,52 @@
...
+	"crypto/rand"
+	"errors"
+)
+
+var (
+	errNoATECC = errors.New("no ATECCx08")
+)
+
+func init() {
+	rand.Reader = &reader{}
```
aykevl, line 15: I'm all for putting this in the main package, but I think it would be great to create a `Reader` object in the driver itself, so you could do something like this:

```go
func main() {
    // ...
    rand.Reader = ateccx08.NewRand(atecc) // or a different name, NewRandReader?
    // ...
}
```

ateccx08/ateccx08.go
```diff
@@ -0,0 +1,258 @@
...
+
+import (
+	"errors"
+	"time"
+
+	"tinygo.org/x/drivers"
+)
+
+var (
+	maxCommandTime = (200 + 50) * time.Millisecond
```
aykevl, line 14: Perhaps it's a good idea to explain these constants? Do they come from the datasheet?

ateccx08/ateccx08.go
```diff
@@ -0,0 +1,258 @@
...
+	if err := d.Wakeup(); err != nil {
+		return false
+	}
+
+	v, err := d.Version()
+	if err != nil {
+		return false
+	}
+
+	return (v == ATECC508 || v == ATECC608)
```
aykevl, line 52: this is fine, but the following is slightly simpler:

```suggestion
	return v == ATECC508 || v == ATECC608
```

ateccx08/ateccx08.go
```diff
@@ -0,0 +1,258 @@
...
+	time.Sleep(maxCommandTime)
+	if err := d.readResponse(version[:]); err != nil {
+		return ATECCNone, err
+	}
+
+	return ATECCVersion(uint16(version[2])<<8 | uint16(version[3])&0xf000), nil
+}
+
+// Random returns an array of 32 byte-sized random numbers.
+func (d *Device) Random() ([32]byte, error) {
```
aykevl, line 119: I'm thinking whether this can be made more generic. I kind of like returning the values as-is (as a `[32]byte`, probably returned on the stack), but if we have another driver that provides a `Random` function it might provide a different number of bytes. An alternative would be to fill a byte slice instead (like in `rand.Reader`).

We can of course leave this as-is for now, but if we want to change this in the future that would be a backwards incompatible change.

aykevl (commented):
Looks good to me, some suggestions below.

examples/ateccx08/lock/main.go
ysoldak: That's true. We could probably even have a boolean constant set to `false` ("apply"? or "dryRun=true") by default and require it to be changed for irreversible modifications to be actually applied.

deadprogram:
After @waj334 made me aware of https://github.com/waj334/tinygo-cryptoauthlib I will close this PR in favor of putting the work into that repo instead.

ysoldak:
Shall probably add that repo to Awesome TinyGo then, for better visibility
https://github.com/tinygo-org/awesome-tinygo

PR #514, opened 2023-01-25
lorawan examples: Add LGT92 sx127x configuration

ofauchon:
Dragino LGT92 board embedds a SX127x lorawan module.
It uses a specific wiring that differs from featherbed's - pybadge.
So It has its own configuration file.

Of course, we could merge parts of common/sx127x.go(featherwing/pybadge) and common/sx127x-lgt92.go.
But I'm not sure it makes sense right now, as lorawan is still work in progress and it would make the code even harder
to understand.


ofauchon:
@deadprogram  : Should we rename "common/sx127x.go" to "common/sx127x-pybadge.go" so the filename is more explicit (like common/sx27x-lgt92.go) ?

deadprogram:
If we want to name the files this way, probably better to use `sx127x_lgt92.go` to follow the Go convention.

deadprogram:
I think you can close this PR now, @ofauchon

ofauchon:
LGT92 SX127x declaration was provided by another PR  #515

PR #515, opened 2023-01-26
sx127x: add RadioController interface to match sx126x

deadprogram:
This PR adds a `RadioController` interface and implementation to the `sx127x` to provide a similar pattern as the `sx126x`.


ofauchon (approved):
The code is working fine.
I could join and send uplinks with LGT92 device.

deadprogram:
Thank you for testing @ofauchon now merging!

PR #518, opened 2023-02-01
sx127x: Fix bug in SetTxPower

ofauchon:
- Some functions like SetTxPower were not updating loraConf object (needed in Rx and Tx functions)
- Reordering of some parts of the code for better understanding

This fix was successfully tested on LGT92 (sx127x) with Orange and TTN networks


deadprogram:
Thank you for the fix @ofauchon now merging.

PR #519, opened 2023-02-01
sx12xx: add README

deadprogram:
This PR adds very basic READMEs for the sx126x and 127x LoRa radios. It also updates the main README so that both are listed as being supported.

### DIFF
--- a/sx126x/README.md
+++ b/sx126x/README.md
@@ -1,0 +1,27 @@
+# SX126x LoRa Radio
+
+The Semtech SX126x family of sub-Ghz radio transceivers come in a number of different formats.
+
+SX126x radios have a wide continuous frequency coverage from 150 MHz to 960 MHz.
+
+SX1261 can transmit up to +15 dBm and the SX1262 can transmit up to +22 dBm
+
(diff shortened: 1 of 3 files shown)


sx126x/README.md
```diff
@@ -0,0 +1,23 @@
+# SX126x LoRa Radio
+
+The Semtech SX126x family of sub-Ghz radio transceivers come in a number of different formats.
+
+They are specific to a particular frequency range, so for example you need a different variation for EU868 vs. for AU915.
```
ofauchon: I was wrong with this statement..... 

-> sx126x radios  have a wide continuous frequency
coverage from 150 MHz to 960 MHz.

-> SX1261 can transmit up
to +15 dBm and the SX1262 can transmit up to +22 dBm

But.... some developments boards with SX126x may contain RF Filters to match only certain frequencies range.

ex: 
NUCLEO-WL55JC1 : High-frequency band. RF frequency
range from 865 to 928 MHz
NUCLEO-WL55JC2 : Low-frequency band. RF frequency
range from 433 to 510 MHz

deadprogram: Thanks for the corrections @ofauchon I've incorporated them into this PR.

PR #520, opened 2023-02-02
Add support for ams AS560x on-axis magnetic rotary position sensors (#510)

### LINKED ISSUE #510: Add a driver for ams AS5600
Add driver support for the ams [AS5600](https://ams.com/as5600) & [AS5601](https://ams.com/as5601) magnetic angle measurement/rotary position sensors.

These are nice contactless sensors with a decent resolution (12-bit) and not too expensive. Can be useful as an alternative to potentiometers, rotary encoders etc where you need the full 360 degree range.

They can work via I2C and/or analog/PWM.

neildavis:
Resolves #510. I think it's time to get some eyes on this :)

* Support for both AS5600 & AS5601 since they're very similar
* Only I2C interface supported. I guess Analog/PWM are too trivial to require driver support?

Issues I'd appreciate opinions on:

1. How many 'convenience' methods (wrappers around lower-level register read/write methods) should we include in the API? I've included only what I think are likely to be commonly used and/or where they can 'add value' (e.g. angle unit conversions, decoding of `STATUS` for magnet position/strength, `BURN` count limit logic)
All device functionality is available via the lower-level register access methods, but should we wrap them all?
2. ~~An open `TODO:` in the `Angle(units AngleUnit)` method when the value is requested in degrees and `MPOS`/`MANG` have been set. See the comment in the method. Maybe just drop conversion to degrees from the driver and let the caller do it?~~

### DIFF
--- a/as560x/as5600.go
+++ b/as560x/as5600.go
@@ -1,0 +1,172 @@
+//	Product: https://ams.com/as5600
+//	Datasheet: https://ams.com/documents/20143/36005/AS5600_DS000365_5-00.pdf
+
+package as560x // import tinygo.org/x/drivers/ams560x
+
+import (
+	"time"
+
+	"tinygo.org/x/drivers"
+)
+
+// AS5600 includes MPOS & MANG in addition to ZPOS to set a 'narrower angle range'
+// ZPOS enables setting the 'zero position' of the device to any RAW_ANGLE value.
+// MPOS ('max position') & MANG 'max angle' enable a 'partial range' on the AS5600.
+// The value in ANGLE is scaled & adjusted by the device according to ZPOS and MPOS/MANG.
+// The entire 12-bit range is 'compressed' into the RAW_ANGLE range of ZPOS->MPOS
+// (or ZPOS->ZPOS+MANG) thus enabling a higher resolution for a partial range.
+// if ZPOS > MPOS (or ZPOS + MANG > 4095) i.e. the incremental range 'crosses zero'
+// then the device will automatically compensate for the correct range.
+// For RAW_ANGLE values outside of the partial range, ANGLE will be 'capped' at either
+// 0 or 4095, depending on 'which end of the partial range is closer.'
+
+// AS5600Device represents an ams AS5600 device driver accessed over I2C
+type AS5600Device struct {
+	// promote BaseDevice
+	BaseDevice
+}
+
+// NewAS5600 creates a new AS5600Device given an I2C bus
+func NewAS5600(bus drivers.I2C) AS5600Device {
+	// Create base device
+	baseDev := newBaseDevice(bus)
+	// Add AS5600 specific registers
+	baseDev.registers[MPOS] = newI2CRegister(MPOS, 0, 0xfff, 2, reg_read|reg_write|reg_program)
+	baseDev.registers[MANG] = newI2CRegister(MANG, 0, 0xfff, 2, reg_read|reg_write|reg_program)
+	// Add AS5600 specific 'virtual registers'
+	conf, ok := baseDev.registers[CONF]
+	if ok {
+		baseDev.registers[PWMF] = newVirtualRegister(conf, 6, 0b11)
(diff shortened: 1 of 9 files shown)


neildavis:
Marking as 'draft' for now as tackling the second issue from my opening comment may lead to a major refactor. Feedback is still welcome though if anyone has the time and interest!

neildavis:
I think this is ready for review now. Thanks to @sago35 for pointing me at the TinyGo `dev` branch which resolved my [USB serial debugging issues](https://github.com/tinygo-org/tinygo/issues/3428) on my Tiny2040, I was able to further investigate the AS5600's behavior around the interaction of setting ZPOS/MPOS/MANG. Whilst this logic is not particularly clear from the datasheet, and introduces a little more complexity than we'd like, I think it's best to handle this in the driver where I can document it rather than every client program struggling to deal with with it!

as560x/as5600.go
```diff
@@ -0,0 +1,169 @@
...
+
+// calculateEffectiveMaxAngle calculates d.maxAngle after one of ZPOS/MPOS/MANG have been written
+func (d *AS5600Device) calculateEffectiveMaxAngle(register uint8, value uint16) error {
+	if MANG == register {
+		d.maxAngle = value // The easy case
+		return nil
+	}
+	var zpos, mpos uint16 = 0, 0
+	var err error = nil
+	if ZPOS == register {
```
deadprogram: To make the flow of control easier to follow, consider replacing this with:

```go
switch register {
    case MANG:
        d.maxAngle = value // The easy case
	return nil
    case ZPOS:
        zpos = value
        mpos, err = d.ReadRegister(MPOS)
    case MPOS:
        mpos = value
        zpos, err = d.ReadRegister(ZPOS)
    default:
        panic("calculateEffectiveMaxAngle() can only work from ZPOS, MPOS or MANG")
}
```

as560x/as5600.go
```diff
@@ -0,0 +1,169 @@
...
+// WriteRegister writes the given value for the given register to the AS560x device via I2C
+func (d *AS5600Device) WriteRegister(address uint8, value uint16) error {
+	// Call the BaseDevice method to do the actual write
+	if err := d.BaseDevice.WriteRegister(address, value); err != nil {
+		return err
+	}
+	// When either ZPOS/MANG/MPOS are set we need to recalculate maxAngle
+	// We also may need to invalidate some cached values for the other two registers
+	recalc := false
+	switch address {
```
deadprogram, line 120: Heh, you already did my above suggestion elsewhere in the code. Consider making them all consistent.

as560x/as560x.go
```diff
@@ -0,0 +1,193 @@
...
+	if BURN_ANGLE == burnCmd {
+		// BURN_ANGLE can only be executed up to 3 times.
+		// We can check this in advance by reading ZMCO before writing to the BURN register.
+		numBurns, err := d.ReadRegister(ZMCO)
+		if nil != err {
+			return err
+		}
+		if numBurns >= BURN_ANGLE_COUNT_MAX {
+			// We're outta BURNs :(
+			return errors.New("Max BURN_ANGLE limit reached")
```
deadprogram: Better to have this error defined as package var

```
var errMaxBURNAngle = errors.New("Max BURN_ANGLE limit reached")
```

And then in this function

```go
return errMaxBURNAngle
```

It avoids extra heap allocations.

deadprogram:
I made a few comments @neildavis but they are just coding style things. You obviously know a lot about these devices, thank you very much for working on this!

neildavis:
> I made a few comments @neildavis but they are just coding style things. You obviously know a lot about these devices, thank you very much for working on this!

@deadprogram Many thanks for the review, and in particular the hint on avoiding allocs when returning errors. Great spot!

I've made all the suggested changes and rebased against latest `dev` branch.

deadprogram:
The research on this device is pretty impressive @neildavis thanks for all the work! Now merging.

PR #526, opened 2023-02-26
initial rfm69 wip

TinHead:
Hello, 

This is my attempt of adding support for rfm69 based radios to TinyGo.
It is based on earlier work in the repos below as reference:
- https://github.com/charles-d-burton/rfm69-1 
- https://github.com/chbmuc/rfm69
- https://github.com/fulr/rfm69
 
At this point it's WIP but I would appreciate any feedback as I'm new to this and not very skilled in golang either :) 
The frequency is currently hardcoded to 433mHz as these are the only modules I have and test with, I think it should be easy to fix later on when tha basics are done.

So far I have tested initializing the module, working on testing actual radio connection.

Regards,
Razvan


TinHead:
realized I should have used dev as base will create new pr

PR #527, opened 2023-02-27
rfm69: Initial support

TinHead:
Hello,

This is my attempt of adding support for rfm69 based radios to TinyGo.
It is based on earlier work in the repos below as reference:

https://github.com/charles-d-burton/rfm69-1
https://github.com/chbmuc/rfm69
https://github.com/fulr/rfm69
At this point it's WIP but I would appreciate any feedback as I'm new to this and not very skilled in golang either :)
The frequency is currently hardcoded to 433mHz as these are the only modules I have and test with, I think it should be easy to fix later on when tha basics are done.

The example was tested on  RPI Pico boards, it should probably work  with other boards too. Arduino AVR boards are not supported as I could not find a nice way to setup the ISR for the DIO0 pin.

Regards,
Razvan


TinHead:
Closing due to no activity.

PR #528, opened 2023-02-28
Add textarea package

osechet:
Hi,
This PR adds a textarea package with a TextArea structure to simplify printing text on any _Displayer_. I'm not totally sure if it belongs in the drivers repo since it isn't really a driver.
It provides a simple font library and uses the _Displayer_'s `SetPixel()` method to print the characters as described by the font. For the moment only one 6x8 font is provided but more could be added.
An example is available in the `examples/textarea` directory, using a ili9341 driver. I only tested the code with that driver.


deadprogram:
@osechet have you seen https://github.com/tinygo-org/tinyfont by any chance?

aykevl:
Yeah, I also think this doesn't quite belong in the drivers repo. Either in tinyfont, or in a separate drawing/gfx package.

deadprogram:
Closing, please take a look at the feedback and decide how best to handle. Thanks @osechet

PR #529, opened 2023-03-02
ds3231: Fix negative temperature conversion

bxparks:
I also verified my interpretation of the DS3231 datasheet and my implementation of this formula by placing a DS3231 chip in the freezer section of my refrigerator for an hour or two, then reading the temperature. I got -16C, which is close enough to the -18C that the freezer was set to.


deadprogram:
Thank you for the fix @bxparks now merging.

PR #530, opened 2023-03-03
ds3231: Document incorrect leap year 2100

bxparks:
The current code interprets the 'century' flag as the year 2100. However the DS3231 hardware does not incorporate this flag in its leap year calculation, so will incorrectly consider the year 2100 as a leap year and increment from 2100-02-28 to 2100-02-29 instead of the correct 2100-03-01.

The 'century' bit is not useful for anything as far as I can tell. But instead of removing the code that uses the 'century' bit, I thought it would be less intrusive to just document the current behavior.

### DIFF
--- a/ds3231/ds3231.go
+++ b/ds3231/ds3231.go
@@ -73,7 +73,17 @@ func (d *Device) SetRunning(isRunning bool) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Fantastic job documenting this @bxparks thanks! Now merging.

PR #532, opened 2023-03-12
[gps] make the date available in addition to the time

pmcanseco:
I was using the GPS driver and found that while getting the time out of the library worked great, I had no way to get the date.  This PR references the [spec](https://docs.novatel.com/OEM7/Content/Logs/GPRMC.htm) present in the code comments for the RMC Sentence Type to grab the date and adds unit test coverage parsing for both time and date.

### DIFF
--- a/gps/gpsparser.go
+++ b/gps/gpsparser.go
@@ -179,7 +179,7 @@ func findSatellites(val string) (n int16) {
 	return 0
 }
 
-// findDate returns the speed from an RMC NMEA sentence.
+// findDate returns the date from an RMC NMEA sentence.
 func findDate(val string) time.Time {
 	if len(val) < 6 {
 		return time.Time{}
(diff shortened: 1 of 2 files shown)


gps/gpsparser.go
```diff
@@ -95,6 +95,8 @@ func (parser *Parser) Parse(sentence string) (Fix, error) {
 		fix.Longitude = findLongitude(fields[5], fields[6])
 		fix.Speed = findSpeed(fields[7])
 		fix.Heading = findHeading(fields[8])
+		date := findDate(fields[9])
```
pmcanseco, line 98: Out of the three supported sentence types (the cases in this `switch` block), only `RMC` contains the date. I was thinking of maybe adding in a type alias such as `type FixType string` and adding it as a field on the `Fix` struct to make it easier for consumers to see if they can meaningfully grab the date out of the `Fix` instance or not:

```go
if fix.Type == gps.RMC {
    month := fix.Time.Month()
}
```

If this sounds good, please let me know and I can add it.

deadprogram:
Thanks for the addition @pmcanseco now squashing/merging.

PR #535, opened 2023-03-22
Add support for E220-900T22S(JP) Lora module

akif999:
I added driver for E220-900T22S(JP) Lora module (UART interface).

* https://clealink.jp/top/products/iot/iot/E220-900T22S(JP)/

### DIFF
--- a/e220/config.go
+++ b/e220/config.go
@@ -1,0 +1,380 @@
+// Package e220 implements a driver for the e220 LoRa module.
+package e220
+
+import "fmt"
+
+// Config represents E220's configuration parameters
+type Config struct {
+	ModuleAddr         uint16
+	UartSerialPortRate uint8
+	AirDataRate        uint8
+	SubPacket          uint8
+	RssiAmbient        uint8
+	TransmitPower      uint8
+	Channel            uint8
+	RssiByte           uint8
+	TransmitMethod     uint8
+	WorCycleSetting    uint8
+	EncryptionKey      uint16
+	Version            uint8
+	errors             []error
+}
+
+func (c *Config) paramsToBytes(
+	bytes *[]byte,
+) error {
+	// byte [8] is read only
+	if len(*bytes) < E220RegisterLength-1 {
+		return fmt.Errorf("length of bytes must be greater than or equal to %d: got=%d", E220RegisterLength-1, (*bytes))
+	}
+	(*bytes)[0] = byte((c.ModuleAddr & 0xFF00) >> 8)
+	(*bytes)[1] = byte((c.ModuleAddr & 0x00FF) >> 0)
+	(*bytes)[2] = byte(((c.UartSerialPortRate & 0x07) << 5) | (c.AirDataRate & 0x1F))
+	reserved := byte(0b000)
+	(*bytes)[3] = byte(((c.SubPacket & 0x03) << 6) | ((c.RssiAmbient & 0x01) << 5) | ((reserved & 0x07) << 2) | (c.TransmitPower & 0x03))
+	(*bytes)[4] = byte(c.Channel)
+	reserved = byte(0b000)
+	(*bytes)[5] = byte(((c.RssiByte & 0x01) << 7) | ((c.TransmitMethod & 0x01) << 6) | ((reserved & 0x07) << 3) | (c.WorCycleSetting & 0x07))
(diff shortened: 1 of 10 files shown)


akif999:
@sago35 Can you do a review?

The result of the response to the indication that you have not yet confirmed is described below.
* https://github.com/akif999/drivers/pull/6#issuecomment-1432524192

deadprogram:
Hello @akif999 if you implement this interface, then this device should be able to work with the LoRaWAN package just like the other LoRa radios: https://github.com/tinygo-org/drivers/blob/dev/lora/radio.go

What do you think?

akif999:
Hello @deadprogram  I get it. I think it is desirable to follow the interface you presented. So I will update my implementation.

I'm thinking of updating the implementation and moving forward with this pull request.

deadprogram:
Sounds good @akif999 please let everyone here know how we can help!

e220/config.go
```diff
@@ -0,0 +1,380 @@
...
+	(*bytes)[0] = byte((c.ModuleAddr & 0xFF00) >> 8)
+	(*bytes)[1] = byte((c.ModuleAddr & 0x00FF) >> 0)
+	(*bytes)[2] = byte(((c.UartSerialPortRate & 0x07) << 5) | (c.AirDataRate & 0x1F))
+	reserved := byte(0b000)
+	(*bytes)[3] = byte(((c.SubPacket & 0x03) << 6) | ((c.RssiAmbient & 0x01) << 5) | ((reserved & 0x07) << 2) | (c.TransmitPower & 0x03))
+	(*bytes)[4] = byte(c.Channel)
+	reserved = byte(0b000)
+	(*bytes)[5] = byte(((c.RssiByte & 0x01) << 7) | ((c.TransmitMethod & 0x01) << 6) | ((reserved & 0x07) << 3) | (c.WorCycleSetting & 0x07))
+	(*bytes)[6] = byte((c.EncryptionKey & 0xFF00) >> 8)
+	(*bytes)[7] = byte((c.EncryptionKey & 0x00FF) >> 0)
```
soypat, line 39: No need to pass in a pointer to a slice. A slice is already composed of a pointer to the underlying array.  So you can change `(*bytes)[0] = byte(...)` to ``bytes[0] = byte(...)`  and `paramsToBytes(bytes *[]byte)` to `paramsToBytes(bytes []byte)` and you'll get same behaviour. 

Also one tends not to break function signatures with newlines in Go, especially if its a sufficiently small function signature.

e220/e220.go
```diff
@@ -0,0 +1,349 @@
...
+// targetUARTbaud represents the UART baud rate of the target board
+const (
+	TargetUARTBaud1200kbps   = 1200
+	TargetUARTBaud2400kbps   = 2400
+	TargetUARTBaud4800kbps   = 4800
+	TargetUARTBaud9600kbps   = 9600
+	TargetUARTBaud19200kbps  = 19200
+	TargetUARTBaud38400kbps  = 38400
+	TargetUARTBaud57600kbps  = 57600
+	TargetUARTBaud115200kbps = 115200
```
soypat, line 31: Does this add any value? It seems like a round about way to say `BaudRate1200` is equal to `1200`. MAybe this should be a `Baud` type instead and these be reduced to `Baud1200`.

e220/e220.go
```diff
@@ -0,0 +1,349 @@
...
+		d.m0.High()
+	case Mode2:
+		d.m1.High()
+		d.m0.Low()
+	case Mode3:
+		d.m1.High()
+		d.m0.High()
+	default:
+		return fmt.Errorf("Invalid mode: %d", mode)
+	}
```
soypat, line 125: this could be reduced to `d.m0.Set(mode&1 != 0); d.m1.Set((mode>>1)&1 != 0)` and a error check if mode is greater than 3

e220/e220.go
```diff
@@ -0,0 +1,349 @@
...
+	// cannot be received immediately after register access
+	time.Sleep(50 * time.Millisecond)
+
+	return d.payload[:readIndex], nil
+}
+
+// WriteConfig writes configuration values to E220
+func (d *Device) WriteConfig(cfg Config) error {
+
+	cfg.Validate()
```
soypat, line 224: Validate should return an `error` type itself. check out the `errors.Join(errSlice...)` function which does pretty much what you are doing below.

soypat (commented):
quick drive-by comments

PR #536, opened 2023-03-25
st7735: add DrawRGBBitmap8 method to draw raw RGB565 buffers

aykevl:
This is needed for high performance graphics. It's the same API as on the ili9341.
Using this API, and using partial updates to the screen (no full screen refreshes), I've been able to get well over 60fps of updates.


deadprogram:
Thanks for the addition @aykevl now merging.

PR #538, opened 2023-03-29
added DrawRGBBitmap8 (same as ili9341 & st7735)

conejoninja:


deadprogram:
Thanks for the addition, @conejoninja now merging.

PR #541, opened 2023-04-07
st7789: make it possible to configure gamma

aykevl:
Some displays need other values than the default to look right. Therefore, add a configuration option to set the gamma table for a particular LCD panel.


conejoninja:
thanks!! merging...

PR #543, opened 2023-04-12
ili9341: support tear-free display updates

aykevl:
Many displays don't have the TE pin exposed. But those that do have the pin (for example, the PyPortal) can use it to synchronize writing a new image to the display. When implemented correctly, tearing can be avoided entirely.

This commit also changes the LCD refresh direction to either top-to-bottom or left-to-right depending on the rotation. Previously it might refresh from right-to-left or bottom-to-top. This has little impact on code that doesn't use the TE line, but code that does now only needs to worry about two cases (top-to-bottom and left-to-right) instead of four.
(Unfortunately, it appears that the hardware doesn't support changing the major LCD refresh order so code that wants to do tear-free rendering still needs to care about these two cases).

### DIFF
--- a/ili9341/ili9341.go
+++ b/ili9341/ili9341.go
@@ -158,6 +158,18 @@ func (d *Device) Display() error {
 	return nil
 }
 
+// EnableTEOutput enables the TE ("tearing effect") line.
+// The TE line goes high when the screen is not currently being updated and can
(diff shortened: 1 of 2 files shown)


sago35:
I will check later.

deadprogram:
Verified that it still works as expected, and code seems good, so now merging. Thanks @aykevl

PR #545, opened 2023-04-13
st7789: avoid heap allocations after the driver is created

aykevl:
Heap allocations can result in significant and unpredictable slowdowns. For example, they can take over 10ms on the Gopher Badge, which results in visible tearing (when tearing is otherwise avoided).


conejoninja:
Nice! thanks again!

PR #548, opened 2023-04-16
all: add sleep mode to various SPI displays

aykevl:
The API looks like this and is based on the PCA9685 PWM chip:

```go
Sleep(sleepEnabled bool) error
```

It will set the LCD panel to a low power state (not displaying any image) but memory contents will be preserved.


conejoninja:
Thanks! Great work as always.

PR #549, opened 2023-04-17
st7789: support the chip select pin

aykevl:
It is needed to control the chip select pin when the st7789 display is wired together with some other SPI device on the same bus, for example if it shares the bus with SPI flash.

This required some refactoring of the code to correctly set the CS pin everywhere. Notably, this removes the public Command, Data, Tx, and Rx methods which poke into private details of the st7789 driver and are therefore best hidden in my opinion.

I've tried to avoid any behavioral changes that might lead to regressions, but considering the size of the diff there is always the possibility of a regression. I tested it on the Gopher Badge and on the PineTime.

### DIFF
--- a/st7789/st7789.go
+++ b/st7789/st7789.go
@@ -46,6 +46,7 @@ type Device struct {
 	batchLength     int32
 	isBGR           bool
 	vSyncLines      int16
+	cmdBuf          [1]byte
 	buf             [6]byte
 }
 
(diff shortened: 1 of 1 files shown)


deadprogram:
@aykevl please resolve merge conflict. Thanks!

aykevl:
Rebased! And also tested the change on my PineTime to be sure there wasn't a regression.

deadprogram:
Thanks @aykevl now merging.

PR #550, opened 2023-04-17
ili9341: st7735: st7789: unify rotation support

aykevl:
* Unify the Rotation type, so that SetRotation can be used in interfaces.
  * Add Rotation() method to these displays, so that the current rotation can be read.
  * Change SetRotation() so that it can return an error (if the given rotation isn't supported).

I believe this should be entirely backwards compatible.

The goal of this change is so that `Rotation` and `SetRotation` can be used in interfaces, so that code using those interfaces can get and set the screen rotation.

### DIFF
--- a/st7789/st7789.go
+++ b/st7789/st7789.go
@@ -17,7 +17,9 @@ import (
 )
 
 // Rotation controls the rotation used by the display.
-type Rotation uint8
+//
+// Deprecated: use drivers.Rotation instead.
+type Rotation = drivers.Rotation
 
 // FrameRate controls the frame rate used by the display.
 type FrameRate uint8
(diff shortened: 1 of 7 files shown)


st7789/st7789.go
```diff
@@ -373,35 +375,37 @@ func (d *Device) FillScreen(c color.RGBA) {
...
+	case drivers.Rotation180:
 		d.rowOffset = 0
 		d.columnOffset = 0
-		break
-	case 3:
+	case drivers.Rotation270:
 		madctl = MADCTL_MX | MADCTL_MV
 		d.rowOffset = 0
 		d.columnOffset = 0
-		break
```
aykevl, line 398: These `break` instructions aren't necessary in Go (it's the default behavior, unlike C/C++).

deadprogram (approved):
LGTM

deadprogram:
Verified, working as expected. Thanks @aykevl now merging.

PR #551, opened 2023-04-23
add Sandbox Electronics NDIR CO2 sensor driver

soypat:
I've only got it to work on a Raspberry Pi 3 and 4 using the periph library. Raspberry Pi Pico has an important i2c bug I will set in issue soon.

This is the sensor: https://sandboxelectronics.com/?product=100000ppm-mh-z16-ndir-co2-sensor-with-i2cuart-5v3-3v-interface-for-arduinoraspeberry-pi

### DIFF
--- a/examples/ndir/main_ndir.go
+++ b/examples/ndir/main_ndir.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/ndir"
+)
+
+var (
+	ndirBus = machine.I2C0
+)
+
+func main() {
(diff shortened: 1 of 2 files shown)


deadprogram:
@soypat does this work with the PR https://github.com/tinygo-org/tinygo/pull/3770 merged?

If so, can you please add a smoke test here https://github.com/tinygo-org/drivers/blob/dev/smoketest.sh

Also add to list of drivers here https://github.com/tinygo-org/tinygo-site/blob/dev/content/docs/reference/devices/_index.md

Thanks!

soypat:
See #580

PR #553, opened 2023-04-27
Skip Xtensa smoke tests

aykevl:
This can be useful, to quickly run the smoke tests without compiler errors because of missing Xtensa support.


deadprogram:
Good idea, thanks @aykevl

PR #555, opened 2023-04-28
docs: cleanup readme list

deadprogram:
This PR removes the full list of devices from the README, better to keep it on the tinygo.org site.

See https://github.com/tinygo-org/tinygo-site/pull/353


aykevl:
I agree having this on the website makes it more accessible, but I'm afraid it will be very easy for it to get out of sync. As long as this list is in the drivers repo, it is much easier to require an addition on each PR that adds a driver.

Issue #534, opened 2023-03-20
Request for MPU6050 with DMP


DiscreteTom:
Thanks for this cool lib!

It seems that the MPU6050 driver doesn't contain functionalities about DMP(Digital Motion Processor), is there a change to support this?

DiscreteTom:
Example: https://github.com/electroniccats/mpu6050/blob/master/examples/MPU6050_DMP6/MPU6050_DMP6.ino this works in Arduino IDE.

brucebotes:
See #733 for an implementation of an Arduino version
https://github.com/tinygo-org/drivers/issues/733

PR #556, opened 2023-04-29
Mpu6050 rewrite

### LINKED ISSUE #544: MPU6050 ReadRotation not working
When I call ReadRotation, MPU6050 will stuck.

Code:

```go
package main

import (
	"machine"
	"time"

	"tinygo.org/x/drivers/mpu6050"
)

func main() {
	// Initialize the I2C bus & mpu6050 sensor
	machine.I2C0.Configure(machine.I2CConfig{})
	sensor := mpu6050.New(machine.I2C0)
	sensor.Configure()
	sensor.SetFullScaleAccelRange(mpu6050.AFS_RANGE_2G)
	sensor.SetFullScaleGyroRange(mpu6050.FS_RANGE_250)

	for {
		ax, ay, az := sensor.ReadAcceleration()
		// rx, ry, rz := sensor.ReadRotation() // this is not working yet, mpu6050 will stuck

		println(ax, ay, az)

		time.Sleep(10 * time.Millisecond)
	}
}

```

But if I use Arduino, with lib in https://github.com/adafruit/Adafruit_MPU6050 , run the https://github.com/adafruit/Adafruit_MPU6050/blob/master/examples/basic_readings/basic_readings.ino example, it works fine.

soypat:
Resolves #544.

I have published the mpu6050 driver I've used independently for years now.

It has greatly increased functionality and is much faster and efficient than the existing implementation. I have broken API compatibility since the existing implementation was extremely limited. In doing so I've taken the liberty of implementing patterns from the RFC https://github.com/tinygo-org/drivers/issues/310.

#### New Features:
* Configurable acceleration range
* Configurable angular velocity range
* Extensible configuration type added for future non-breaking changes
* Efficient sensor reading using one-shot methodology
* Expose raw data for even more efficient algorithms for power users.

### DIFF
--- a/examples/mpu6050/main.go
+++ b/examples/mpu6050/main.go
@@ -13,10 +13,8 @@ func main() {
 
 	mpuDevice := mpu6050.New(machine.I2C0, mpu6050.DefaultAddress)
 
-	err := mpuDevice.Configure(mpu6050.Config{
-		AccelRange: mpu6050.ACCEL_RANGE_16,
-		GyroRange:  mpu6050.GYRO_RANGE_2000,
-	})
+	// Configure the device with default configuration.
+	err := mpuDevice.Configure(mpu6050.Config{})
 	if err != nil {
 		panic(err.Error())
 	}
--- a/examples/mpu6050/main.go
+++ b/examples/mpu6050/main.go
@@ -18,8 +18,17 @@ func main() {
 	})
 
 	for {
-		x, y, z := accel.ReadAcceleration()
-		println(x, y, z)
 		time.Sleep(time.Millisecond * 100)
+		err := mpuDevice.Update()
+		if err != nil {
+			println("error reading from mpu6050:", err.Error())
+			continue
+		}
+		print("acceleration: ")
+		println(mpuDevice.Acceleration())
+		print("angular velocity:")
+		println(mpuDevice.AngularVelocity())
+		print("temperature centigrade:")
+		println(mpuDevice.Temperature() / 1000)
 	}
 }
--- a/examples/mpu6050/main.go
+++ b/examples/mpu6050/main.go
@@ -11,8 +11,11 @@ import (
 func main() {
 	machine.I2C0.Configure(machine.I2CConfig{})
 
-	accel := mpu6050.New(machine.I2C0)
-	accel.Configure()
+	mpuDevice := mpu6050.New(machine.I2C0)
+	mpuDevice.Init(mpu6050.IMUConfig{
+		AccRange:  mpu6050.ACCEL_RANGE_16,
+		GyroRange: mpu6050.GYRO_RANGE_2000,
+	})
 
 	for {
 		x, y, z := accel.ReadAcceleration()
--- a/examples/mpu6050/main.go
+++ b/examples/mpu6050/main.go
@@ -11,12 +11,15 @@ import (
 func main() {
 	machine.I2C0.Configure(machine.I2CConfig{})
 
-	mpuDevice := mpu6050.New(machine.I2C0)
-	mpuDevice.Init(mpu6050.IMUConfig{
+	mpuDevice := mpu6050.New(machine.I2C0, mpu6050.DefaultAddress)
+
+	err := mpuDevice.Configure(mpu6050.Config{
 		AccRange:  mpu6050.ACCEL_RANGE_16,
 		GyroRange: mpu6050.GYRO_RANGE_2000,
 	})
-
+	if err != nil {
+		panic(err.Error())
+	}
 	for {
 		time.Sleep(time.Millisecond * 100)
 		err := mpuDevice.Update()
--- a/examples/mpu6050/main.go
+++ b/examples/mpu6050/main.go
@@ -14,8 +14,8 @@ func main() {
 	mpuDevice := mpu6050.New(machine.I2C0, mpu6050.DefaultAddress)
 
 	err := mpuDevice.Configure(mpu6050.Config{
-		AccRange:  mpu6050.ACCEL_RANGE_16,
-		GyroRange: mpu6050.GYRO_RANGE_2000,
+		AccelRange: mpu6050.ACCEL_RANGE_16,
+		GyroRange:  mpu6050.GYRO_RANGE_2000,
 	})
 	if err != nil {
 		panic(err.Error())
@@ -31,7 +31,7 @@ func main() {
 		println(mpuDevice.Acceleration())
 		print("angular velocity:")
 		println(mpuDevice.AngularVelocity())
-		print("temperature centigrade:")
+		print("temperature celsius:")
 		println(mpuDevice.Temperature() / 1000)
 	}
 }
--- a/mpu6050/mpu6050.go
+++ b/mpu6050/mpu6050.go
@@ -167,12 +167,13 @@ func (p *Device) setRangeGyro(gyroRange RangeGyro) (err error) {
 		p.gRange = 500
 	case RangeGyro1000:
 		p.gRange = 1000
-	case RangeGyro2000:
+	case RangeGyro2000, rangeGyroDefault:
+		gyroRange = RangeGyro2000
 		p.gRange = 2000
 	default:
 		return errInvalidRangeGyro
 	}
-	return p.writeMasked(_GYRO_CONFIG, _G_FS_SEL, uint8(gyroRange)<<_G_FS_SHIFT)
+	return p.writeMasked(_GYRO_CONFIG, _G_FS_SEL, uint8(gyroRange-1)<<_G_FS_SHIFT)
 }
 
 // setRangeAccel configures the full scale range of the accelerometer.
@@ -187,12 +188,13 @@ func (p *Device) setRangeAccel(accRange RangeAccel) (err error) {
 		p.aRange = 4
 	case RangeAccel8:
 		p.aRange = 8
-	case RangeAccel16:
+	case RangeAccel16, rangeAccelDefault:
+		accRange = RangeAccel16
 		p.aRange = 16
 	default:
 		return errInvalidRangeAccel
 	}
-	return p.writeMasked(_ACCEL_CONFIG, _AFS_SEL, uint8(accRange)<<_AFS_SHIFT)
+	return p.writeMasked(_ACCEL_CONFIG, _AFS_SEL, uint8(accRange-1)<<_AFS_SHIFT)
 }
 
 // Sleep sets the sleep bit on the power managment 1 field.
@@ -217,13 +219,4 @@ func b2u8(b bool) byte {
 		return 1
 	}
 	return 0
-}
-
-func DefaultConfig() Config {
-	return Config{
-		AccelRange:  RangeAccel16,
-		GyroRange:   RangeGyro2000,
-		sampleRatio: 0, // TODO add const values.
-		clkSel:      0,
-	}
 }
--- a/mpu6050/mpu6050.go
+++ b/mpu6050/mpu6050.go
@@ -3,13 +3,14 @@ package mpu6050
 import (
 	"encoding/binary"
 	"fmt"
+	"math"
 
 	"tinygo.org/x/drivers"
 )
 
 const DefaultAddress = 0x68
 
-type IMUConfig struct {
+type Config struct {
 	// Use ACCEL_RANGE_2 through ACCEL_RANGE_16.
 	AccRange byte
 	// Use GYRO_RANGE_250 through GYRO_RANGE_2000
@@ -18,8 +19,8 @@ type IMUConfig struct {
 	clkSel      byte
 }
 
-// Dev contains MPU board abstraction for usage
-type Dev struct {
+// Device contains MPU board abstraction for usage
+type Device struct {
 	conn    drivers.I2C
 	aRange  int32 //Gyroscope FSR acording to SetAccelRange input
 	gRange  int32 //Gyroscope FSR acording to SetGyroRange input
@@ -29,8 +30,8 @@ type Dev struct {
 
 // New instantiates and initializes a MPU6050 struct without writing/reading
 // i2c bus. Typical I2C MPU6050 address is 0x68.
-func New(bus drivers.I2C, addr uint16) *Dev {
-	p := &Dev{}
+func New(bus drivers.I2C, addr uint16) *Device {
+	p := &Device{}
 	p.address = uint8(addr)
 	p.conn = bus
 	return p
@@ -40,7 +41,7 @@ func New(bus drivers.I2C, addr uint16) *Dev {
 // MPU6050. It sets the range of both the accelerometer
 // and the gyroscope, the sample rate, the clock source
 // and wakes up the peripheral.
-func (p *Dev) Init(data IMUConfig) (err error) {
+func (p *Device) Configure(data Config) (err error) {
 
 	// setSleep
 	if err = p.SetSleep(false); err != nil {
@@ -55,19 +56,19 @@ func (p *Dev) Init(data IMUConfig) (err error) {
 		return fmt.Errorf("set sample: %w", err)
 	}
 	// setFullScaleGyroRange
-	if err = p.SetGyroRange(data.GyroRange); err != nil {
+	if err = p.SetRangeGyro(data.GyroRange); err != nil {
 		return fmt.Errorf("set gyro: %w", err)
 	}
 	// setFullScaleAccelRange
-	if err = p.SetAccelRange(data.AccRange); err != nil {
+	if err = p.SetRangeAccel(data.AccRange); err != nil {
 		return fmt.Errorf("set accelrange: %w", err)
 	}
 	return nil
 }
 
 // Update fetches the latest data from the MPU6050
-func (p *Dev) Update() (err error) {
-	if err = p.read(ACCEL_XOUT_H, p.data[:]); err != nil {
+func (p *Device) Update() (err error) {
+	if err = p.read(_ACCEL_XOUT_H, p.data[:]); err != nil {
 		return err
 	}
 	return nil
@@ -75,7 +76,7 @@ func (p *Dev) Update() (err error) {
 
 // Get treats the MPU6050 like an ADC and returns the raw reading of the channel.
 // Channels 0-2 are acceleration, 3-5 are gyroscope and 6 is the temperature.
-func (d *Dev) RawReading(channel int) int16 {
+func (d *Device) RawReading(channel int) int16 {
 	if channel > 6 {
 		// Bad value.
 		return 0
@@ -87,13 +88,13 @@ func convertWord(buf []byte) int16 {
 	return int16(binary.BigEndian.Uint16(buf))
 }
 
-func (d *Dev) ChannelLen() int { return 7 }
+func (d *Device) ChannelLen() int { return 7 }
 
 // Acceleration returns last read acceleration in µg (micro-gravity).
 // When one of the axes is pointing straight to Earth
 // and the sensor is not moving the returned value will be around 1000000 or
 // -1000000.
-func (d *Dev) Acceleration() (ax, ay, az int32) {
+func (d *Device) Acceleration() (ax, ay, az int32) {
 	const accelOffset = 0
 	ax = int32(convertWord(d.data[accelOffset+0:])) * 15625 / 512 * d.aRange
 	ay = int32(convertWord(d.data[accelOffset+2:])) * 15625 / 512 * d.aRange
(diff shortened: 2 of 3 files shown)


examples/mpu6050/main.go
```diff
@@ -11,12 +11,24 @@ import (
...
+	mpuDevice.Init(mpu6050.IMUConfig{
+		AccRange:  mpu6050.ACCEL_RANGE_16,
+		GyroRange: mpu6050.GYRO_RANGE_2000,
+	})
 
 	for {
-		x, y, z := accel.ReadAcceleration()
-		println(x, y, z)
 		time.Sleep(time.Millisecond * 100)
+		err := mpuDevice.Update()
```
aykevl, line 23: I believe `Update` should have a parameter with the sensor values it should update?

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,251 @@
...
+)
 
-import "tinygo.org/x/drivers"
+const DefaultAddress = 0x68
 
-// Device wraps an I2C connection to a MPU6050 device.
-type Device struct {
-	bus     drivers.I2C
-	Address uint16
+type IMUConfig struct {
```
aykevl: In most drivers this is simply called `Config`. I recommend using that instead.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,251 @@
...
+	// Use ACCEL_RANGE_2 through ACCEL_RANGE_16.
+	AccRange byte
```
aykevl: We might want to generalize this, as most accelerometer sensors have exactly this sensor range (2g, 4g, 8g, 16g). It would be nice to be able to use common types for this.

Not very important for this PR, but would perhaps be nice to fix in the future.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,251 @@
...
 }
 
-// New creates a new MPU6050 connection. The I2C bus must already be
-// configured.
-//
-// This function only creates the Device object, it does not touch the device.
-func New(bus drivers.I2C) Device {
-	return Device{bus, Address}
+// Dev contains MPU board abstraction for usage
+type Dev struct {
```
aykevl: Most drivers use `Device`, I recommend following that pattern.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,251 @@
...
+	p.address = uint8(addr)
+	p.conn = bus
+	return p
+}
+
+// Init configures the necessary registers for using the
+// MPU6050. It sets the range of both the accelerometer
+// and the gyroscope, the sample rate, the clock source
+// and wakes up the peripheral.
+func (p *Dev) Init(data IMUConfig) (err error) {
```
aykevl: This is called `Configure` on most drivers.

```suggestion
func (p *Dev) Configure(data IMUConfig) (err error) {
```

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,251 @@
...
+	}
+	return nil
+}
+
+// SetSleep sets the sleep bit on the power managment 1 field.
+// When the recieved bool is true, it sets the bit to 1 thus putting
+// the peripheral in sleep mode.
+// When false is recieved the bit is set to 0 and the peripheral wakes
+// up.
+func (p *Dev) SetSleep(sleep bool) (err error) {
```
aykevl: I've started using a plain `Sleep` call in https://github.com/tinygo-org/drivers/pull/548. I suggest using that, or change other drivers to use `SetSleep` instead for consistency.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,251 @@
...
+	}
+	return nil
 }
 
-// Configure sets up the device for communication.
-func (d Device) Configure() error {
-	return d.SetClockSource(CLOCK_INTERNAL)
+// Get treats the MPU6050 like an ADC and returns the raw reading of the channel.
+// Channels 0-2 are acceleration, 3-5 are gyroscope and 6 is the temperature.
+func (d *Dev) RawReading(channel int) int16 {
```
aykevl: Do we need to expose this?
I'd prefer not to expose it, as it is a detail of the specific sensor that is in use. I can't think of any case where you'd want to use the raw value.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,251 @@
...
-
-// SetFullScaleAccelRange allows the user to configure the scale range for the accelerometer.
-func (d Device) SetFullScaleAccelRange(rng uint8) error {
-	return d.bus.WriteRegister(uint8(d.Address), ACCEL_CONFIG, []uint8{rng})
+// you would get a value close to 6.3 radians (360 degrees).
+func (d *Dev) AngularVelocity() (gx, gy, gz int32) {
+	const angvelOffset = 8
+	gx = int32(convertWord(d.data[angvelOffset+0:])) * 2663 / 5000 * d.gRange
+	gy = int32(convertWord(d.data[angvelOffset+2:])) * 2663 / 5000 * d.gRange
+	gz = int32(convertWord(d.data[angvelOffset+4:])) * 2663 / 5000 * d.gRange
```
aykevl: This introduces a division that was not there previously. I recommend changing this to a power-of-two division (that is optimized to a bitshift) and adjusting the multiply to keep the same output.

Most simple microcontrollers (Cortex-M0 etc) do not include a division instruction and have to emulate it in software, which is slow.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,251 @@
...
+	return gx, gy, gz
+}
+
+// Temperature returns the temperature of the device in milli-centigrade.
+func (d *Dev) Temperature() (Celsius int32) {
+	const tempOffset = 6
+	return 1000*int32(convertWord(d.data[tempOffset:]))/340 + 37*1000
+}
```
aykevl: Same here, this introduces a new division (a division by 340 is not optimized away).

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,251 @@
...
+	return nil
+}
+
+func DefaultConfig() IMUConfig {
+	return IMUConfig{
+		AccRange:    ACCEL_RANGE_16,
+		GyroRange:   GYRO_RANGE_2000,
+		sampleRatio: 0, // TODO add const values.
+		clkSel:      0,
+	}
```
aykevl: Any reason not to use the zero value for these options as the default value?

mpu6050/registers.go
```diff
@@ -1,132 +1,65 @@
...
+	CONFIG       uint8 = 0x1A
+	GYRO_CONFIG  uint8 = 0x1B
+	ACCEL_CONFIG uint8 = 0x1C
+	FIFO_EN      uint8 = 0x23
+	I2C_MST_CTRL uint8 = 0x24
+	PWR_MGMT_1   uint8 = 0x6B
+	ACCEL_XOUT_H uint8 = 0x3B
+	TEMP_OUT_H   uint8 = 0x41
+	GYRO_XOUT_H  uint8 = 0x43
+)
```
aykevl: This is entirely optional, but:
I think we should change these constants to be unexported. There is no reason why a user of the package would need these, and if they do that's a leaky abstration.
Most drivers do currently export these values, but I think it would be a good idea to unexport these.

mpu6050/registers.go
```diff
@@ -1,132 +1,65 @@
...
-	FIFO_COUNTH     = 0x72 // FIFO count registers (high bits)
-	FIFO_COUNTL     = 0x73 // FIFO count registers (low bits)
-	FIFO_R_W        = 0x74 // FIFO read/write
-	WHO_AM_I        = 0x75 // Who am I
+// Accelerometer ranges for Init configuration
+const (
+	ACCEL_RANGE_2  byte = 0x00
+	ACCEL_RANGE_4  byte = 0x01
+	ACCEL_RANGE_8  byte = 0x02
+	ACCEL_RANGE_16 byte = 0x03
```
aykevl: I think it would be nice to be able to share these values across drivers, so that you could use `SetAccelRange` in an interface for any IMU. (This is a more extensive change, so perhaps it's easier to leave this to a future PR).

mpu6050/mpu6050.go
DiscreteTom: Actually I have a use case:
I want to use a weak microcontroller to only read sensor's raw data, and send to PC via USB without processing the raw data.

aykevl: Have you measured whether this helps performance? I doubt it will make a significant difference.

DiscreteTom: Not tested yet.

Another use case is that some MPU 6050 chip will enable DMP by default, and when I read acceleration data, it actually returns the computed position data. In that case, maybe those bytes shouldn't be further processed. I don't know whether this is a standard behaviour for MPU 6050.
For MPU 6050 with DMP I also opened an issue #534.

Well, I'm new to embeded programming, sorry if I wrote some stupid questions.

examples/mpu6050/main.go
soypat: Yes, we should probably solve #345 before merging

mpu6050/mpu6050.go
soypat: I've worked in a similar fashion the last years. Where I can I defer computation to a PC. I could also expose the actual `data` field of Device as `RawData` to get the same effect, probably would be even better for users wanting to squeeze out performance

mpu6050/mpu6050.go
soypat: I'm not sure I understand- I've added them here redundantly just so I can mark it with a TODO

mpu6050/mpu6050.go
soypat: PTAL to confirm the change is what you meant

mpu6050/mpu6050.go
soypat: I've applied changes to expose RawData field, which I believe is clearer for users and adds less noise to the API. It also explicitly defines that part of the API as a mpu6050 specific thing since we still do not have generic field APIs in Go, so no risk of locking in an API with this decision.

mpu6050/mpu6050.go
aykevl: Yes, that's what I mean! Didn't check the formula but it looks like this should be easy to optimize by a compiler.

mpu6050/mpu6050.go
aykevl: To be very honest, this sounds like premature optimization to me. So until someone gives me the numbers that show it helps in practice, I'd rather _not_ expose these raw values. So far, I've only seen the _assumption_ that it will be slow.

If you have a very specific use case, you can of course still read these values manually from the I2C bus.

examples/mpu6050/main.go
```diff
@@ -11,12 +11,27 @@ import (
 func main() {
 	machine.I2C0.Configure(machine.I2CConfig{})
 
-	accel := mpu6050.New(machine.I2C0)
-	accel.Configure()
+	mpuDevice := mpu6050.New(machine.I2C0, mpu6050.DefaultAddress)
```
deadprogram, line 14: In other drivers, this has been solved by exporting `Address` and then just setting it before calling `Configure`. That way we do not have to break the existing API and all of the many projects that use this driver. I have several such projects myself :smiley_cat: 

```go
mpuDevice := mpu6050.New(machine.I2C0)
mpuDevice.Address = 0x66 // overrides default

// now do everything else
```

aykevl: With Go modules, I'm not too worried about breakage. Yes, it will break once you do a `go get -u tinygo.org/x/drivers`, but before that it will keep working just fine (and if you're doing a `go get -u tinygo.org/x/drivers`, you might as well add the default address).

I'm fine either way.

examples/mpu6050/main.go
```diff
@@ -11,12 +11,27 @@ import (
...
 
+	err := mpuDevice.Configure(mpu6050.Config{
+		AccRange:  mpu6050.ACCEL_RANGE_16,
+		GyroRange: mpu6050.GYRO_RANGE_2000,
+	})
```
aykevl: ...and here we have another breaking change, that I think is actually really useful.

examples/mpu6050/main.go
aykevl: #345 has been merged, so this can be updated.

examples/mpu6050/main.go
```diff
@@ -11,12 +11,27 @@ import (
...
+		if err != nil {
+			println("error reading from mpu6050:", err.Error())
+			continue
+		}
+		print("acceleration: ")
+		println(mpuDevice.Acceleration())
+		print("angular velocity:")
+		println(mpuDevice.AngularVelocity())
+		print("temperature centigrade:")
+		println(mpuDevice.Temperature() / 1000)
```
aykevl: Normally this is called Celsius, not centigrade (centigrade is an older term for the same unit).

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
-// Package mpu6050 provides a driver for the MPU6050 accelerometer and gyroscope
-// made by InvenSense.
-//
-// Datasheets:
-// https://store.invensense.com/datasheets/invensense/MPU-6050_DataSheet_V3%204.pdf
-// https://www.invensense.com/wp-content/uploads/2015/02/MPU-6000-Register-Map1.pdf
-package mpu6050 // import "tinygo.org/x/drivers/mpu6050"
```
aykevl: Any reason to remove these comments?

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
+	clkSel      byte
+}
+
+// Device contains MPU board abstraction for usage
 type Device struct {
-	bus     drivers.I2C
-	Address uint16
+	conn   drivers.I2C
+	aRange int32 //Gyroscope FSR acording to SetAccelRange input
+	gRange int32 //Gyroscope FSR acording to SetGyroRange input
```
aykevl, line 44: FSR?

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
+	// RawData contains the accelerometer, gyroscope and temperature RawData read
+	// in the last call via the Update method.
+	RawData [14]byte
```
aykevl: I still don't agree with exposing these. They are implementation details of the driver. Exposing these would make the mpu6050 a leaky abstraction.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
+	p.conn = bus
+	return p
 }
 
-// Connected returns whether a MPU6050 has been found.
-// It does a "who am I" request and checks the response.
-func (d Device) Connected() bool {
-	data := []byte{0}
-	d.bus.ReadRegister(uint8(d.Address), WHO_AM_I, data)
-	return data[0] == 0x68
```
aykevl: Why did you remove this method?

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
+	if err = p.SetSampleRate(data.sampleRatio); err != nil {
+		return errors.New("set sampleratio: " + err.Error())
+	}
+	if err = p.SetRangeGyro(data.GyroRange); err != nil {
+		return errors.New("set gyrorange: " + err.Error())
+	}
+	if err = p.SetRangeAccel(data.AccRange); err != nil {
+		return errors.New("set accelrange: " + err.Error())
+	}
+	return nil
```
aykevl: 1. Are you intentionally defaulting to very low GyroRange and AccRange?
2. This creates a bunch of heap allocations in the error paths. I would suggest returning the error directly instead of wrapping it in `errors.New`.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
+	const accelOffset = 0
+	ax = int32(convertWord(d.RawData[accelOffset+0:])) * 15625 / 512 * d.aRange
+	ay = int32(convertWord(d.RawData[accelOffset+2:])) * 15625 / 512 * d.aRange
+	az = int32(convertWord(d.RawData[accelOffset+4:])) * 15625 / 512 * d.aRange
+	return ax, ay, az
 }
 
-// ReadRotation reads the current rotation from the device and returns it in
-// µ°/s (micro-degrees/sec). This means that if you were to do a complete
+// Rotations reads the current rotation from the device and returns it in
```
aykevl: `AngularVelocity`, not `Rotation`.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
+// is chosen, the same accelerometer sample will be output.
+func (p *Device) SetSampleRate(srDiv byte) (err error) {
+	// setSampleRate
+	var sr [1]byte
+	sr[0] = srDiv
+	if err = p.write8(_SMPRT_DIV, sr[0]); err != nil {
+		return err
+	}
+	return nil
+}
```
aykevl: This is a new API.
Why not use `time.Duration` as an input? Or a sample rate in Hz, as an uint32? (Similar to `machine.CPUFrequency()` for example). The parameter that you're using here looks like a mpu6050-specific value.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
+
+	if err = p.read(_PWR_MGMT_1, pwrMgt[:]); err != nil {
+		return err
+	}
+	pwrMgt[0] = (pwrMgt[0] & (^_CLK_SEL_MASK)) | clkSel // Escribo solo el campo de clk_sel
+	if err = p.write8(_PWR_MGMT_1, pwrMgt[0]); err != nil {
+		return err
+	}
+	return nil
+}
```
aykevl: Is this actually useful? Can you give me an example where this would be used in practice?

If not, I suggest removing it to declutter the API.

(Also, it contains a Spanish comment?)

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
+	if err = p.read(_GYRO_CONFIG, gConfig[:]); err != nil {
+		return err
+	}
+	gConfig[0] = (gConfig[0] & (^_G_FS_SEL)) | (gyroRange << _G_FS_SHIFT) // Escribo solo el campo de FS_sel
+
+	if err = p.write8(_GYRO_CONFIG, gConfig[0]); err != nil {
+		return err
+	}
+	return nil
+}
```
aykevl: Can you make `gyroRange` a type? Like is done in the LIS3DH driver:

https://github.com/tinygo-org/drivers/blob/820298bfe3e9f8192e6c7c1147b95d7eee02eb46/lis3dh/lis3dh.go#L62

https://github.com/tinygo-org/drivers/blob/820298bfe3e9f8192e6c7c1147b95d7eee02eb46/lis3dh/registers.go#L54-L61

This makes it possible to easily change the type at a later time, if needed.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
+
+// SetRangeAccel configures the full scale range of the accelerometer.
+// It has four possible values +- 2g, 4g, 8g, 16g.
+// The function takes values of accRange from 0-3 where 0 means the
+// lowest FSR (2g) and 3 is the highest FSR (16g)
+func (p *Device) SetRangeAccel(accRange byte) (err error) {
```
aykevl: Same here. Please don't use magic constants (0, 1, 2, 3) and instead instruct users to use named constants.

mpu6050/mpu6050.go
```diff
@@ -1,95 +1,232 @@
...
-func (d Device) SetFullScaleAccelRange(rng uint8) error {
-	return d.bus.WriteRegister(uint8(d.Address), ACCEL_CONFIG, []uint8{rng})
+func DefaultConfig() Config {
+	return Config{
+		AccRange:    ACCEL_RANGE_16,
+		GyroRange:   GYRO_RANGE_2000,
+		sampleRatio: 0, // TODO add const values.
+		clkSel:      0,
+	}
 }
```
aykevl: Instead of DefaultConfig, have you considered making the zero value the default config?
That way, users can just do `Configure(mpu6050.Config{GyroRange: ...})` and know that the acceleration range is also set to some reasonable default.

mpu6050/mpu6050.go
soypat: sorry, added!

examples/mpu6050/main.go
soypat: I'd leave this for last since I still don't have a good rebase methodology. As soon as everything else is reviewed and OK I'll try my hand at rebasing one more time

mpu6050/mpu6050.go
soypat: I have pushed changes that make the zero value useful in latest commit!

soypat:
Closing this PR in favor of #577

PR #560, opened 2023-04-30
wifinina: add generated strings, improved debugging system and messages

deadprogram:
This PR adds generated strings, improved debugging system and improved error messages to the wifinina driver.

Issue #386, opened 2022-02-23
Wifi issue: boards not consistent in whether reset is active high or low


slzatz:
When configuring the wifi on an Arduino Nano 33 IoT, the reset pin is active low; however, for the Arduino MKR 1010, the reset pin is active high.  (see [problem-with-nina...](https://forum.arduino.cc/t/problem-with-nina-w102-startup-on-mkr-1010/923644/14) for a somewhat confusing discussion.)  The result of this is that the Configure function in wifinina.go doesn't work for the MKR 1010.  Changing the function as below so it takes a parameter as to whether reset is active high or low does fix the problem -- not sure how you guys want to handle a change like this -- could create a second slightly renamed Configure function so it doesn't break anything using the current one.
```go
func (d *Device) Configure2(resetActiveHigh bool) {
   net.UseDriver(d)
   pinUseDevice(d)
   
   d.CS.Configure(machine.PinConfig{Mode: machine.PinOutput})
   d.ACK.Configure(machine.PinConfig{Mode: machine.PinInput})
   d.RESET.Configure(machine.PinConfig{Mode: machine.PinOutput})
   d.GPIO0.Configure(machine.PinConfig{Mode: machine.PinOutput})
   
   d.GPIO0.High()
   d.CS.High()
   d.RESET.Set(resetActiveHigh)
   time.Sleep(1 * time.Millisecond)
   d.RESET.Set(!resetActiveHigh)
   time.Sleep(1 * time.Millisecond)
     
   d.GPIO0.Low()
   d.GPIO0.Configure(machine.PinConfig{Mode: machine.PinInput})   
} 
```
If this makes sense to anyone happy to do a pull request.

Steve

racecarparts:
I have a MKR 1010, I added this to the end of `setup()` in [the connect example](https://github.com/tinygo-org/drivers/blob/dbff576c9ec1670c16e961624a43de936efdd045/examples/wifinina/connect/main.go#L27-L43) and got things working:

```go
adaptor.RESET.High()
time.Sleep(1 * time.Millisecond)
adaptor.RESET.Low()
```
result:

```go
func setup() {

	// Configure SPI for 8Mhz, Mode 0, MSB First
	spi.Configure(machine.SPIConfig{
		Frequency: 8 * 1e6,
		SDO:       machine.NINA_SDO,
		SDI:       machine.NINA_SDI,
		SCK:       machine.NINA_SCK,
	})

	adaptor = wifinina.New(spi,
[...]

PR #537, opened 2023-03-28
Netdev

scottfeldman:
(I'm closing the original PR https://github.com/tinygo-org/drivers/pull/523 and submitting this one as a replacement).

This PR adds a network device driver model called netdev. There is a companion PR for TinyGo to update the "net" package here https://github.com/tinygo-org/tinygo/pull/3614. This PR is for the netdev interface definitions, netdev drivers (wifinina, rtl8720dn, espat) and network examples.

See netdev documentation here: README-net.md.

Testing results are [here](https://docs.google.com/spreadsheets/d/e/2PACX-1vT0cCjBvwXf9HJf6aJV2Sw198F2ief02gmbMV0sQocKT4y4RpfKv3dh6Jyew8lQW64FouZ8GwA2yjxI/pubhtml?gid=1013173032&single=true).

PR #561, opened 2023-05-02
wifinina: connect improvements

deadprogram:
This PR adds a few improvements for connecting the WiFiNINA.

It corrects the timing used for the RESET pins when calling `Configure()` to match other libraries. 

It also adds a new `ResetIsHigh`  property to allow changing the behavior of the RESET pin for boards like the Arduino MKR 1010, which correct #386 

Lastly, it makes some changes to the various examples to make a few retry attempts when connecting to the WiFi access point, which appears to solve most of the issues when connecting especially right after flashing.

Note that this PR depends on #560 which needs to be merged first, and then this PR rebased.

### DIFF
--- a/examples/wifinina/webclient/main.go
+++ b/examples/wifinina/webclient/main.go
@@ -20,8 +20,8 @@ var (
 	pass string
 )
 
-// IP address of the server aka "hub". Replace with your own info.
-const server = "tinygo.org"
+// IP address of the "example.com" server. Replace with your own info.
+const server = "93.184.216.34"
 
 // these are the default pins for the Arduino Nano33 IoT.
 // change these to connect to a different UART or pins for the ESP8266/ESP32
@@ -63,6 +63,7 @@ func main() {
 	waitSerial()
 
 	connectToAP()
+	displayIP()
 
 	for {
 		readConnection()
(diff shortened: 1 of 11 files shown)


deadprogram:
Now that I have been testing #537 I am going to close this PR. The `ResetIsHigh` can be added in a separate PR once #537 makes it in.

deadprogram:
Reopening since #537 will not merge until TinyGo 0.29

deadprogram:
Rebased against `dev` with #560 included. Please review! :smile_cat:

scottfeldman:
I've picked up the `ResetIsHigh` changes from this PR and copied them to #537.

All the other timing changes have already been picked up in #537.

wifinina/wifinina.go
```diff
@@ -126,6 +126,12 @@ type Device struct {
 	ip    uint32
 	port  uint16
 	mu    sync.Mutex
+
+	// ResetIsHigh controls if the RESET signal to the processor
+	// should be High or Low (the default). Set this to true
+	// before calling Configure() for boards such as the Arduino MKR 1010,
+	// where the reset signal needs to go high instead of low.
+	ResetIsHigh bool
```
ysoldak, line 134: This does look to me as a hack/workaround.
Shall we not have `RESET_HIGH` and `RESET_LOW` pins then instead?
Alternatively `RESET_HIGH` and `RESET` for backwards compatibility.
Then users can choose which pin to assign and which leave as "nopin".
And logic in `Configure()` then decides which pin to use (can panic / error when both defined or none)

ysoldak: Another alternative: control behaviour of `RESET` pin with a build tag.
`Arduino MKR 1010` target could have, say, `wifinina_reset_high` build tag. Next, wifinina driver has an extra file with this tag here that sets an internal flag for such behaviour.

examples/wifinina/webclient/main.go
```diff
@@ -20,8 +20,8 @@ var (
 	pass string
 )
 
-// IP address of the server aka "hub". Replace with your own info.
-const server = "tinygo.org"
+// IP address of the "example.com" server. Replace with your own info.
+const server = "93.184.216.34"
```
ysoldak, line 24: Why this change? It was not working with domain name? Have to use IP here?

deadprogram: Correct. That was never domain name.

wifinina/wifinina.go
deadprogram: Both sound more complicated somehow.

ysoldak: Second alternative is more complicated to implement perhaps, but easier and transparent to use.
To me it feels the reset pin behaviour is board/hardware specific and weird to have to think about it in the application code.

The first approach, I like it less myself, just brainstormed how can we do it differently.

ysoldak: We have drivers that have different behaviour for different build tags already, see https://github.com/tinygo-org/drivers/tree/release/ws2812

deadprogram: Yes, but they do more than just swap high/low for a specific pin. Adding a build tag for that seems like overkill, IMO.

deadprogram: The change I made seemed to me like the "simplest thing that could possibly work" to address the swap of setting high/low.  Adding more code/logic just seemed like a lot extra for that.

deadprogram: > To me it feels the reset pin behaviour is board/hardware specific and weird to have to think about it in the application code.

It is a single line you need to add if you are using one of only a couple of specific boards. Hence why I tried to make the least intrusive change possible.

ysoldak (approved):
After discussions in slack and in the light of the big upcoming refactoring in this package (netdev), it is no point to dwell on minor API changes now.

[seal of approval]

deadprogram:
Merging to keep the progress/bug fixes moving along. Thanks everyone!

Issue #557, opened 2023-04-29
Extract real-time waits to own package


soypat:
I can't help but think maybe packages like [`i2csoft`](https://github.com/tinygo-org/drivers/blob/release/i2csoft/i2csoft_esp32.go) and others could benefit from having a package that implements real-time waits for several architectures instead of having each driver implement their own waits for every architecture.

Suggestion for API:

```go
func NewRealTimeWaiter(waitTime time.Duration) Waiter

func (w Waiter) Wait() {
    return w.wait() // architecure specific implementation.
}

type Waiter struct {
    nopcount int // Or alternatively, guard waiter data structure for every implementation
}
```

aykevl:
Sounds reasonable.
However, the current implementation is not very precise. I think it might be a good idea to implement this in assembly, like is done for the WS2812 driver to wait a very precise number of cycles without the guesstimates that are currently used (that might change as a result of compiler optimizations for example). As shown in the ws2812 driver, it is possible on most chips to wait a very specific number of cycles.

bxparks:
Maybe the following proposal goes against some design philosophy regarding TinyGo, but I would like to see a slightly different API, something that looks like this:

```go
// Return a monotonically increasing integer representing the number of microseconds
// from an arbitrary point in time. The counter wraps around to 0 upon overflow.
func micros() uint32

// Return a monotonically increasing integer representing the number of milliseconds
// from an arbitrary point in time. The counter wraps around to 0 upon overflow.
func millis() uint32
```

So I guess I'm assuming that a hardware timer is available that can interrupt the CPU and increment a counter quickly and efficiently.

We can build `delayMicros(n)` and `delayMillis(n)` on top of `micros()` and `millis()`, but not the other way around:

```go
func delayMicros(n uint32) {
  start := micros()
  while micros() - start < n {}
}

func delayMillis(n uint32) {
  start := millis()
  while millis() - start < n {}
}
```

Occasionally, my device driver needs to perform some operation while waiting for a short time interval (i.e. anywhere between 1 and 1000 microseconds) to pass. The following pseudo-code is possible with `micros()` but not with `delayMicros()`:

```go
var start uint32
for {
  elapsed := micros() - start
  if elapsed >= duration {
    // doThing1
    break
  }
  // doThing2 while waiting
}
```

[...]

Issue #100, opened 2019-12-02
compilation error when using tinygo.org/x/drivers/espat/mqtt


titimoby:
During GDG Europe Summit, a tinygo workshop was nicely handled by @deadprogram and @conejoninja (thanks a lot guys)

The last step involved MQTT but some environments couldn't compile the code.

The issue is here:
https://github.com/hybridgroup/gdg-2019/issues/6

The code involved here:
https://github.com/hybridgroup/gdg-2019/blob/master/sensor/arduino/step7/main.go

I post it here because I suspect it is something that espat/mqtt driver but feel free to close it if I'm wrong.

deadprogram:
Anyone else able to reproduce this one? I have not been able to myself yet.

deadprogram:
https://github.com/tinygo-org/tinygo/pull/787 should have fixed that, so I am now closing as that code was merged into TinyGo `dev` branch.

Thank you!

PR #562, opened 2023-05-03
i2csoft: count cycles for more accurate timing

aykevl:
This PR adds a new delay package that implements cycle-accurate timing, and changes i2csoft to use this package. This fixes https://github.com/tinygo-org/drivers/issues/557.

Here is the rest of the commit message with some interesting details:

---

This is implemented in inline assembly using machine.CPUFrequency() to
know how long a single CPU cycle takes. As long as it is called with a
constant duration, it should be fully inlined and all values can be
const-propagated resulting in very tight inline assembly.

For example, when I convert i2csoft to use this delay function, the
entire delay function compiles to something like this:

    8784:   movs    r6, #100
    8786:   mov     r0, r6
    8788:   nop
    878a:   nop
    878c:   nop
    878e:   nop
    8790:   nop
    8792:   subs    r0, #1
    8794:   bne     0x8788

That means that all the math to calculate the number of cycles is
entirely optimized away (in this case, to 100 loops).

I ran the example on a few boards to see how well it works:

| board                 | 100ms wait | CPU core
|-----------------------|------------|------
| microbit              | 121.6ms    | Cortex-M0 so it has 12% overhead
| circuitplay-express   | 100.1ms    | Cortex-M0+ so it is cycle accurate
| pico                  | 100.2ms    | Cortex-M0+
| pyportal              | 100.3ms    | Cortex-M4
| circuitplay-bluefruit | 125.8ms    | Cortex-M4
| esp8266               | 125.1ms    |

This shows that there is some loop overhead because of conservative
estimates, but note that even though there may be a 25% overhead, the
actual overhead per `delay.Sleep()` call is very small. It should be
good enough for software I2C at least, and can potentially be improved
in the future.


i2csoft/i2csoft.go
```diff
@@ -278,3 +280,8 @@ func (i2c *I2C) WriteRegister(address uint8, register uint8, data []byte) error
 func (i2c *I2C) ReadRegister(address uint8, register uint8, data []byte) error {
 	return i2c.Tx(uint16(address), []byte{register}, data)
 }
+
+// wait waits for half the time of the SCL operation interval.
+func (i2c *I2C) wait() {
+	delay.Sleep(50_000) // half of a 100kHz cycle (50µs)
```
soypat: Why not `50 * time.Microsecond` as an argument?

aykevl: Yes you are right, I've fixed it.

deadprogram:
Great work @aykevl and @soypat for working thru this here and in #557 now merging.

PR #563, opened 2023-05-04
ci: run smoke tests in parallel

aykevl:
This speeds up the smoke tests by over 5x on my laptop, but keeps two features that I find very important:

 1. A stable output, which is used by tools like sizediff.
 2. An easy to change text file with all the smoke tests.

This means that parsing is a little bit harder, but with the help of packages like shlex and flag this is actually not very difficult.

(This should also help https://github.com/tinygo-org/tinygo/pull/3687, which is currently very slow).


deadprogram:
It cuts the time to do the tests in half, so yes! Now merging, thanks for improving this @aykevl

Issue #564, opened 2023-05-07
Driver for Adafruit Seesaw & Neotrellis


trichner:
As far as I can see there is currently no support for the Adafruit Seesaw chip which powers a lot of Adafruit boards. Among them the NeoTrellis 4x4 keypad with RGB background lights.

I'd like to add partial support for the Adafruit Seesaw, namely what is used by NeoTrellis as well as NeoTrellis support.

Reference:
- [Neotrellis](https://learn.adafruit.com/adafruit-neotrellis/overview)
- [adafruit-seesaw-atsamd09-breakout.pdf](https://github.com/tinygo-org/drivers/files/11414864/adafruit-seesaw-atsamd09-breakout.pdf)

According to the contribution guidelines I'm supposed to open an issue first :)

Draft implementations:
- https://github.com/trichner/gonybox/tree/main/pkg/seesaw
- https://github.com/trichner/gonybox/tree/main/pkg/neotrellis

Is there already support that I missed?

conejoninja:
Hello!

As far as I know, there's no work on the seesaw yet, so feel free to continue and update this issue once the PR is ready.

Thanks!

trichner:
Thanks for the quick reply, I'll open a PR here as soon as it's 'presentable'.

PR #565, opened 2023-05-07
Seesaw Driver

trichner:
As advertised in https://github.com/tinygo-org/drivers/issues/564

Most of it is refactored and goifyed from the offical Arduino driver: https://github.com/adafruit/Adafruit_Seesaw/blob/master/Adafruit_seesaw.cpp

PR #566, opened 2023-05-07
Seesaw Neopixel Support

trichner:
As advertised in https://github.com/tinygo-org/drivers/issues/564

Most of it is refactored and goifyed from the offical Arduino driver: https://github.com/adafruit/Adafruit_Seesaw/blob/master/Adafruit_seesaw.cpp

This is support for the seesaw device as well as its usage with NeoPixels.

### DIFF
--- a/seesaw/neopixel/neopixel.go
+++ b/seesaw/neopixel/neopixel.go
@@ -1,0 +1,181 @@
+package neopixel
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+	"tinygo.org/x/drivers/seesaw"
+)
+
+// seesawWriteDelay the seesaw is quite timing sensitive and times out if not given enough time,
+// this is an empirically determined delay that seems to have good results
+const seesawWriteDelay = time.Millisecond * 10
+
+// RGBW represents the RGBW color of an LED.
+type RGBW struct {
+	R, G, B, W uint8
+}
+
+type Device struct {
+	seesaw          *seesaw.Device
+	ledCount        int
+	pin             uint8
+	pixelType       PixelType
+	lastOperationAt time.Time
+}
+
+func New(dev *seesaw.Device, pin uint8, ledCount int, pixelType PixelType) (*Device, error) {
+
+	pixel := &Device{
+		seesaw:    dev,
+		ledCount:  ledCount,
+		pin:       pin,
+		pixelType: pixelType,
+	}
+
+	if !pixel.checkBufferLength(ledCount) {
+		return nil, fmt.Errorf("invalid LED count: %d", ledCount)
+	}
+
+	time.Sleep(seesawWriteDelay)
+
+	err := pixel.setupPin()
+	if err != nil {
+		return nil, fmt.Errorf("failed to update NeoPixel pin %d: %w", pin, err)
+	}
+
+	time.Sleep(seesawWriteDelay)
+
+	err = pixel.setupLedCount()
+	if err != nil {
+		return nil, fmt.Errorf("failed to update LED count %d: %w", ledCount, err)
+	}
+
+	time.Sleep(seesawWriteDelay)
+
+	err = pixel.setupSpeed()
+	if err != nil {
+		return nil, fmt.Errorf("failed to update pixel type: %w", err)
+	}
+
+	time.Sleep(seesawWriteDelay)
+
+	return pixel, nil
+}
+
+func (s *Device) setupLedCount() error {
+
+	lenBytes := calculateBufferLength(s.ledCount, s.pixelType)
+	buf := []byte{byte(lenBytes >> 8), byte(lenBytes & 0xFF)}
+	return s.seesaw.Write(seesaw.ModuleNeoPixelBase, seesaw.FunctionNeopixelBufLength, buf)
+}
+
+func calculateBufferLength(ledCount int, pixelType PixelType) int {
+	return ledCount * pixelType.EncodedLen()
+}
+
+func (s *Device) setupSpeed() error {
+	speed := byte(0)
+	if s.pixelType.Is800KHz() {
+		speed = 1
+	}
+	return s.seesaw.WriteRegister(seesaw.ModuleNeoPixelBase, seesaw.FunctionNeopixelSpeed, speed)
+}
+
+func (s *Device) setupPin() error {
+	return s.seesaw.WriteRegister(seesaw.ModuleNeoPixelBase, seesaw.FunctionNeopixelPin, s.pin)
+}
+
+// WriteColorAtOffset updates the color for a single LED at the given offset
+func (s *Device) WriteColorAtOffset(offset uint16, color RGBW) error {
+
+	encodedLen := s.pixelType.EncodedLen()
+
+	buf := make([]byte, encodedLen)
+	l := s.pixelType.PutRGBW(buf, color)
+	if l != encodedLen {
+		panic("unexpected encoded length: " + strconv.Itoa(l) + " != " + strconv.Itoa(encodedLen))
+	}
+	byteOffset := offset * uint16(encodedLen)
+	return s.writeBuffer(byteOffset, buf)
+}
+
+// WriteColors writes the given colors to the seesaws NeoPixel buffer
+func (s *Device) WriteColors(buf []RGBW) error {
+
+	if len(buf) > s.ledCount {
+		return fmt.Errorf("buffer too big, only %d LEDs setup: %d > %d", s.ledCount, len(buf), s.ledCount)
+	}
+
+	encodedLen := s.pixelType.EncodedLen()
+
+	tx := make([]byte, encodedLen*len(buf))
+	pos := 0
+	for _, c := range buf {
+		w := tx[pos:]
+		n := s.pixelType.PutRGBW(w, c)
+		pos += n
+	}
+
+	// the seesaw can at most deal with 30 bytes according to the datasheet, but
+	// crashes after 29 bytes. So we only send 29 data bytes at a time
+	const chunkSize = 29
+
+	// write the data chunk-by-chunk
+	for i := 0; i < len(tx); i += chunkSize {
+		toSend := tx[i:min(i+chunkSize, len(tx))]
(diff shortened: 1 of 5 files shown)


seesaw/seesaw.go
```diff
@@ -0,0 +1,113 @@
+package seesaw
+
+import (
+	"errors"
+	"strconv"
```
trichner: Is there a more lightweight alternative to `strconv`?

seesaw/seesaw.go
```diff
@@ -0,0 +1,113 @@
...
+	"time"
+)
+
+const DefaultSeesawAddress = 0x49
+
+// empirically determined delay, the one from the official library seems to be too short (250us)
+const defaultDelay = 10 * time.Millisecond
```
trichner, line 11: I'd be happy to have a better alternative. That's also how the official library does it :/

aykevl:
If I understand correctly, this driver works with an external I2C device that actually sends out the WS2812 protocol?

We already have a ws2812 driver, perhaps this driver could match the interface? In particular, the `WriteColors` method.

trichner:
> If I understand correctly, this driver works with an external I2C device that actually sends out the WS2812 protocol?
> 
> We already have a ws2812 driver, perhaps this driver could match the interface? In particular, the `WriteColors` method.

You are correct.

@aykevl it seems the current driver deals with RGBA colors discarding the A part. According to the datasheet NeoPixels are either RGB or RGBW. Some ideas:

1. add `WriteColorsRGB` and `WriteColorsRGBW` or similar
2. add something similar to color.RGBA for RGBW and have a single `WriteColors` (my preference) with maybe a convenience `WriteColorsRGB` accepting RGBA discarding the A

What are your thoughts?

For my own reference: https://github.com/tinygo-org/drivers/blob/release/ws2812/ws2812.go#L36
Also found: https://github.com/davidadeleon/gophercon2022Badge/blob/main/neopixel/neopixel.go

seesaw/neopixel/neopixel.go
```diff
@@ -0,0 +1,181 @@
...
+	l := s.pixelType.PutRGBW(buf, color)
+	if l != encodedLen {
+		panic("unexpected encoded length: " + strconv.Itoa(l) + " != " + strconv.Itoa(encodedLen))
+	}
+	byteOffset := offset * uint16(encodedLen)
+	return s.writeBuffer(byteOffset, buf)
+}
+
+// WriteColors writes the given colors to the seesaws NeoPixel buffer
+func (s *Device) WriteColors(buf []RGBW) error {
```
trichner, line 104: Similar to https://github.com/tinygo-org/drivers/blob/release/ws2812/ws2812.go#L36 but with RGBW (that's what the device supports).

trichner:
@aykevl 
What else is needed to get a review/get this merged?

I'm eager to add more but don't want to make this PR any bigger ;)

trichner:
@deadprogram may I request a review? :)

seesaw/seesaw.go
```diff
@@ -0,0 +1,126 @@
...
+	time.Sleep(delay)
+
+	return d.bus.Tx(d.addr, nil, buf)
+}
+
+func (d *Device) Write(module ModuleBaseAddress, function FunctionAddress, buf []byte) error {
+	prefix := []byte{byte(module), byte(function)}
+	data := append(prefix, buf...)
+	return d.bus.Tx(d.addr, data, nil)
+}
```
aykevl, line 126: These look like internal methods, why did you export them?

seesaw/seesaw.go
```diff
@@ -0,0 +1,126 @@
...
+		lastErr = err
+		time.Sleep(10 * time.Millisecond)
+	}
+
+	if lastErr != nil {
+		return fmt.Errorf("failed to read hardware ID after reset, tryed %d times: %w", tries, lastErr)
+	}
+
+	return nil
+}
```
aykevl, line 59: This also looks like an internal function.

seesaw/seesaw.go
aykevl: Sounds like this is actually a bug in the Seesaw (I2C should not need a delay like this).
Seems fine by me, if this is what's needed.

seesaw/seesaw.go
```diff
@@ -50,7 +52,7 @@ func (d *Device) Begin() error {
 	}
 
 	if lastErr != nil {
-		return lastErr
+		return fmt.Errorf("failed to read hardware ID after reset, tryed %d times: %w", tries, lastErr)
```
aykevl, line 55: Please don't use `fmt` in drivers, that package is far too big for some microcontrollers.

seesaw/seesaw.go
```diff
@@ -67,7 +69,18 @@ func (d *Device) ReadHardwareID() (byte, error) {
...
+
+// ReadVersion reads the version bytes from the device (undocumented in the datasheet)
+func (d *Device) ReadVersion() (uint32, error) {
+
+	buf := make([]byte, 4)
+	err := d.Read(ModuleStatusBase, FunctionStatusVersion, buf, defaultDelay)
+	if err != nil {
+		return 0, err
+	}
+	return (uint32(buf[0]) << 24) | (uint32(buf[1]) << 16) | (uint32(buf[2]) << 8) | uint32(buf[3]), nil
```
aykevl, line 83: You have made some changes to the generic seesaw support (seesaw/seesaw.go), perhaps this should be included in #565?

aykevl (commented):
I'm experimenting with different APIs in https://github.com/aykevl/board, and I've actually added addressable LEDs yesterday. It uses generics to abstract over different color formats.
I will make a proposal for a new API sometime in the future, but that's the future, and not very useful now for the Seesaw.

If you ask me, I'd actually recommend making the PR a lot smaller to only implement the basics. That is:

* `WriteColors` - but perhaps using `color.RGBA` to use the same interface as the ws2812 package.

And not this:

  * `PixelType` (just default to GRB for now, like the ws2812 driver).
  * `RGBW` type.

I'm not sure about `SpeedKHz800` - are people still using 400kHz? As far as I know, practically all WS2812-like LED strips are 800kHz by now. I recommend not including it, and adding it if there is demand for it (to reduce the API surface).

Maybe I'm going a bit too far though, I like to keep things simple. As they say: "no is temporary, yes is forever". It's easy to add new features at a later time if needed, it's very difficult to remove or change them once included.

aykevl:
Oh and I forgot, can you also add an example to the examples directory, and a smoke test to smoketest.sh?

seesaw/seesaw.go
trichner: Fair enough. I went through the other drivers and didn't find a decent solution for adding crucial information
to errors apart from straight byte-to-string conversions.

I can whip up a quick and dirty byte-to-hex formatter or just not add the context to the error.
What is the preferred solution?

seesaw/seesaw.go
trichner: I'll clean it up and move it into the `SoftReset`, it's anyways just a 'send reset command and wait for boot' method.

seesaw/seesaw.go
trichner: They are not internal, they are quite crucial to build drivers for the other modules ontop of it. An alternative would be to just dump everything into the same package, but most of the time driver users are interested in one or two of the seesaw modules but not the rest.

I think putting it all in the same package probably doesn't make sense.

Hence the Read/Writer as well as ReadRegister and WriteRegister are exposed.

trichner:
First of, thanks a ton for your review and feedback! I'm still a bit new to tinygo and its drivers, so bear with me while I learn :)

In general regarding the structure that I envisoned:

1. package `seesaw` supporting the basic function Read/Write operations as well as reset. I'll remove the other bits, they are not necessary and can be added later.
2. package `seesaw/neopixel` abstracting the neopixel module on the seesaw, depends on seesaw package. Only implementing the pixel type & speed I need for 4. for now.
3. package `seesaw/keypad` abstracting the keypad module on the seesaw, depends on seesaw package
4. package `neotrellis`, which is a 4x4 keyboard driven by the seesaw chip using the `neopixel` and `keypad` module and depending on them

Overall I don't think having one package trying to support all modules and functions is helpful. It also stops others from implementing their own abstractions for the seesaw modules without either copy-ing the package or extending it here.

I think it makes sense to also structure my PRs that way, so I'll incorporate your feedback and split it into those four.


> If you ask me, I'd actually recommend making the PR a lot smaller to only implement the basics. That is:
>     WriteColors - but perhaps using color.RGBA to use the same interface as the ws2812 package.
> 
> And not this:
> 
>    PixelType (just default to GRB for now, like the ws2812 driver).
>    RGBW type.

Removing `PixelType` in favour of hardcoding a default sounds good to me. Regarding RGBW see below.

> I'm not sure about SpeedKHz800 - are people still using 400kHz? As far as I know, practically all WS2812-like LED strips are 800kHz by now. I recommend not including it, and adding it if there is demand for it (to reduce the API surface).

I'm not sure either, that's what the seesaw itself supports. But can be added later, will remove in favour of hardcoded 800KHz

> Maybe I'm going a bit too far though, I like to keep things simple. As they say: "no is temporary, yes is forever". It's easy to add new features at a later time if needed, it's very difficult to remove or change them once included.

There I agree very much :)

Though one thing regarding the `color.RGBA`: This is not a 'temporary no' but rather intentionally building an API that I can not extend to the full functionality of what the seesaw device can without breaking compatibility or adding another method.
Therefore I think it makes sense to have the complexity of the RGBW type while dropping all actual support for flexibility for now.

What do you think?

trichner:
> Oh and I forgot, can you also add an example to the examples directory, and a smoke test to smoketest.sh?

Yes, shall do :)

trichner:
I've finally come around to give it another go, apparently https://github.com/tinygo-org/tinygo/commit/9126b9573711b1990f80ca1bca564c4886b8f8f2 fixed my timing issues that I couldn't resolve. Once the next tinygo release is out I intend to give this another go.

deadprogram:
Sounds great @trichner

trichner:
As promised, taken up again in #612. Closing this.

PR #567, opened 2023-05-09
st7789: allow changing the color format using COLMOD

aykevl:
This is only really useful for applications that use DrawRGBBitmap8. The RGB444 format has fewer colors, but can be up to 25% faster than the default RGB565.

I have tested this change on the Gopher Badge and the PineTime. Both can be substantially faster once the code is modified to output RGB444 instead of RGB565.


deadprogram:
Seems legit, thanks @aykevl now merging.

PR #568, opened 2023-05-09
st7789: update saved rotation in SetRotation

aykevl:
I missed this issue in https://github.com/tinygo-org/drivers/pull/550. Without it, `Rotation()` returns the wrong rotation value.


deadprogram:
Thanks for the fix @aykevl now merging.

Issue #552, opened 2023-04-24
Add MPU6886 driver


ivoszz:
I want to add M5Stick-C to tinygo as a supported platform with its MPU6886 (also M5Stack Core/Core2) accelerometer/gyroscope as a driver. I started by adding basic functions of the driver based on the MPU6050. But I found that I don't quite understand the concept of the return values of the ReadAcceleration() and ReadRotation() functions. These functions do not work with sensitivity settings (e.g. +/- 2g, 4g, 8g, 16g) and always return scaled raw counter values. So if I set the sensitivity to 16g, the return values are not in the range +/- 1000000, but approximately +/- 125000. Is it then up to the user to scale the result correctly?

I am able to add basic functionality in a short time. But MPU6050 doesn't seem like an ideal model to me, is it possible to recommend some other accelerometer/gyroscope unit? I would like to add more advanced features like using FIFO or IRQ later. Is there a good example? I'm not very good at naming variables and functions and a template would come in handy.

I've read the terms and conditions for contributing, is there anything I should pay particular attention to? 

Thanks for any suggestions.

soypat:
Yes, the existing mpu6050 driver is extremely limited. Please refer to the rewrite here https://github.com/tinygo-org/drivers/pull/556 for a more ideal approach to implementing future IMU drivers.

This implementation does not have FIFO or IRQ but could possibly be extended.

deadprogram:
Closing since it was added already. Thanks everyone!

PR #569, opened 2023-05-10
mpu6886: initial implementation

ivoszz:
This is an initial implementation of MPU6886 driver. I have tried to follow all the suggestions in Issues and Wiki. ( #552 )

### DIFF
--- a/examples/mpu6886/main.go
+++ b/examples/mpu6886/main.go
@@ -1,0 +1,22 @@
+// Connects to an MPU6886 I2C accelerometer/gyroscope.
+package main
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/mpu6886"
+)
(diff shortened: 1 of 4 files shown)


deadprogram:
This looks very good @ivoszz could you also please add to the smoke tests located here: https://github.com/tinygo-org/drivers/blob/dev/smoketest.sh

Thank you!

ivoszz:
Which platform should I add a test for? Mpu6886 is usually part of M5Stack systems, but also exists as a standalone module. However, M5Stack systems do not yet have the I2C bus implemented. I use the I2Csoft driver for my needs.

deadprogram:
How about `-target=pico`?

deadprogram (approved):
Looking good, thanks for adding the smoke test @ivoszz

PR #571, opened 2023-05-19
sdcard: remove tinyfs example and replace with link to tinyfs repo

deadprogram:
This PR removes the `sdcard` tinyfs example and replaces it with a link to the `tinyfs` repo in the docs. Better to maintain that in one place. This also allows us to remove the tinyfs dependency from drivers which was only needed for the one example, which already exists in the tinyfs repo anyhow.


aykevl (commented):
Seems good to me. It doesn't look like the TinyFS example has all the board specific files though?

deadprogram:
Once the https://github.com/tinygo-org/tinygo/issues/3460 has been addressed, we can work on adding example SD-card based code to the TinyFS repo. But for now, I am basically just removing things that are currently broken or confusing.

PR #539, opened 2023-03-29
initial support for TTP229 (BSF only)

conejoninja:

PR #572, opened 2023-05-20
Initial support TTP229 bsf driver

conejoninja:
Original PR here : https://github.com/tinygo-org/drivers/pull/539


deadprogram:
Thanks for adding this @conejoninja now merging.

Can you please also add it to the tinygo-site like this https://github.com/tinygo-org/tinygo-site/pull/359

Issue #540, opened 2023-04-01
WS2812 driver does not work with XIAO SAMD21


ysoldak:
I've tried to drive RGB Led with XIAO board (SAMD21, 48MHz chip)  and it does not work.

When trying the example, all 10 LEDs are very bright, almost white. One can see a hint of green and red, changing though.

Same example on a RP2040 (125MHz) board just works (blinking green and red).

aykevl:
I checked on the Circuit Playground Express (same chip) and it works fine.
Can you double check:
- the wires (is it really connected correctly - perhaps try another pin)
- power requirements (maybe it doesn't have enough power to drive these LEDs correctly), also note that the ground must be shared between the MCU and the LEDs.

If it's not a wire, it's hard to say what's wrong. It might be a faulty crystal oscillator or any number of other things. The way to know for sure is to connect it to a logic analyzer and look at the signal.

Something else you could try: testing the same setup but using the Arduino IDE and a library like FastLED.

ysoldak:
I've verified again, here are results:
Control (XIAO RP2040) - Correct
![XIAO_RP2040](https://github.com/tinygo-org/drivers/assets/1061565/488e090c-e5b8-4589-ad2c-922e55274425)
Experiment (XIAO SAMD21) - Too bright
![XIAO_SAMD21](https://github.com/tinygo-org/drivers/assets/1061565/cd865434-6a41-4768-8e81-e1fa92840361)

Here is overview of the setup, wires indeed quite long and connections may be weak, but it works for control...
![Setup](https://github.com/tinygo-org/drivers/assets/1061565/263eac21-abbb-4273-a6a8-ba2c106e9fab)

To make this work, I've added following two files to the ws2812 example.
And unmatched these tags in `others.go`

XIAO RP2040
`./examples/ws2812/xiao-rp2040.go`
```
//go:build xiao_rp2040

package main

import "machine"

var neo = machine.D8
var led = machine.LED

func init() {
	machine.NEOPIXEL_POWER.Configure(machine.PinConfig{Mode: machine.PinOutput})
	machine.NEOPIXEL_POWER.High()
}
```
`tinygo flash -target=xiao-rp2040 -size=full ./examples/ws2812/`

XIAO SAMD21
`./examples/ws2812/xiao.go`
```
//go:build xiao

package main

import "machine"

var neo = machine.D8
var led = machine.LED
```
`tinygo flash -target=xiao -size=full ./examples/ws2812/`

[...]

PR #575, opened 2023-05-25
ws2812: examples for some SAMD21 and RP2040 boards

ysoldak:
Also as for demonstration of #540 issue


aykevl (changes requested):
I have made a fix for #540, so is this still necessary?

If you still would like to merge this, please update the comments.

ysoldak:
No, these are just examples, not necessary.

PR #576, opened 2023-05-25
Add support for TM1638 based boards

ghost:
I added driver for TM1638.  Chip using serial interface (CLK, STB, DIO) unlike TM1637 which exists in drivers repository.

Datasheet describes TM1638 as an LED Controller driven on a 1/16 to 14/16 duty factor. 10 segment output lines, 8 grid output lines, 8 segment/key scan output lines, one display memory, control circuit, key scan circuit are all incorporated into a single chip to build a highly reliable peripheral device for a single chip microcomputer. Serial data is fed to TM1638 via a three-line serial interface.


tm1638/tm1638.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Create new TM1638 device
+func New(strobe machine.Pin, clock machine.Pin, data machine.Pin) Device {
+	strobe.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	clock.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	return Device{strobe: strobe, clock: clock, data: data}
+}
+
+// Configure TM1638
+func (d *Device) Configure() {
```
aykevl: Can you add a `Config` parameter?

```go
type Config struct {
    Brightness uint8
}
```

Now it's only for the Brightness value, but it may be useful for other configuration later.

tm1638/tm1638.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	d.strobe.High()
+}
+
+// Set display brightness
+func (d *Device) Brightness(value uint8) {
+	if value == 0 {
+		d.sendCommand(zeroBrightness)
+	} else {
+		d.sendCommand(brightness | (value & maxBrightness))
```
aykevl: This can be confusing to users: if they use the value 8 then they suddenly get a dark screen. Instead, I suggest:

* Exporting maxBrightness as `MaxBrightness`
* using `if value > MaxBrightness { value = MaxBrightness }` to limit the value

tm1638/tm1638.go
```diff
@@ -0,0 +1,137 @@
...
+	d.transmissionDelay()
+	d.write(baseAddress)
+	for i := 0; i < 16; i++ {
+		d.write(0)
+	}
+	d.strobe.High()
+}
+
+// Set display brightness
+func (d *Device) Brightness(value uint8) {
```
aykevl: The Go convention is to prefix setters with `Set`, so `SetBrightness`. Otherwise it looks like a getter.

tm1638/tm1638.go
```diff
@@ -0,0 +1,137 @@
...
+
+const (
+	addressAutoIncrement = 0x40
+	fixedAddress         = 0x44
+	baseAddress          = 0xC0
+	maxAddress           = 0x0F
+	readKeys             = 0x42
+	brightness           = 0x88
+	maxBrightness        = 0x07
+	zeroBrightness       = 0x80
```
aykevl: Some (but not all) of these constants are commands. Can you prefix them in some ways? For example, `cmdBrightness` would be a lot clearer to me.

tm1638/tm1638.go
```diff
@@ -0,0 +1,137 @@
...
+		d.transmissionDelay()
+		d.data.Set((value & (1 << i)) > 0)
+		d.transmissionDelay()
+		d.clock.High()
+		d.transmissionDelay()
+	}
+}
+
+func (d *Device) transmissionDelay() {
+	time.Sleep(time.Microsecond * time.Duration(2))
```
aykevl, line 160: Perhaps this is a good use for [`delay.Sleep`](https://pkg.go.dev/tinygo.org/x/drivers/delay#Sleep)?

tm1638/tm1638.go
```diff
@@ -0,0 +1,137 @@
...
+func (d *Device) Brightness(value uint8) {
+	if value == 0 {
+		d.sendCommand(zeroBrightness)
+	} else {
+		d.sendCommand(brightness | (value & maxBrightness))
+	}
+}
+
+// Write one display memory element
+func (d *Device) Write(offset uint8, data uint8) {
```
aykevl: Is this really useful when we also have `WriteArray` (or `WriteAt`, see next comment)?

tm1638/tm1638.go
```diff
@@ -0,0 +1,137 @@
...
+	d.strobe.Low()
+	d.transmissionDelay()
+	d.data.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	d.write(baseAddress | (offset & maxAddress))
+	d.write(data)
+	d.strobe.High()
+}
+
+// Write array to display memory
+func (d *Device) WriteArray(offset uint8, data []uint8) {
```
aykevl: What do you think of the `WriteAt` API? I think it would fit here really well.

examples/tm1638/tm1638.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/tm1638"
+)
+
+func main() {
+	// codes of nombers from 0 to 7 at odd indexes
```
aykevl:
```suggestion
	// codes of numbers from 0 to 7 at odd indexes
```

tm1638/tm1638.go
```diff
@@ -0,0 +1,137 @@
...
+	d.data.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	d.write(baseAddress | (offset & maxAddress))
+	for _, element := range data {
+		d.write(element)
+	}
+	d.strobe.High()
+}
+
+// Scan keyboard
+func (d *Device) ScanKeyboard(buffer *[4]uint8) {
```
aykevl: Can you clarify this a bit? A comment "Scan keyboard" for a function called `ScanKeyboard` is entirely useless. A comment should at the very least explain what the `buffer` parameter means.

examples/tm1638/tm1638.go
```diff
@@ -0,0 +1,86 @@
...
+	toShow := []uint8{0x3f, 1, 0x06, 0, 0x5b, 1, 0x4f, 0, 0x66, 1, 0x6d, 0, 0x7d, 1, 0x07}
```
aykevl: That's an odd sequence of bytes, it doesn't look at all like ASCII text. I think you should explain what this means, and/or add documentation to `WriteArray` to explain what this array really means (if it isn't ASCII text).

tm1638/tm1638.go
ghost: Do I understand correctly that the [proposed function ](https://github.com/tinygo-org/drivers/blob/dev/delay/sleep.go) makes optimization for delays less than 16 ms only?

For data transmission, accurate delay is not very important. The minimum delay of 1 micro second is described in the datasheet. Data transfer with longer delays  succeeds as well. Therefore, I would prefer to use the standard delay. However, I do not insist on this. I'm new to Go and some aspects I may not understand yet. To apply [delay.Sleep](https://pkg.go.dev/tinygo.org/x/drivers/delay#Sleep) will I need to undo this fork and fork the [dev branch](https://github.com/tinygo-org/drivers/tree/dev)?

tm1638/tm1638.go
ghost: Ok

tm1638/tm1638.go
ghost: Ok

tm1638/tm1638.go
ghost: Ok

tm1638/tm1638.go
ghost: Ok

examples/tm1638/tm1638.go
ghost: Ok

tm1638/tm1638.go
ghost: Ok

examples/tm1638/tm1638.go
ghost: Ok

tm1638/tm1638.go
ghost: From API description [io.WriterAt](https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/io/io.go;l=245)
```go
type WriterAt interface {
	WriteAt(p []byte, off int64) (n int, err error)
}
```

I think that parameter `off` is oversized. For PC programs it doesn't matter. But for small places additional 3 bytes are expensive. Perhaps it's a good reason not to use an interface?

Anyway I already renamed the function and reordered the parameters so it looks like the given interface:
```go
func (d *Device) WriteAt(data []uint8, offset uint8)
```

tm1638/tm1638.go
ghost: In the driver for an IC, I want to provide an interface to all functions described in the datasheet. The next step is to add the driver for the specific MDU1093 board. The chip driver will help those who assemble their own device. The board driver will provide convenient access to the control of specific indicators and buttons.

So I need to implement fixed address writing function.

tm1638/tm1638.go
aykevl: Okay, `time.Sleep` seems good.

tm1638/tm1638.go
aykevl: > In the driver for an IC, I want to provide an interface to all functions described in the datasheet.

We don't do that for most drivers, we implement the things that make sense and are needed in practice.
Remember, it's very easy to add a new method if this is needed (just make a PR!), but refactoring/maintaining code that may or may not be used at all is tedious.

tm1638/tm1638.go
aykevl: `int64` has very little overhead on a Cortex-M chip: it takes up just two registers. In fact, `uint32` is equal to (or better!) than `uint8` on Cortex-M chips because the registers are 32-bits in size.
You don't need to worry about the `int64` in this case.

tm1638/tm1638.go
ghost: Ok

tm1638/tm1638.go
ghost: Ok

ghost:
I have changed target branch to dev

PR #578, opened 2023-05-28
I2C: discard WriteRegister and ReadRegister methods in interface

### LINKED ISSUE #559: Discard `ReadRegister`+`WriteRegister` methods from `drivers.I2C` interface
### Background
Today the `I2C` interface looks like so:
```go
// I2C represents an I2C bus. It is notably implemented by the machine.I2C type.
type I2C interface {
	ReadRegister(addr uint8, r uint8, buf []byte) error
	WriteRegister(addr uint8, r uint8, buf []byte) error
	Tx(addr uint16, w, r []byte) error
}
```
As the name suggests, the `machine` package has a file called `i2c.go` that takes the usual `I2C implementation, which is often implemented with only the `Tx` method for most if not all microcontrollers, and implements the `ReadRegister` and `WriteRegister` methods for the I2C type roughly as follows:

```go
func (i2c *I2C) WriteRegister(address uint8, register uint8, data []byte) error {
	return i2c.Tx(uint16(address), append([]byte{address}, data...), nil)
}

func (i2c *I2C) ReadRegister(address uint8, register uint8, data []byte) error {
	return i2c.Tx(uint16(address), []byte{register}, data)
}
```

It cannot be denied this is **extremely** convenient for writing drivers.

That said, it does pose an unavoidable disadvantage: incompatibility with `periph`'s I2C implementation. This means tinygo drivers cannot be used with the `periph` package.

### Possible Solution
If we discard the `ReadRegister` and `WriteRegister` methods we will immedately break lots of drivers. There are 207 and 206 usages of these methods in this repository alone across 47 files.

Since most of these drivers store the I2C interface in-struct we can change the in-struct storage to be a type wrapping the `drivers.I2C` type which does implement these two methods like the `machine` package does and all should be solved.

### Impact
This would allow gophers that work with Raspberry Pi 3/4 to use TinyGo drivers freely. It would also mean that we can omit the ReadRegister WriteRegister implementations in the `machine` package, which to me seem out of place- an I2C implementation is "completely" defined by only the `Tx` method.

I've also tended to always avoid these two aforementioned methods since I am aware they allocate memory and I can easily implement methods on my device that avoid allocations since I usually know the size of the buffers I'm working with.

soypat:
Resolves #559. So I managed to pull this off because all usages of the I2C type were VERY consistent, so props to the tinygo developers!

### DIFF
--- a/i2c.go
+++ b/i2c.go
@@ -3,7 +3,7 @@ package drivers
 // I2C represents an I2C bus. It is notably implemented by the
 // machine.I2C type.
 type I2C interface {
-	ReadRegister(addr uint8, r uint8, buf []byte) error
-	WriteRegister(addr uint8, r uint8, buf []byte) error
+	// ReadRegister(addr uint8, r uint8, buf []byte) error
+	// WriteRegister(addr uint8, r uint8, buf []byte) error
 	Tx(addr uint16, w, r []byte) error
 }
--- a/i2c.go
+++ b/i2c.go
@@ -3,7 +3,15 @@ package drivers
 // I2C represents an I2C bus. It is notably implemented by the
 // machine.I2C type.
 type I2C interface {
-	// ReadRegister(addr uint8, r uint8, buf []byte) error
-	// WriteRegister(addr uint8, r uint8, buf []byte) error
+	// Tx performs a [I²C] transaction with address addr.
+	// Most I2C peripherals have some sort of register mapping scheme to allow
+	// users to interact with them:
+	//
+	//  bus.Tx(addr, []byte{reg}, buf) // Reads register reg into buf.
+	//  bus.Tx(addr, append([]byte{reg}, buf...), nil) // Writes buf into register reg.
+	//
+	// The semantics of most I2C transactions require that the w write buffer be non-empty.
+	//
+	// [I²C]: https://en.wikipedia.org/wiki/I%C2%B2C
 	Tx(addr uint16, w, r []byte) error
 }
--- a/tester/device8.go
+++ b/tester/device8.go
@@ -38,6 +38,9 @@ func (d *I2CDevice8) ReadRegister(r uint8, buf []byte) error {
 	if d.Err != nil {
 		return d.Err
 	}
+	if len(buf) == 0 {
+		d.c.Fatalf("no register buffer to read into")
+	}
 	d.assertRegisterRange(r, buf)
 	copy(buf, d.Registers[r:])
 	return nil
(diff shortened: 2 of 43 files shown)


i2c.go
```diff
@@ -3,7 +3,7 @@ package drivers
 // I2C represents an I2C bus. It is notably implemented by the
 // machine.I2C type.
 type I2C interface {
-	ReadRegister(addr uint8, r uint8, buf []byte) error
-	WriteRegister(addr uint8, r uint8, buf []byte) error
+	// ReadRegister(addr uint8, r uint8, buf []byte) error
+	// WriteRegister(addr uint8, r uint8, buf []byte) error
```
aykevl: Why not remove them entirely? I see no point in keeping them but commenting them out.
(If their purpose is documentation, then you can write that in regular English - something like "There used to be ReadRegister and WriteRegister methods but they were removed because XXX").

tester/device8.go
```diff
@@ -55,8 +58,18 @@ func (d *I2CDevice8) WriteRegister(r uint8, buf []byte) error {
...
+	case 0:
+		bus.c.Fatalf("i2c mock: need a write byte")
+		return nil
+	case 1:
+		return bus.ReadRegister(w[0], r)
+	default:
+		if len(r) > 0 || len(w) == 1 {
+			bus.c.Fatalf("i2c mock: unsupported lengths in Tx(%d, %d)", len(w), len(r))
+		}
+		return bus.WriteRegister(w[0], w[1:])
```
aykevl: I guess you could make `ReadRegister` and `WriteRegister` unexported now (meaning, `readRegister` and `writeRegister`) - but it's fine as-is.

aykevl (commented):
Overall looks good to me.

i2c.go
soypat: You got me- I have this horrible tendency to comment things out and just.... leave them there. Will remove!

tester/device8.go
soypat: Fixed!

soypat:
@aykevl Hey, do you think it's worth rewriting WriteRegister to use the stack in the optimistic case `data` is small?

```go
func WriteRegister(bus drivers.I2C, addr uint8, reg uint8, data []byte) error {
	var stackbuf [8]byte
	if len(data) <= 7 {
		stackbuf[0] = reg
		copy(stackbuf[1:], data)
		return bus.Tx(uint16(addr), stackbuf[:len(data)+1], nil)
	}
	buf := make([]uint8, len(data)+1)
	buf[0] = reg
	copy(buf[1:], data)
	return bus.Tx(uint16(addr), buf, nil)
}
```

aykevl:
Yeah that's an option, but actually I think a more reliable option would be to add a new type for drivers to use:

```go
type I2CWrapper struct { // name TBD
    drivers.I2C
    buf [8]byte
}

func (bus *I2CWrapper) WriteRegister(addr uint8, reg uint8, data []byte) error {
    if len(data) > len(bus.buf)-1 {
        panic("out of range") // or some other panic message
    }
    bus.buf[0] = reg
    copy(bus.buf[1:], data)
    return bus.Tx(uint16(addr), bus.buf[:len(data)+1], nil)
}
```

With this, you could verify that a driver never sends more than 7 bytes (for example) and replace `bus drivers.I2C` with `bus somepkg.I2CWrapper`.

The reason for this is that for some chips, `Tx` lets the parameters escape (so your alloc on the stack is still heap-allocated). Allocating the buffer while allocating the driver avoids this issue. This is not possible in the general case, but for specific drivers you _know_ how much you're ever going to send at once.

soypat:
> Yeah that's an option, but actually I think a more reliable option would be to add a new type for drivers to use:

Agreed, though for now lets leave it at this? Wrapping all i2c instances seems like labored work. I just had to replace patterns throughout the whole repo to get this PR.

deadprogram:
Thanks for all the work on this @soypat 

Merging now, and if we decide to do further work on the i2c interface it can happen from that point.

PR #580, opened 2023-06-10
add Sandbox Electronics NDIR CO2 sensor driver (2)

soypat:
Rebased previous PR #551. Implemented the new Sensor interface and added a smoke tests for it.


deadprogram:
Thanks for the addition @soypat now going to squash/merge.

Issue #573, opened 2023-05-23
SetScroll does not work well with NO_ROTATION in ST7789


sago35:
The reproduction method is as follows.

1. Change the following examples/initdisplay/gopher-badge.go setting to NO_ROTATION
2. tinygo flash --target gopher-badge --size short examples/basic

https://github.com/tinygo-org/tinyterm/tree/b6dd97f222d168362e41097602a4ecb14018e635

sago35:
As long as ROTATION_180 is used, it works fine.

aykevl:
`SetScrollArea` and `SetScroll` do not currently respect screen rotation.

My suggested fix is to add new APIs that look like this (and similar for horizontal scroll in 90° and 270° rotated screens):

```go
// Set up a vertical scroll area, with top and bottom being the number of pixels
// from the top or bottom that are fixed (do not scroll).
// This returns ErrInvalidScroll if the display does not currently support vertical scrolling,
// for example when it is rotated 90° or 270°.
func SetVerticalScrollArea(top, bottom int16) error

// Set the vertical scroll amount in pixels.
// For example, SetVerticalScroll(10) rotates the scrollable area around by 10 pixels.
func SetVerticalScroll(pixels int16) error
```

Notes:

  * It should respect the screen rotation, so that if the screen is rotated 180° then top and bottom are swapped for example (the behavior should be the same).
  * It should respect the row/column offset.
  * For ease of use, I think `SetXxxScroll` should also include the fixed top area so that `SetXxxScroll(10)` scrolls it down by 10 pixels, regardless of screen rotation, row/column offset, etc.

This can't easily be retrofitted into `SetScroll` because there is no way to know which direction is valid. It would be possible to at least fix it for 180° rotated screens though.

I wanted to add such an API a while ago for my own needs but haven't done so yet. I'll probably look into it soon.

aykevl:
Ran into this issue myself. It looks like the rotation configuration on the st7789 is in fact not correct: it seems like all rotation values are rotated by 180° (which is why 180° would work but 0° not). And of course, SetScrollArea doesn't respect the rotation.

I plan on making a PR to fix `SetScrollArea` and `SetScroll` to respect screen rotation, and to fix the rotation to be the actual rotation of the screen (and not off-by-180°). One step in that direction is #581.

aykevl:
I have a patch ready, but it depends on a few other patches.

aykevl:
[...]

PR #581, opened 2023-06-15
ili9341: st7789: fix SetScrollArea

aykevl:
The existing code was broken in a few ways:

  - It didn't use the correct operator precedence for the VSCRDEF VSA variable: it needed some extra parentheses to be correct.
  - It used the configured height instead of the actual display height for calculating VSA, which is incorrect. TFA+VSA+BFA must always be exactly 320, even if a lower screen height is configured.
  - If a lower than 320 pixel height is configured, the bottomFixedArea parameter applied to the whole 320 pixel screen height. Because this seems counter intuitive (and relies on properties of any given screen), I've changed it to work from the actual visible bottom of the screen (which may be smaller than 320 pixels).
    TODO: this doesn't take RowOffset into account, while it probably should.

I haven't fixed the st7735 implementation, because I didn't have example code on hand that would easily work on a st7735 screen. This is left as a TODO for the future.

Found while working on hardware scrolling support in https://github.com/aykevl/tinygl.

### DIFF
--- a/ili9341/ili9341.go
+++ b/ili9341/ili9341.go
@@ -320,10 +320,16 @@ func (d *Device) SetRotation(rotation drivers.Rotation) error {
 // SetScrollArea sets an area to scroll with fixed top/bottom or left/right parts of the display
 // Rotation affects scroll direction
 func (d *Device) SetScrollArea(topFixedArea, bottomFixedArea int16) {
+	if d.height < 320 {
+		// The screen doesn't use the full 320 pixel height.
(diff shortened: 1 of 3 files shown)


aykevl:
I have a patch ready that will fix https://github.com/tinygo-org/drivers/issues/573 but it depends on this patch and a decision on #582.

conejoninja:
It didn't break my code, but I don't have a screen of different size (320), I don't see any issue with the code itself, so ... merge?

deadprogram:
Thanks for the corrections @aykevl and to @conejoninja for review/test. Now merging.

PR #584, opened 2023-06-20
st7789: fix scrolling when rotated by 180°

aykevl:
This fixes https://github.com/tinygo-org/drivers/issues/573.
It doesn't handle 90° or 270°, I guess it needs a fix for 270° but I haven't tested that so didn't include it in the patch.

Tested on my PineTime on top of the current dev branch and it works in both 0° and 180° while previously it would only work with 180°.

Note that with this patch, the behavior matches the ili9341 driver, which appears to do exactly this in hardware (but the st7789 apparently doesn't).


conejoninja:
Thanks for your work one more time! merged!

PR #587, opened 2023-07-14
bma42x: add new BMA421/BMA425 driver

aykevl:
I wrote this for the PineTime, and all available sensors (accelerometer, step counter, temperature sensor) do work.

This commit also includes two "configuration files", that actually appear to be firmware files to run on the accelerometer for special features like step counting.
I'm not sure where they originally came for, and I don't know the copyright status of them. However, Bosch has open-sourced the driver for the very similar BMA423 which includes a similar binary blob and the InfiniTime and Wasp-OS projects have been shipping these blobs without issues, so I think it's reasonably safe to include these binary blobs directly in the source.

The chip is configured by default for high efficiency: it only consumes 14µA when enabled. There is also a performance mode, which can be supported by the driver if needed in the future.


deadprogram:
@aykevl smoke test?

aykevl:
I forgot, I added a minimal smoke test.

The problem with that smoke test is that I don't really have a way to actually test it. The PineTime doesn't really have a serial output and I have no other separate BMA421/BMA425 to test with. Therefore, it really is just a smoke test, to make sure the code keeps compiling.

(But of course, I do know the driver works because I'm actually using it in my PineTime).

deadprogram:
Thanks for the addition @aykevl now merging.

PR #589, opened 2023-08-07
Fix i2c legacy calls

kenbell:
Fix some uses of I2C WriteRegister that were causing `make test` to fail.  (Side note: looks like CI isn't running `make test`)


soypat (approved):
Missed these! Thanks for the PR. Looks good to me!

deadprogram:
Thanks for the fixes @kenbell now merging.

PR #597, opened 2023-09-17
Sensirion Sht4x Support 

trichner:
Support for the SHT4x family of humidity & temperature sensors from Sensirion.

I've tried to keep the interface aligned with the existing SHT3x driver where it made sense but kept it
simple.

### DIFF
--- a/sht4x/sht4x.go
+++ b/sht4x/sht4x.go
@@ -1,0 +1,79 @@
+// Package sht4x provides a driver for the SHT4x digital humidity sensor series by Sensirion.
+// Datasheet: https://www.sensirion.com/media/documents/33FD6951/64D3B030/Sensirion_Datasheet_SHT4x.pdf
+package sht4x
+
+import (
+	"time"
+
+	"tinygo.org/x/drivers"
+)
+
+const DefaultAddress = 0x44
+
+const (
+	// single-shot, high-repeatability measurement
+	commandMeasurement = 0xfd
+)
+
+// Device represents a SHT4x sensor
+type Device struct {
+	bus     drivers.I2C
+	address uint8
+}
+
+// New creates a new SHT4x connection. The I2C bus must already be
+// configured.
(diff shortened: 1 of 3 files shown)


deadprogram:
Hello @trichner thanks for the new device support.

Can you please also add it to the smoketests here? https://github.com/tinygo-org/drivers/blob/dev/smoketest.sh

trichner:
> Hello @trichner thanks for the new device support.
> 
> Can you please also add it to the smoketests here? https://github.com/tinygo-org/drivers/blob/dev/smoketest.sh

Done!
Though running them fails on some other test locally:
```
> make test
...
src/machine/board_bluepill.go:6:2: package device/stm32 is not in GOROOT (<redacted>/.cache/tinygo/goroot-3db76b2bbe6512878c9fc4ff18b9674eba0696a404ae4e2915298e4f7a087fad/src/device/stm32)
```

sht4x/sht4x.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Device represents a SHT4x sensor
+type Device struct {
+	bus     drivers.I2C
+	address uint8
+}
+
+// New creates a new SHT4x connection. The I2C bus must already be
+// configured.
+func New(bus drivers.I2C, addr uint8) Device {
```
deadprogram: I would be better to follow the pattern used by most drivers here, which is 

```go
type Device struct {
	bus     drivers.I2C
	Address uint8
}

func New(bus drivers.I2C) Device {
        return Device{
		bus:     bus,
		Address: DefaultAddress,
	}
}
```

If the dev wants to set the address to something else, it is then a matter of

```go
d := sht4x.New(machine.I2C0)
d.Address = 0x45
```

trichner: Done!

deadprogram:
Thank you for making the requested changes @trichner now going to squash/merge.

PR #599, opened 2023-09-18
pcf8523: RTC driver

trichner:
Driver for the pcf8523 RTC clocks as found on https://www.adafruit.com/product/2922 .

Interface similar to other RTC drivers :)


pcf8523/pcf8523_test.go
```diff
@@ -0,0 +1,91 @@
...
+func assertNoError(t testing.TB, e error) {
+	if e != nil {
+		t.Fatalf("unexpected error: %v", e)
+	}
+}
+func assertEquals[T comparable](t testing.TB, a, b T) {
+	if a != b {
+		t.Fatalf("%v != %v", a, b)
+	}
+}
```
trichner, line 91: I've tried to use quicktest instead but always run into a `panic: unimplemented: (reflect.Type).IsVariadic()
` error.

Happy to change if somebody can point me towards how to fix that.

trichner:
Test's fail because I use some go1.18 features, e.g. `pcf8523/pcf8523_test.go:30:2: implicit function instantiation requires go1.18 or later (-lang was set to go1.15; check go.mod)`

Is it intentionally still at go1.15 or is that just not updated?

deadprogram:
Should probably update the minimum version.

trichner:
> Should probably update the minimum version.

Bumped to 1.18, works all fine now.

examples/pcf8523/main.go
```diff
@@ -0,0 +1,35 @@
+package main
+
+import (
+	"fmt"
```
deadprogram: It is best to avoid the use of `fmt` package in examples, since that requires reflection and increases the size of the resulting binary. "Plain old println()" please. :smile_cat:

pcf8523/pcf8523.go
```diff
@@ -0,0 +1,127 @@
...
+//
+// Datasheet: https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf
+package pcf8523
+
+import (
+	"time"
+	"tinygo.org/x/drivers"
+)
+
+const DefaultAddress = 0x68
```
deadprogram: might be best to put all of the register const values into a separate `registers.go` file?

pcf8523/pcf8523_test.go
deadprogram: See https://github.com/tinygo-org/drivers/blob/dev/mcp23017/device_test.go for driver that is using it now.

deadprogram: I think you need

```go
c := qt.New(t)
bus := tester.NewI2CBus(c)
```

instead of 

```go
bus := tester.NewI2CBus(t)
```

trichner: Actually the linked test also doesn't work :face_with_head_bandage: 
```
~> tinygo test -v ./mcp23017
panic: unimplemented: (reflect.Type).IsVariadic()
FAIL    tinygo.org/x/drivers/mcp23017   3.049s
```

I use:
```
~> tinygo version
tinygo version 0.29.0 linux/amd64 (using go version go1.20.7 and LLVM version 16.0.6)
```

Looking at the sources that is indeed not implemented:
https://github.com/search?q=repo%3Atinygo-org/tinygo%20IsVariadic&type=code

But I see the tests are run against some bleeding edge version where the tests don't fail:
https://github.com/tinygo-org/drivers/actions/runs/6232602259/job/16916095548#step:5:5

But I couldn't find the corresponding tag/branch in the tinygo repo.

deadprogram:
@trichner I pulled your commits into #600 where I rebased them together and merged into `dev`. Thank you very much for all your work on this! Now closing this PR.

trichner:
> @trichner I pulled your commits into #600 where I rebased them together and merged into `dev`. Thank you very much for all your work on this! Now closing this PR.

Wonderful, thanks for the reviews and help!

PR #601, opened 2023-09-22
all: prepare for CGo changes in TinyGo

aykevl:
For details, see: https://github.com/tinygo-org/tinygo/pull/3927

This change just means we need to be more careful to use the right type, when types like C.uint32_t won't match to the Go equivalent (like uint32).


deadprogram (approved):
Seems better/safer.

PR #603, opened 2023-09-27
LoRa WAN US915 Support

mirackara:
Adds a driver for us 915 protocol to the lora WAN drivers.


iamemilio:
https://github.com/mirackara/drivers/pull/1

iamemilio:
@deadprogram should be good to go now, sorry I didn't get to test it. I got food poisoning 💀

iamemilio:
Note that because this re-structures the implementation of this driver, it will be breaking for https://github.com/tinygo-org/drivers/pull/535

deadprogram:
Here is an example where changing to `Channel` interface using functions instead of struct with values is a problem: https://github.com/hybridgroup/tinyglobo/blob/main/main.go#L25-L28

deadprogram:
The most straightforward solution to the TinyGlobo example within how this PR works, would be to provide `SetXXX()` functions for each of the values.

Example:

```go
func (c *ChannelAU) SetFrequency(freq uint32)  error { 
    c.frequency = freq
    return nil
}
```

There is some duplication of code, but it would allow for validation to be added at some point.

What do you think @mirackara @iamemilio ?

iamemilio:
I like that. It feels like a much safer way to handle this data structure.

deadprogram:
@iamemilio can you add to this PR please? :smile_cat:

lora/lorawan/region/au915.go
```diff
@@ -7,27 +7,56 @@ const (
 	AU915_DEFAULT_TX_POWER_DBM = 20
 )
 
+type ChannelAU struct {
```
soypat, line 10: This looks like identical code for all ChannelXX types- 
Should we have a base `channel` type and compose these "higher" level types with it?

soypat: ie:
```go
type channel struct {
	frequency       uint32
	bandwidth       uint8
	spreadingFactor uint8
	codingRate      uint8
	preambleLength  uint16
	txPowerDBm      int8
}
// Implement methods for channel here...

type ChannelAU struct {
     channel // Exposes exported channel methods.
}

soypat (commented):
Observed maybe a lot of code could be simplified a bunch, left comments

lora/lorawan/region/au915.go
deadprogram: @soypat that sounds like an excellent refactoring for a subsequent PR.

deadprogram:
Forgot to mention that I tested this in EU868 and it still works. As long as the US915 region is working, then I suggest we merge and then can do some additional refactoring from there.

lora/lorawan/region/au915.go
deadprogram: I will submit a PR that does exactly this later on...

deadprogram: See https://github.com/tinygo-org/drivers/pull/611

deadprogram:
Commits picked up in #611 so now closing this PR. Thank you very much everyone!

PR #604, opened 2023-10-06
 Fixing Copy-Paste Error in AT24CX and DS3231

graugans:
Hej, I came across those minor documentation mismatches when I was tinkering with my zs-042 module. I hope I followed the best practice with the commit message style and the correct branching model and naming.


conejoninja:
Thank you for your fix, now merging!

PR #605, opened 2023-10-08
Another minor documentation fix

graugans:
I came across another copy paste left over which I have corrected.

### DIFF
--- a/sh1106/sh1106.go
+++ b/sh1106/sh1106.go
@@ -51,7 +51,7 @@ type Buser interface {
 
 type VccMode uint8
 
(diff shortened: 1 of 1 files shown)


conejoninja:
Thanks for your work, now merging

graugans:
Hej @conejoninja it looks like the label `hacktoberfest` is not enough to be counted in the Hacktoberfest challenge. From my understaning you also need the `#hacktoberfest` topic in the project itself Otherwise it will not be listed.

conejoninja:
Unfortunately I do not have enough permissions to do that, we should invoke the mighty @deadprogram :smiley:  so he could add the tag to `drivers` repo, and maybe all of Tinygo  related ones too

PR #607, opened 2023-10-12
sh1106: fix I2C interface and add smoketest

sago35:
Fixed because build did not go through.
I also added smoketest since it was missing.


deadprogram:
Thanks for the fix and improvement @sago35 now merging.

Issue #608, opened 2023-10-14
SK6812 Support


SoftTacos:
The WS2812 driver does exactly what you want for RGB strips, but there are a considerable number of RGBA/SK6812 strips as well, SK6812 is the same as WS2812 but it as an additional byte for white/alpha. I went ahead and made the following modifications locally to accommodate for that. If you think this should be a separate driver all together so no breaking changes are added I'm happy to do that too.

Additionally, while I was poking around I noticed WriteByte can return an error, but Write and WriteColors ignore the error. I modified WriteColors to return the error from the last WriteByte call made. Would it make sense to instead check the machine.CPUFrequency() for valid values in Device.New() and return an error there? That way we avoid dealing with errors in Write, WriteByte, and WriteColor. This assumes machine.CPUFrequency() is constant, which the comments say it *usually* is.

Please let me know if you would like to see any changes made.

Example Code:
```
type DeviceType uint8

const(
	WS2812 DeviceType = iota // RGB
	SK6812 //RGBA / RGBW
)

// Device wraps a pin object for an easy driver interface.
type Device struct {
	Pin machine.Pin
	DeviceType DeviceType
}

// New returns a new WS2812 driver. It does not touch the pin object: you have
// to configure it as an output pin before calling New.
// WS2812 is for RGB, SK6812 is for RGBA
func New(pin machine.Pin,deviceType DeviceType) Device {
	return Device{
		Pin:pin,
		DeviceType:deviceType,
	}
}

// Device.Write() is unchanged

// Write the given color slice out using the WS2812 protocol.
// Colors are sent out in the usual GRB(A) format.
func (d Device) WriteColors(buf []color.RGBA) (err error) {
	switch d.DeviceType {
	case WS2812:
		err = d.writeColorsRGB(buf)
	case SK6812:
		err = d.writeColorsRGBA(buf)
	}
	return
}

func (d Device) writeColorsRGB(buf []color.RGBA) (err error) {
	for _, color := range buf {
		d.WriteByte(color.G)       // green
		d.WriteByte(color.R)       // red
		err = d.WriteByte(color.B) // blue
	}
	return
}

func (d Device) writeColorsRGBA(buf []color.RGBA) (err error) {
	for _, color := range buf {
		d.WriteByte(color.G)       // green
		d.WriteByte(color.R)       // red
		d.WriteByte(color.B)       // blue
		err = d.WriteByte(color.A) // alpha
	}
	return
}
```

[...]

PR #610, opened 2023-10-14
added support for SK6812 to WS2812 device

SoftTacos:
Adds support for SK6812 (RGBA) LEDs to WS2812 device.

PR for issue: https://github.com/tinygo-org/drivers/issues/608

### DIFF
--- a/ws2812/ws2812.go
+++ b/ws2812/ws2812.go
@@ -26,7 +26,7 @@ type Device struct {
 }
 
 // deprecated, use NewWS2812 or NewSK6812 depending on which device you want.
-// calls NewWS2812() to avoid breaking everyone's existing code. 
+// calls NewWS2812() to avoid breaking everyone's existing code.
 func New(pin machine.Pin) Device {
 	return NewWS2812(pin)
 }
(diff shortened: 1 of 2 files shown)


ws2812/ws2812.go
```diff
@@ -12,15 +12,27 @@ import (
 
 var errUnknownClockSpeed = errors.New("ws2812: unknown CPU clock speed")
 
+type DeviceType uint8
+
+const (
+	WS2812 DeviceType = iota // RGB
+	SK6812                   // RGBA / RGBW
+)
```
SoftTacos: If we would rather opt for making an entirely separate SK6812 device and leverage the same calls that Deivce.WriteByte uses, I'm happy to do that too.

deadprogram:
Probably creating specific constructors vs. modifying the current one would be a better approach.

```go
// deprecated, use NewWS2812 or NewSK6812 depending on which device you want.
// calls NewWS2812() to avoid breaking everyone's existing code.
func New(pin machine.Pin) Device

// returns WS2812
func NewWS2812(pin machine.Pin) Device

// returns SK6812
func NewSK6812(pin machine.Pin) Device
```

SoftTacos:
Love it, I'll get that changed soon.

deadprogram:
Thanks for working on this @SoftTacos now squash/merging.

PR #612, opened 2023-10-17
Adafruit 'seesaw' support

trichner:
Here we go again. Simple driver for the _"i2c to anything"_ seesaw chip.

https://learn.adafruit.com/adafruit-seesaw-atsamd09-breakout/overview

This also addresses half of #564.

### DIFF
--- a/seesaw/seesaw.go
+++ b/seesaw/seesaw.go
@@ -1,14 +1,21 @@
 // Package seesaw provides a driver implementation to communicate with Adafruit's seesaw chip.
-// There are many Adafruit boards that use a seesaw. Moisture sensors, LED keyboards, ...
+// There are many Adafruit boards that use a seesaw. Soil moisture sensors, LED keyboards, etc.
+//
+//   - Documentation: https://learn.adafruit.com/adafruit-seesaw-atsamd09-breakout/overview
+//   - Arduino driver: https://github.com/adafruit/Adafruit_Seesaw
+//   - Seesaw firmware: https://github.com/adafruit/seesaw
 package seesaw
 
 import (
 	"errors"
 	"strconv"
 	"time"
+
 	"tinygo.org/x/drivers"
 )
 
+// DefaultAddress is the I2C address the chips have by default. Most boards
+// built on top of it come with their own respective default addresses.
 const DefaultAddress = 0x49
 
 // empirically determined standardDelay, the one from the official library seems to be too short (250us)
@@ -18,16 +25,6 @@ const (
 	seesawHwIdCodeSAMD09  = 0x55 // HW ID code for SAMD09
 	seesawHwIdCodeTINY8x7 = 0x87 // HW ID code for ATtiny817
 )
-
-type Seesaw interface {
-
-	// Read reads a number of bytes from the device after sending the read command and waiting 'delay'. The delays depend
-	// on the module and function and are documented in the seesaw datasheet
-	Read(module ModuleBaseAddress, function FunctionAddress, buf []byte, delay time.Duration) error
-
-	// Write writes an entire array into a given module and function
-	Write(module ModuleBaseAddress, function FunctionAddress, buf []byte) error
-}
 
 type Device struct {
 	bus           drivers.I2C
@@ -54,8 +51,7 @@ func (d *Device) SoftReset() error {
 }
 
 func (d *Device) waitForReset() error {
-
-	//give the device a little bit of time to reset
+	// give the device a little bit of time to reset
 	time.Sleep(time.Second)
 
 	var lastErr error
@@ -115,8 +111,8 @@ func (d *Device) Read(module ModuleBaseAddress, function FunctionAddress, buf []
 		return err
 	}
 
-	//This is needed for the client seesaw device to flush its RX buffer and process the command.
-	//See seesaw datasheet for timings for specific modules.
+	// This is needed for the client seesaw device to flush its RX buffer and process the command.
+	// See seesaw datasheet for timings for specific modules.
 	time.Sleep(delay)
 
 	return d.bus.Tx(d.Address, nil, buf)
--- a/seesaw/seesaw.go
+++ b/seesaw/seesaw.go
@@ -18,8 +18,9 @@ import (
 // built on top of it come with their own respective default addresses.
 const DefaultAddress = 0x49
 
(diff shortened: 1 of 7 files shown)


seesaw/seesaw_test.go
```diff
@@ -0,0 +1,84 @@
...
+		when(append([]byte{0x0E, 0x04}, expectedWrite...), nil, nil),
+	)
+
+	sut := New(mocked)
+
+	err := sut.Write(ModuleNeoPixelBase, FunctionNeopixelBuf, expectedWrite)
+	assertDeepEquals(t, err, nil)
+}
+
+func assertDeepEquals[e any](t *testing.T, actual, expected e) {
```
trichner: I tried quicktest again, does not compute for me. I.e. even the existing usages on `dev` don't work locally.

deadprogram: >  even the existing usages on dev don't work locally.

That seems to imply something wrong with your local config. no?

trichner: Very fair point :)

1. I now went through re-building tinygo from v0.30.0 and still get the same issue:
```
~> tinygo test -v ./tester
panic: unimplemented: (reflect.Type).IsVariadic()
FAIL    tinygo.org/x/drivers/tester     1.314s
~> tinygo version
tinygo version 0.30.0 linux/amd64 (using go version go1.20.8 and LLVM version 16.0.1)
```

2. then I had a look at the workflow on the repo that runs it successfully, if I do the same with the docker image I still get the error:
```
> docker run --rm -it -v /<redacted>/tinygo_drivers:/build ghcr.io/tinygo-org/tinygo-dev:latest /bin/bash
root@2632e6b72f70:/go# cd /build
root@2632e6b72f70:/build# tinygo test ./tester
go: downloading github.com/frankban/quicktest v1.10.2
...
go: downloading golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543
panic: unimplemented: (reflect.Type).IsVariadic()
FAIL    tinygo.org/x/drivers/tester     1.366s
```

3. furthermore I found that even the [GitHub Action on the `dev`](https://github.com/tinygo-org/drivers/actions/runs/6516908120/job/17700893731) branch seems to succeed with **failing tests:**
Screenshot from 2023-10-22 01-16-21

I still don't know what's wrong with my environment, but I found a fix for the CI action not properly failing: #614

Any pointers? At least the docker environment should work, no?

trichner: Ok, I figured it out :facepalm: 

Apparently the way to run unit tests is to use `go test` and not `tinygo test`.

Out of curiosity, what's the purpose of `tinygo test` then?

It runs tests fine without the `quicktest` dependency :D

seesaw/seesaw.go
```diff
@@ -0,0 +1,126 @@
...
+	cmd[1] = byte(function)
+
+	err := d.bus.Tx(d.Address, cmd[:], nil)
+	if err != nil {
+		return err
+	}
+
+	// This is needed for the client seesaw device to flush its RX buffer and process the command.
+	// See seesaw datasheet for timings for specific modules.
+	time.Sleep(delay)
```
bgould: Was looking over some seesaw experiments I was working on a while back.  One thing that I did for the "standardDelay" was to use a functional interface ... I think I did it this way because I wanted flexibility to switch between busy-wait vs sleep (I've noticed sometimes when dealing with microseconds that is time.Sleep is not always super precise on every target, so that flexibility is useful to me).

Not saying that decision was correct/best (the code might not even be working, I see I still have some debugging print statements in there), but here is an example of what I mean in case you are curious about it:

https://github.com/bgould/go-seesaw/blob/main/seesaw.go#L13-L21
https://github.com/bgould/go-seesaw/blob/main/mini-tft-wing.go#L115-L119

bgould:
This looks generally good to me, I will look over some more tomorrow.  Was trying to find if I have my TFT Featherwing with me to test, but I think I might not have it (also will check more tomorrow).

I can picture that maybe in addition to this low-level package, eventually there could be sub-packages for devices that use it, like /drivers/seesaw/mini-tft-featherwing.  A "device" implemented on top of this package could look something like this:

https://github.com/bgould/go-seesaw/blob/main/mini-tft-wing.go

associated example:

https://github.com/bgould/go-seesaw/blob/main/examples/mini-tft-wing/main.go

(like I said I will try to see if I still have that board so that I can give your package a test drive)

trichner:
> This looks generally good to me, I will look over some more tomorrow. Was trying to find if I have my TFT Featherwing with me to test, but I think I might not have it (also will check more tomorrow).
> 
> I can picture that maybe in addition to this low-level package, eventually there could be sub-packages for devices that use it, like /drivers/seesaw/mini-tft-featherwing. A "device" implemented on top of this package could look something like this:
> 
> https://github.com/bgould/go-seesaw/blob/main/mini-tft-wing.go
> 
> associated example:
> 
> https://github.com/bgould/go-seesaw/blob/main/examples/mini-tft-wing/main.go
> 
> (like I said I will try to see if I still have that board so that I can give your package a test drive)

Indeed! And actually that's what I've done here: https://github.com/tinygo-org/drivers/issues/564

I added ontop Keypad & NeoPixel and then wrapped it all to make the NeoTrellis work nicely :)

Though in hindsight I think those are not yet polished enough for the tinygo-org/drivers repo ;)

seesaw/seesaw.go
trichner: Good though. I've actually not considered that.

I've now simplified it further to leave more room open to also add a `ReadWithWaiter(...)` or similar in the future while simplifying the interface for the most common use cases.
The 'delay' argument felt a bit awkward anyways. 

I think as a first start the current implementation will fit most of the applications.

If one must, one can change the `dev.ReadDelay = ...` before a read call.

trichner:
Bump? Any chance for a review?

trichner:
What's missing? Any interest?

bgould:
Sorry I was trying to find a board with seesaw on it to test, but I could not find.  All in all, LGTM, now merging.  Thank you for your contribution @trichner

trichner:
> Sorry I was trying to find a board with seesaw on it to test, but I could not find. All in all, LGTM, now merging. Thank you for your contribution @trichner

Ahh, that would have been great indeed. Thanks for merging.
I'll give it a go with the Neotrellis, its a more complex component with a seesaw than my soilsensor.

PR #615, opened 2023-11-06
epd2in13: update to common display APIs

aykevl:
This updates the e-paper display to use the same rotation and sleep APIs as other displays.


conejoninja:
I agree with this change, I see no problem merging it

deadprogram:
Now merging, thanks for the update @aykevl and to @conejoninja for review.

PR #616, opened 2023-11-06
epd2in13: use better black/white approximation

aykevl:
The previoius behavior was that entirely black pixels were treated as white, and anything else as black. That's at least counter-intuitive.
This patch changes the behavior to actually look at the color values and use a cutoff around medium gray: darker colors are treated as black, and lighter colors are treated as white.

This is a backwards incompatible change, but I think this behavior makes a lot more sense.


conejoninja (approved):
I agree with this change, it will break things, but it makes sense

deadprogram:
Thanks for improving this @aykevl and to @conejoninja for review. Now merging.

PR #619, opened 2023-11-17
sgp30: add the now-obsolete SGP30 air quality sensor

aykevl:
Tested with a rp2040.

TODO: add the ability to set the absolute humidity for more accurate sensor details. The formula for that is rather complex, so I've left this as a future addition.

---

I've combined this sensor with the SCD40 CO₂ sensor to create a simple air quality monitor at home.

### DIFF
--- a/examples/sgp30/main.go
+++ b/examples/sgp30/main.go
@@ -1,0 +1,53 @@
+package main
+
+// Example for the SGP30 to be used on a Raspberry Pi pico.
+// Connect the sensor I2C pins to GP26 and GP27 to test.
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/sgp30"
+)
+
+func main() {
+	time.Sleep(time.Second)
+	println("start")
+
+	// Configure the I2C bus.
+	bus := machine.I2C1
+	err := bus.Configure(machine.I2CConfig{
+		SDA:       machine.GP26,
+		SCL:       machine.GP27,
+		Frequency: 400 * machine.KHz,
+	})
+	if err != nil {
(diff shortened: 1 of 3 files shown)


conejoninja:
Thanks, I was using a SGP30 for a project of mine that got delayed, never got the SGP30 working properly as the values weren't what I expected and it stopped working at random times (probably a power issue I fixed later, but never got the time to check it).

Why use two different CO_2 sensors?

aykevl:
> Thanks, I was using a SGP30 for a project of mine that got delayed, never got the SGP30 working properly as the values weren't what I expected and it stopped working at random times (probably a power issue I fixed later, but never got the time to check it).

It stopped working for me too, for some reason. I haven't figured out the details yet, except it manages to bring down the attached Raspberry Pi too (so probably a power issue of some sort, maybe even a short?).
Note that the first ~15 seconds you will get stub results, not the actual measurements. Perhaps that was the issue?

> Why use two different CO_2 sensors?

The SGP30 is not a CO₂ sensor. It is a VOC sensor - even though marketing seems to suggest otherwise (and companies are probably only too happy to let this confusion remain). Instead of CO₂, it measures various other gases and makes a CO₂ estimate based on that, that is in practice highly inaccurate.

Historically, real CO₂ sensors have been somewhat expensive but nowadays a SGP40 can be bought at around €15 so it's not that expensive anymore. And it can be quite accurate after it's been calibrated to outside air.

deadprogram:
Thanks for the addition @aykevl and to @conejoninja for review. Now merging.

deadprogram:
Thanks for the addition @aykevl and to @conejoninja for review. Now merging.

PR #620, opened 2023-11-23
gbadisplay: add simple driver for the GameBoy Advance display

aykevl:
This only implements the 15 bits per pixel mode, not any of the other possible display modes. This matches conventional SPI displays most closely. Future additions might add more display modes.

Normally I'd argue interfacing with chip/board specific hardware should be done in the machine package using a generic interface, but the GBA is kind of special and I don't want the machine package to depend on the tinygo.org/x/drivers/pixel package.

### DIFF
--- a/gbadisplay/gbadisplay.go
+++ b/gbadisplay/gbadisplay.go
@@ -1,0 +1,85 @@
+// Package gbadisplay implements a simple driver for the GameBoy Advance
+// display.
+package gbadisplay
+
+import (
+	"device/gba"
+	"errors"
+	"image/color"
+	"runtime/volatile"
+	"unsafe"
+
+	"tinygo.org/x/drivers/pixel"
+)
+
+// Image buffer type used by the GameBoy Advance.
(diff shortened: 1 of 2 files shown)


deadprogram:
Perhaps this would be better off in the https://github.com/tinygo-org/tinygba repo?

aykevl:
Yes, sounds like a good idea. I'll make a PR there.

aykevl:
Ok, made a PR for the RGB555 pixel format here: https://github.com/tinygo-org/drivers/pull/621

Honestly I just wanted to add the RGB555 pixel format here, so that I could use it with my in-development board package here: https://github.com/aykevl/board/blob/main/board-gameboy-advance.go
I could also make a PR to the tinygba repo with the display, if you like? I'm also fine with keeping everything inside that board package and only adding the RGB555 pixel format.

deadprogram:
> I could also make a PR to the tinygba repo with the display, if you like?

It seems like a useful thing for people to be able to find, so yes please.

PR #621, opened 2023-11-24
pixel: implement RGB555 pixel format as used by the GameBoy Advance

aykevl:
See https://github.com/tinygo-org/drivers/pull/620 for background.


deadprogram:
Thanks @aykevl now merging.

PR #622, opened 2023-12-02
move netdev interface to use net/netip for IP addr/ports

scottfeldman:
Matches changes to netdever interface made in tinygo-org/net#8.  These two PRs are paired.  If one goes, both go.

### DIFF
--- a/espat/espat.go
+++ b/espat/espat.go
@@ -45,8 +45,7 @@ type Config struct {
 type socket struct {
 	inUse    bool
 	protocol int
-	lip      netip.Addr
-	lport    uint16
+	laddr    netip.AddrPort
 }
 
 type Device struct {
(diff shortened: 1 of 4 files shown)


espat/espat.go
```diff
@@ -44,8 +45,8 @@ type Config struct {
 type socket struct {
 	inUse    bool
 	protocol int
-	lip      net.IP
-	lport    int
+	lip      netip.Addr
+	lport    uint16
```
soypat: could be paired as a netip.AddrPort as `laddr`

soypat (approved):
LGTM!

espat/espat.go
scottfeldman: fixed

deadprogram:
Now merging, thanks for working on this @scottfeldman and for review @soypat

soypat:
Was looking at the code again and just thought to myself this was a really good move! Lots of heap allocations seem to be gone now! Thanks @scottfeldman for the contribution!

scottfeldman:
Yes, it was worth it.  What next @soypat?

soypat:
@scottfeldman Now we use the API! 
As for the netdev-3 PR I'm still working on the TCP implementation. it's coming along nicely but its got a while to go before being lwip!

PR #623, opened 2023-12-05
fix smoketest failures on examples/net examples

scottfeldman:
examples/net/socket test was failing because of missing updates to recent net/netip changes to netdev interface.  Oops!

examples/net/mqttclient/paho was failing because of new dependencies brought in by paho package that haven't been ported to TinyGo's net package.  Will open an issue on tinygo-org/net to fix the dependencies.

### DIFF
--- a/examples/net/socket/main.go
+++ b/examples/net/socket/main.go
@@ -13,8 +13,7 @@ import (
 	"fmt"
 	"log"
 	"machine"
-	"net"
-	"strconv"
+	"net/netip"
 	"time"
 
 	"tinygo.org/x/drivers/netdev"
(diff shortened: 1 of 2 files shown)


scottfeldman:
Added https://github.com/tinygo-org/net/issues/10 to fix the missing paho/mqtt dependencies.

deadprogram:
What about all the other examples that I mentioned in https://github.com/tinygo-org/net/issues/9 sounds like that is all the same issue?

scottfeldman:
> What about all the other examples that I mentioned in [tinygo-org/net#9](https://github.com/tinygo-org/net/issues/9) sounds like that is all the same issue?

I'll take a look at that one in the morning.  Probably some issue when converting to netip API.

deadprogram:
Thanks @scottfeldman

scottfeldman:
I believe tinygo-org/net#9 is fixed with tinygo-org/net/pull/11

deadprogram:
Thanks for the fixes @scottfeldman now merging.

PR #626, opened 2023-12-10
Add more CAN bus line speeds to mcp2515

sdismo:
Thanks for creating TinyGo. Truly fantastic. 
 
Completing a TODO in the code by adding all odd CAN bus line speeds available. I need only 250 kbps for playing with NMEA 2000 marine instruments but filled in the blanks also for other line speeds.


deadprogram:
Thanks for the addition @sdismo now squash/merging.

PR #628, opened 2023-12-18
correct netdever Accept() prototype

scottfeldman:
According to man page accept(2), accept returns new client sockfd and remote peer ip:port.  This patch corrects the Accept() prototype in the netdever interface to not take in an ip:port arg, but rather return an ip:port for remote peer.

Tested with examples/net/tcpecho on wioterminal and nano-rp2040.  Here's a run with wioterminal:

```
SERVER
============
sfeldma@nuc:~/work/drivers$ tinygo flash -monitor -target wioterminal -size short -stack-size=8kb ./examples/net/tcpecho
110876    2552   11212 |  113428   13764
Connected to /dev/ttyACM2. Press Ctrl-C to exit.

Realtek rtl8720dn Wifi network device driver (rtl8720dn)

Driver version           : 0.0.1
RTL8720 firmware version : 2.1.2
MAC address              : 2c:f7:f1:1c:9b:2f

Connecting to Wifi SSID 'test'...CONNECTED

DHCP-assigned IP         : 10.0.0.140
DHCP-assigned subnet     : 255.255.255.0
DHCP-assigned gateway    : 10.0.0.1

Starting TCP server listening on :8080
Client 10.0.0.190:50000 connected
Client 10.0.0.190:50000 closed

CLIENT
=============
nc -p 50000 10.0.0.140 8080
```

### DIFF
--- a/wifinina/wifinina.go
+++ b/wifinina/wifinina.go
@@ -676,10 +676,10 @@ func (w *wifinina) Listen(sockfd int, backlog int) error {
 	return nil
 }
 
-func (w *wifinina) Accept(sockfd int, ip netip.AddrPort) (int, error) {
+func (w *wifinina) Accept(sockfd int) (int, netip.AddrPort, error) {
 
 	if debugging(debugNetdev) {
-		fmt.Printf("[Accept] sockfd: %d, peer: %s\r\n", sockfd, ip)
+		fmt.Printf("[Accept] sockfd: %d\r\n", sockfd)
 	}
 
 	w.mu.Lock()
@@ -692,7 +692,7 @@ func (w *wifinina) Accept(sockfd int, ip netip.AddrPort) (int, error) {
 	switch socket.protocol {
 	case netdev.IPPROTO_TCP:
 	default:
-		return -1, netdev.ErrProtocolNotSupported
+		return -1, netip.AddrPort{}, netdev.ErrProtocolNotSupported
 	}
 
 	for {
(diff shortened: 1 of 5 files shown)


wifinina/wifinina.go
```diff
@@ -1123,6 +1125,23 @@ func (w *wifinina) accept(s sock) sock {
 	return newsock
 }
 
+func (w *wifinina) getRemoteData(s sock) netip.AddrPort {
+
+	if debugging(debugCmd) {
+		fmt.Printf("    [cmdGetRemoteData] sock: %d\r\n", s)
+	}
+
+	sl := make([]string, 2)
```
deadprogram, line 1134: Can we avoid the heap allocation here?

scottfeldman: I tried this:

```
@@ -1131,8 +1131,8 @@ func (w *wifinina) getRemoteData(s sock) netip.AddrPort {
                fmt.Printf("    [cmdGetRemoteData] sock: %d\r\n", s)
        }
 
-       sl := make([]string, 2)
-       l := w.reqRspStr1(cmdGetRemoteData, uint8(s), sl)
+       var sl [2]string
+       l := w.reqRspStr1(cmdGetRemoteData, uint8(s), sl[:])
        if l != 2 {
                w.faultf("getRemoteData wanted l=2, got l=%d", l)
                return netip.AddrPort{}
```

And then compared before and after using -print-allocs=.

No difference.

deadprogram: Just noticed it was a slice of string not byte, and also we are already doing this same thing in `getIP()`

deadprogram:
Thanks for the corrections @scottfeldman now merging.

PR #631, opened 2023-12-26
APDS9960 refactor

soypat:
PTAL @deadprogram @alankrantas @aykevl 

Changes:
* Remove heap allocations
* Handle all errors
* Much more idiomatic status and enable config setting/getting
* idiomatic transaction based I2C error handling
* simplify some parts of code.
* Change `~` character in comments (rendered as strikethrough in newer versions of gopls)

Still WIP: needs testing on my side.

### DIFF
--- a/apds9960/apds9960.go
+++ b/apds9960/apds9960.go
@@ -17,6 +17,8 @@ type Device struct {
 	Address uint8
 	mode    uint8
 	gesture gestureData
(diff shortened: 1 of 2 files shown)


soypat:
So it turns out what I have here is not an APDS9960, but rather an APDS9930 :facepalm: - guess I won't be able to test.

PR #635, opened 2024-01-10
fix wifinina UDP send

scottfeldman:
Fix an error I introduced in porting wifinina to netdev.  The driver was starting a client on the socket once, during Connect.  The first UDP send on the socket would succeed, any subsequent sends would fail.  The fix is to start the client on the socket for each UDP send.

I think I see the logic in this design, so the fix makes sense.  If the device was sending to many UDP clients, it could use a single socket, but change the dst addr for each send.  The pkt data would be queued to hw just once, and then sent from hw to each client dst addr.  This would be a real efficient way to multicast to many clients.

### DIFF
--- a/wifinina/wifinina.go
+++ b/wifinina/wifinina.go
@@ -165,7 +165,8 @@ type hwerr uint8
 
 type socket struct {
 	protocol int
-	ip       netip.AddrPort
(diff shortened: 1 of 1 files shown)


deadprogram:
Works to fix my problem @scottfeldman thank you very much! Now merging.

Issue #636, opened 2024-01-10
net, wioterminal: connect to access point claims success even if it has failed


deadprogram:
Tried to use the `wioterminal` with the latest `dev` branch, and it claimed to connect to a WiFi access point that was not actually turned on at the time. :ghost: 

```
$ tinygo flash -size short -target wioterminal -ldflags="-X main.ssid=noexist -X main.pass=dsfsdf" -monitor ./examples/net/ntpclient/
   code    data     bss |   flash     ram
 110112    2508   10128 |  112620   12636
Connected to /dev/ttyACM0. Press Ctrl-C to exit.

Realtek rtl8720dn Wifi network device driver (rtl8720dn)

Driver version           : 0.0.1
RTL8720 firmware version : 2.1.2
MAC address              : 2c:f7:f1:1b:48:35

Connecting to Wifi SSID 'noexist'...CONNECTED
```

scottfeldman:
Minimum wifi passphrase is 8 chars...I think that's the root of the problem, but need to look at the logic that checks the connection...

scottfeldman:
Fixed in https://github.com/tinygo-org/drivers/pull/637

deadprogram:
Closing as completed in the most recent release.

PR #637, opened 2024-01-11
wioterminal: fix check for bad Wifi connect

scottfeldman:
rpc_wifi_connect() will return 0 on successful connection, so check for that.  Also, check that if passphrase is given, that it's the minimum 8 chars per WPA Wifi.


scottfeldman:
Fixes https://github.com/tinygo-org/drivers/issues/636

deadprogram:
Thanks for the fix @scottfeldman now merging.

PR #639, opened 2024-01-14
`sd` package: `sdcard` package redesign

soypat:
New Features:
* Idiomatic CSD V1,V2 support with V3 API compatibility
* No heap allocations
* Complete error handling
* CRC support
* Design for interface compatibility with other SDCard implementations i.e. Using raw SDIO with a Raspberry Pi Pico.
* Added tests for what is testable
* Renamed ReadData to more idiomatic ReadBlock, same with Write methods
* More checks for OOB write/read requests
* Allow multi block writing and multi block reading in a double method interface, making it easier to use.
* Easily enable highest transfer rate available for SDCard using CSD 

~Is still a WIP. CRC7 still does not work and I'm getting errors on ever other ReadBlock call.~


bgould:
Is it possible to make this implement BlockDevice so that it is natively compatible with tinyfs? Skimming `definitions.go`, it looks there are some similar methods in there

https://github.com/tinygo-org/tinyfs/blob/release/tinyfs.go#L43

soypat:
@bgould Will do, but I'll do it through another type that uses the ReadBlocks/WriteBlocks interface set to implement the tinyfs.BlockDevice interface. Already had this in mind as the next addition to `sd`

soypat:
@bgould  I've added a draft of what I mean: https://github.com/tinygo-org/drivers/pull/639/files#diff-92f519b8e01de4c93b473c245b482f67da4e20bfbfbe48aaca799f35e45afb43R38

soypat:
@bgould OK, I've finished the sd.BlockDevice type, seems to be working OK. Check it out and let me know what you think :)

deadprogram:
@soypat would you perhaps be able to add a README with some small explanation about how this works?

For example what is `rustref.go` do? :smile_cat:

soypat:
OK, added a Readme and switched some filenames around to make navigation easier.

deadprogram:
@bgould have you had a chance to try this out yet?

deadprogram:
Paging Dr. @bgould :smile_cat: any chance to try this out?

soypat:
I need this for a class I'll be teaching in 1.5 weeks :)

soypat:
Can we merge this?

deadprogram:
You should add something to `smoketest.sh` right? And maybe just a couple more docs, perhaps borrow some from the `sdcard` driver this is intended to "replace".

Thanks!

sd/spicard.go
```diff
@@ -0,0 +1,558 @@
...
+		return d.read_block_single(dst, startBlockIdx)
+
+	} else if numblocks > 1 {
+		// TODO: implement multi block transaction reading.
+		// Rust code is failing here.
+		blocksize := int(d.blk.size())
+		for i := 0; i < numblocks; i++ {
+			dataoff := i * blocksize
+			d.csEnable(true)
+			_, err := d.read_block_single(dst[dataoff:dataoff+blocksize], int64(i)+startBlockIdx)
```
deadprogram, line 268: Multi-block read addressing bug for non-SDHC cards.
In ReadBlocks, after startBlockIdx <<= 9 for non-SDHC cards (converting block index to byte address), the multi-block loop passes int64(i) + startBlockIdx to read_block_single. This adds a block count to a byte address. It should be int64(i)*int64(blocksize) + startBlockIdx. SDHC (block-addressed) probably works correctly by accident.

sd/responses.go
```diff
@@ -0,0 +1,325 @@
...
+}
+func (r *r6) rca() uint16 {
+	return binary.BigEndian.Uint16(r.data[1:3])
+}
+func (r *r6) CardStatus() status {
+	moveBit := func(b status, from, to uint) status {
+		return (b & (1 << from)) >> from << to
+	}
+	// See 4.9.5 R6 (Published RCA response) of the SD Simplified Specification.
+	s := status(binary.BigEndian.Uint16(r.data[1:5]))
```
deadprogram, line 209: binary.BigEndian.Uint16(r.data[1:5]) reads bytes 1–2 (the RCA field), not bytes 3–4 (the card status bits). Should be Uint16(r.data[3:5]).

sd/definitions.go
```diff
@@ -0,0 +1,585 @@
...
+	100,
+	1000,
+	10000,
+	100000,
+	1000000,
+}
+
+// RateKilobits returns the transfer rate in kilobits per second.
+func (t TransferSpeed) RateKilobits() int64 {
+	return 100 * log10table[t&0b111]
```
deadprogram, line 509: log10table has 7 entries (indices 0–6), but t&0b111 can be 7. Index 7 panics. The rate-unit value 7 is reserved in the spec but could appear on a misbehaving card.

sd/definitions.go
```diff
@@ -0,0 +1,585 @@
...
+}
+
+// AccessTime returns the asynchronous part of the data access time.
+func (t TAAC) AccessTime() (d time.Duration) {
+	return time.Duration(log10table[t&0b111]) * time.Nanosecond
```
deadprogram, line 514: Same as above.

sd/blockdevice.go
```diff
@@ -0,0 +1,231 @@
...
+type BlockDevice struct {
+	card      Card
+	blockbuf  []byte
+	blk       blkIdxer
+	numblocks int64
+}
+
+// ReadAt implements the [io.ReaderAt] interface for an SD card.
+// Reads need not be aligned to block boundaries.
+func (bd *BlockDevice) ReadAt(p []byte, off int64) (n int, err error) {
```
deadprogram, line 65: No bounds checking against device size in this function.

sd/blockdevice.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+		n += copy(p, bd.blockbuf)
+	}
+	return n, nil
+}
+
+// WriteAt implements the [io.WriterAt] interface for an SD card. Writes need
+// not be aligned to block boundaries: partial blocks are read, modified and
+// written back.
+func (bd *BlockDevice) WriteAt(p []byte, off int64) (n int, err error) {
```
deadprogram, line 108: No bounds checking against device size in this function either.

soypat:
Found a huge number of bugs after checking against Zephyr/Linux reference and also added EraseBlock and multiblock read/write optimization based on these. Ready for another review!

soypat:
Ready for merge! @deadprogram

deadprogram:
Thanks for the addition @soypat now squash/merging.

PR #642, opened 2024-01-22
netdev: fix concurrency issues with multiple sockets

scottfeldman:
The wifinina driver was not handling multiple concurrent sockets correctly.  This PR fixes the wifinina driver and adds an example/net/snake test for testing multiple concurrent sockets.  The wioterminal driver (rtl8270dn) needed no changes to pass the snake test.

### DIFF
--- a/examples/net/snake/main.go
+++ b/examples/net/snake/main.go
@@ -1,0 +1,133 @@
+// This example is the classic snake network test.  The snake is feed a steady
+// diet of pkts and the pkts work themselves thru the snake segments and exit
+// the tail.  Each snake segment is a TCP socket connection to a server.  The
+// server echos pkts received back to the snake, and serves each segment on a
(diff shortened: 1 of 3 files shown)


examples/net/snake/main.go
```diff
@@ -0,0 +1,133 @@
...
+	"net"
+	"strings"
+	"time"
+)
+
+var (
+	server string = "10.0.0.100:8080"
+)
+
+// Boa Constrictor, a poem by Shel Silverstein
```
deadprogram: As much as I enjoy the poem, I do not think we have usage rights to include this.

deadprogram:
I tested these changes and they appear to be working as expected.

deadprogram:
Rebased to squash the last 3 commits together, then merged to `dev`. Thanks for the fixes @scottfeldman

PR #643, opened 2024-01-22
remove pico from examples/net/tcpclient

scottfeldman:
Not ready for pico yet.  Somehow this slipped in an earlier commit.


deadprogram:
Thanks for the cleanup @scottfeldman now merging.

Issue #646, opened 2024-01-26
hd44780 not working in 4-bit mode with the LCD from Arduino Starter Kit


sorf:
The code exhibiting the problem is pretty much a hello-world type (from [here](https://github.com/sorf/tinygo-arduino/blob/2c4998409486ee85528d50ae6492eb8ef21ad61e/countdown-lcd-1602-4bits/main.go))

``` go
package main

import (
	"bytes"
	m "machine"
	"strconv"
	"time"

	"tinygo.org/x/drivers/hd44780"
)

func main() {
	m.Serial.Configure(m.UARTConfig{BaudRate: 9600})
	lcd, err := hd44780.NewGPIO4Bit([]m.Pin{m.D5, m.D4, m.D3, m.D2}, m.D11, m.D12, m.NoPin)
	if err != nil {
		println("error: create LCD", err.Error())
		return
	}
	if err := lcd.Configure(hd44780.Config{Width: 16, Height: 2}); err != nil {
		println("error: configure LCD", err.Error())
		return

	}

	println("Start")
	lcd.Write([]byte("Countdown:"))
	lcd.Display()
	time.Sleep(1 * time.Second)
	
	// snip
```

It happens with latest version (`v0.26.0`), on both:
- the LCD coming with the [Arduino starter kit](https://store-usa.arduino.cc/products/arduino-starter-kit-multi-language?selectedStore=us) (datasheet: [LCDscreen.PDF](https://wiki-content.arduino.cc/documents/datasheets/LCDscreen.PDF)), and 
- the standard variant of the [virtual LCD on wokwi.com](https://docs.wokwi.com/parts/wokwi-lcd1602)

The easiest way to reproduce the issue is with a virtual circuit on [wokwi.com](https://wokwi.com/) (such as this [one](https://wokwi.com/projects/387688265780407297)), by pressing `F1` there, then selecting `Upload Firmware and Start Simulation` and uploading the `tinygo` built hex file. The mis-behavior is of random "gibberish" showing up on the LCD.

Th expected behavior of the virtual circuit should be like exhibited by this [one](https://wokwi.com/projects/387681602763182081) - programmed with the [Arduino LiquidCrystal library](https://www.arduinolibraries.info/libraries/liquid-crystal). 

Possibly related to #380?

sorf:
[...]

PR #647, opened 2024-01-26
hd44780: configure and write fixes

sorf:
- Write only the high-part of the byte when configuring in 4-bit mode similar to the LiquidCrystal library around
[here](https://github.com/arduino-libraries/LiquidCrystal/blob/1.0.7/src/LiquidCrystal.cpp#L116)

- Pulse enable after writing data similar to the LiquidCrystal library:
[4-bits here](https://github.com/arduino-libraries/LiquidCrystal/blob/1.0.7/src/LiquidCrystal.cpp#L312), [8-bits here](https://github.com/arduino-libraries/LiquidCrystal/blob/1.0.7/src/LiquidCrystal.cpp#L320)

Addresses issue [#646](https://github.com/tinygo-org/drivers/issues/646).

### DIFF
--- a/hd44780/gpio.go
+++ b/hd44780/gpio.go
@@ -2,8 +2,17 @@ package hd44780
 
 import (
 	"errors"
+	"time"
 
 	"machine"
+)
+
+type WriteByteType int
+
+const (
+	FullByte   WriteByteType = 3
+	HighNibble WriteByteType = 2
+	LowNibble  WriteByteType = 1
 )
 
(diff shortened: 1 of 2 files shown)


0pcom:
I tested this as requested and was still getting panic with my setup and this particular 16X1 VFD. I posted my workaround in #646
the panic i was experiencing is related to `lcd.Display()` however and is perhaps unrelated to the fixes in this PR

PR #648, opened 2024-01-26
Rotary encoder support

bgould:
This is a driver for typical rotary encoders based on:

https://gist.github.com/aykevl/3fc1683ed77bb0a9c07559dfe857304a

Not sure if the package name I've chosen is the best, I'm open to suggestions to change it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![PkgGoDev](https://pkg.go.dev/badge/tinygo.org/x/drivers)](https://pkg.go.dev/tinygo.org/x/drivers) [![Build](https://github.com/tinygo-org/drivers/actions/workflows/build.yml/badge.svg?branch=dev)](https://github.com/tinygo-org/drivers/actions/workflows/build.yml)
 
 
-This package provides a collection of 101 different hardware drivers for devices such as sensors and displays that can be used together with [TinyGo](https://tinygo.org).
+This package provides a collection of 102 different hardware drivers for devices such as sensors and displays that can be used together with [TinyGo](https://tinygo.org).
 
 For the complete list, please see:
 https://tinygo.org/docs/reference/devices/
--- a/encoders/quadrature_device.go
+++ b/encoders/quadrature_device.go
@@ -1,0 +1,34 @@
+package encoders
+
+type QuadratureDevice struct {
+	cfg  QuadratureConfig
+	impl quadratureImpl
+}
+
+type QuadratureConfig struct {
+	Precision int
+}
+
+type quadratureImpl interface {
+	configure(cfg QuadratureConfig) error
+	readValue() int
+	writeValue(int)
+}
+
+func (enc *QuadratureDevice) Configure(cfg QuadratureConfig) error {
+	if cfg.Precision < 1 {
+		cfg.Precision = 4
+	}
+	enc.cfg = cfg
+	return enc.impl.configure(cfg)
+}
+
+// Position returns the stored int value for the encoder
+func (enc *QuadratureDevice) Position() int {
+	return enc.impl.readValue() / enc.cfg.Precision
+}
+
+// SetPosition overwrites the currently stored value with the specified int value
+func (enc *QuadratureDevice) SetPosition(v int) {
+	enc.impl.writeValue(v * enc.cfg.Precision)
+}
--- a/encoders/quadrature_interrupt.go
+++ b/encoders/quadrature_interrupt.go
@@ -1,0 +1,69 @@
+//go:build tinygo && (rp2040 || stm32 || k210 || esp32c3 || nrf || (avr && (atmega328p || atmega328pb)))
+
+// Implementation based on:
+// https://gist.github.com/aykevl/3fc1683ed77bb0a9c07559dfe857304a
+
+// Note: build constraints in this file list targets that define machine.PinToggle.
+// If this is supported for additional targets in the future, they can be added above.
+
+package encoders
+
+import (
+	"machine"
+	"runtime/volatile"
+)
+
+var (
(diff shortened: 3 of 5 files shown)


deadprogram:
> Not sure if the package name I've chosen is the best, I'm open to suggestions to change it.

Might as well just name it `rotary_encoder` to make Go happy.

Thanks for adding this @bgould

soypat:
I'd suggest naming the package `encoders` and adding this as a `QuadratureViaInterrupt` type (or similar name) to identify what kind of encoder it is  and what it's working principle is. This would also allow future additions of other encoder types under a single, easy to find package. I'd also name the Config type `QuadratureConfig`.

rotary-encoder/rotary_encoder.go
```diff
@@ -0,0 +1,66 @@
...
+type Config struct {
+	Precision int
+}
+
+func (enc *Device) Configure(cfg Config) {
+	enc.pinA.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
+	enc.pinA.SetInterrupt(machine.PinRising|machine.PinFalling, enc.interrupt)
+
+	enc.pinB.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
+	enc.pinB.SetInterrupt(machine.PinRising|machine.PinFalling, enc.interrupt)
```
aykevl: Please use `PinToggle` instead of `PinRising|PinFalling`. This bitwise operation is not allowed on most platforms, only on AVR, K210, and stm32 (and rp2040 I guess).

(I see it's not supported on rp2040, so feel free to make a PR for that).

examples/rotary-encoder/main.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func main() {
+
+	enc.Configure(rotary_encoder.Config{
+		Precision: 4,
+	})
+
+	for oldValue := 0; ; {
+		time.Sleep(100 * time.Microsecond) // doesn't work without this?
```
aykevl: I suspect this is because you haven't used volatile (see below).

rotary-encoder/rotary_encoder.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+type Device struct {
+	pinA machine.Pin
+	pinB machine.Pin
+
+	precision int
+
+	oldAB int
+	value int
```
aykevl: `value` needs to be accessed using volatile operations. Please use the runtime/volatile package for this. Without it, you'll run into undefined behavior (this is probably why you needed the `time.Sleep` function).

(Yes, this means the gist and the original example on circuitsathome.com are both subtly incorrect).

rotary-encoder/rotary_encoder.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if bHigh {
+		enc.oldAB |= 1
+	}
+	enc.value.Reg += uint32(states[enc.oldAB&0x0f])
+	// enc.value += int(states[enc.oldAB&0x0f])
+}
+
+func (enc *Device) Value() int {
+	return int(enc.value.Reg) / enc.precision
```
aykevl: This doesn't help. You need to call the `.Get()` and `.Set(value)` methods to access the value in a volatile way.

I realize there isn't a lot of documentation on that package, but this might help: https://tinygo.org/docs/concepts/compiler-internals/volatile/

bgould: Ok cool thanks, I was going to ask about that because the program still wasn't working without the `time.Sleep` in there

bgould: Switching to use the Get/Set methods worked, the program functions as expected without the `time.Sleep`

rotary-encoder/rotary_encoder.go
bgould: Added `machine.PinToggle` definition for RP2040 in https://github.com/tinygo-org/tinygo/pull/4104

encoders/device.go
```diff
@@ -0,0 +1,34 @@
+package encoders
+
+type Device struct {
```
soypat: Why do we abstract away the underlying encoder into a Device type? Seems like a big ask to say "all future encoders will behave this way". In fact I'm not sure I'd be able to implement this interface with the encoders I've worked with (half arc+small slits encoder)

rotary-encoder/rotary_encoder.go
bgould: Updated to use `machine.PinToggle` as suggested

encoders/device.go
soypat: A suggestion: Say we implement `QuadratureRotary` type- it could implement the methods `Value` and `SetValue`, which then could be used by users who define the `Encoder` interface:

```go
type Encoder interface {
     Value() int32
     SetValue(int32) error
}
```
This way the code is simpler and users are able to choose their level of abstraction. As a forethought, I would not rush into defining the interface on our side just yet- let users experiment and see what works best!

As a final comment I also suggest using more evocative method names: `Ticks`/`SetTicks`? Or `Position`/`SetPosition`? `Increments`/`SetIncrements`? I feel these all convey more domain specific ideas than `Value`.

bgould: Ok good point ... I agree with what you're saying, I was just thinking of the `Device` as rotary device.  The abstraction was meant to allow for implementing different methods for counting and debouncing the position.

bgould: @soypat  I've updated the naming according to your suggestions

soypat (approved):
@bgould Thank you for your contribution- looks great!

deadprogram:
Thanks for adding this @bgould and to @soypat for all the feedback. Now merging.

PR #649, opened 2024-01-27
rtl8720dn, wifinina: use drivers package version as the driver version

deadprogram:
This PR modifies the `rtl8720dn` and `wifinina` implementation to show the overall drivers package version as the driver version, instead of using their own.


scottfeldman (approved):
LGTM

deadprogram:
Thanks for review @scottfeldman now merging.

PR #655, opened 2024-02-21
ssd1306: improvements needed for Thumby SPI display

deadprogram:
This PR modifies the ssd1306 with improvements needed for Thumby SPI display.

### DIFF
--- a/ssd1306/ssd1306.go
+++ b/ssd1306/ssd1306.go
@@ -13,6 +13,8 @@ import (
 	"tinygo.org/x/drivers/internal/legacy"
 )
 
+type ResetValue [2]byte
+
 // Device wraps I2C or SPI connection.
 type Device struct {
 	bus        Buser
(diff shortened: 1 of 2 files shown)


ssd1306/ssd1306.go
```diff
@@ -30,6 +32,13 @@ type Config struct {
 	Height   int16
 	VccState VccMode
 	Address  uint16
+	// ResetCol and ResetPage are used to reset the screen to 0x0
+	// This is useful for some screens that have a different size than 128x64
+	// For example, the Thumby's screen is 72x40
+	// The default values are normally set automatically based on the size.
+	// If you're using a different size, you might need to set these values manually.
+	ResetCol  [2]byte
```
rminnich: you night consider a type.
You can do all the same stuff, https://go.dev/play/p/Yz_Bon3epiC, but what's nice is the zero value works for your test, and you can easily change it later, without having to find every instance of [2]byte{} in your code.

deadprogram: > you night consider a type.

Good idea, I have done that now.

> you can easily change it later, without having to find every instance of [2]byte{} in your code.

I decided to just change it now anyhow,

bgould:
Looks good to me, thank you @deadprogram , merging

PR #46, opened 2019-04-26
adxl345: update ADXL345 driver and example to make uniform

deadprogram:
This PR updates the ADXL345 driver and example to make it uniform with other accelerometers.

PR #656, opened 2024-02-22
Use int16 for ADXL345 readings

nmaggioni:
The ADXL345 accelerometer outputs readings as 16-bit integers (see [the first page of its datasheet][datasheet]), and thus parsing them as 32-bit ones skews the results.

Here is [the example code][example] running on an RP2040 with the [current implementation][current] and giving out of range readings (the sensor is sitting facing downwards and slightly tilted towards the Y axis):
```
X: 20 Y: 448 Z: 261212
X (raw): 5 Y (raw): 112 Z (raw): 65303
```

And here with proper typecasting, with higher values now in the expected µg (micro-gravity) range:
```
X: 12 Y: 432 Z: -940
X (raw): 3 Y (raw): 108 Z (raw): -235
```

Given the comments in the source code itself, I guess that this driver porting error was previously masked by the limited bitness of other microcontrollers?

~~While I was at it I deleted the error return param from ReadAcceleration since I didn't see anything prone to throwing in its codepath, but I'm wondering if https://github.com/tinygo-org/drivers/pull/46#issue-437786421 is still applicable and it should be kept instead.~~ CI gave me the answer :)

[datasheet]: https://www.sparkfun.com/datasheets/Sensors/Accelerometer/ADXL345.pdf
[example]: https://github.com/tinygo-org/drivers/blob/308763f5007acbf05ba7dd147082ad962ba006bb/examples/adxl345/main.go
[current]: https://github.com/tinygo-org/drivers/blob/308763f5007acbf05ba7dd147082ad962ba006bb/adxl345/adxl345.go#L192

### DIFF
--- a/adxl345/adxl345.go
+++ b/adxl345/adxl345.go
@@ -92,12 +92,12 @@ func (d *Device) Restart() {
 // it in µg (micro-gravity). When one of the axes is pointing straight to Earth
 // and the sensor is not moving the returned value will be around 1000000 or
 // -1000000.
-func (d *Device) ReadAcceleration() (x int16, y int16, z int16, err error) {
+func (d *Device) ReadAcceleration() (x int32, y int32, z int32, err error) {
 	rx, ry, rz := d.ReadRawAcceleration()
 
-	x = d.dataFormat.convertToIS(rx)
-	y = d.dataFormat.convertToIS(ry)
-	z = d.dataFormat.convertToIS(rz)
+	x = int32(d.dataFormat.convertToIS(rx))
+	y = int32(d.dataFormat.convertToIS(ry))
+	z = int32(d.dataFormat.convertToIS(rz))
 
 	return
 }
--- a/adxl345/adxl345.go
+++ b/adxl345/adxl345.go
@@ -92,7 +92,7 @@ func (d *Device) Restart() {
 // it in µg (micro-gravity). When one of the axes is pointing straight to Earth
 // and the sensor is not moving the returned value will be around 1000000 or
 // -1000000.
-func (d *Device) ReadAcceleration() (x int32, y int32, z int32, err error) {
+func (d *Device) ReadAcceleration() (x int16, y int16, z int16, err error) {
 	rx, ry, rz := d.ReadRawAcceleration()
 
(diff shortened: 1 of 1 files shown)


adxl345/adxl345.go
```diff
@@ -92,7 +92,7 @@ func (d *Device) Restart() {
 // it in µg (micro-gravity). When one of the axes is pointing straight to Earth
 // and the sensor is not moving the returned value will be around 1000000 or
 // -1000000.
-func (d *Device) ReadAcceleration() (x int32, y int32, z int32, err error) {
+func (d *Device) ReadAcceleration() (x int16, y int16, z int16, err error) {
```
deadprogram: It would be better to not change the signature of the exported functions, so they are consistent with others devices in the drivers package.

Instead perhaps just use casting from the internal functions?

nmaggioni: Roger, done.

graugans: Coming from a C world where an int was supposed to be  the best "native" length of a CPU. In go this seems to apply as-well [see](https://go.dev/tour/basics/11)

> The int, uint, and uintptr types are usually 32 bits wide on 32-bit systems and 64 bits wide on 64-bit systems. When you need an integer value you should use int unless you have a specific reason to use a sized or unsigned integer type. 

So I would assume that the signature is `int` but maybe this is due to the fact that there is support for the old 8 bit CPUS....

Back in the day when I did a bunch of DSP work, there was overhead when not using the native data type caused by shifting variables in registers. This may all fells apart when there are 8bit integers.

deadprogram:
Thanks for the fixes @nmaggioni now going to squash/merge.

aykevl:
> Given the comments in the source code itself, I guess that this driver porting error was previously masked by the limited bitness of other microcontrollers?

No, integers are all the same on every microcontroller. The only one that might be different are `int`, `uint`, and `uintptr`: `int` and `uint` are either 32-bit or 64-bit (but almost always 32-bit on microcontrollers), `uintptr` is the size of a pointer but is rarely used.

(Just wanted to clear something up that might cause confusion).

PR #659, opened 2024-02-27
tests: use one less core than all available for 'breathing' room.

deadprogram:
This PR improves the test runner to use one less core than all available for 'breathing' room.
It also modifies the behavior to not return usage info for normal operational errors to avoid confusion when debugging failed test runs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
         # We're not on a multi-user machine, so this is safe.
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: TinyGo version check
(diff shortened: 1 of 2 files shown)


loraxipam:
Is there *always* >1 CPU?

deadprogram:
@loraxipam on the machine doing the compiling, yes.

aykevl:
> This PR improves the test runner to use one less core than all available for 'breathing' room.

How does this help? Unless it runs out of RAM, using all cores shouldn't be a problem.
(In fact, `ninja` uses 2 more threads than available cores to allow slightly better scheduling on some operating systems, for example when a thread is blocked on I/O).

I'd also be more comfortable with making sure there's at least 1 goroutine. While single-core CPUs are rare nowadays, it's trivial to make sure they'll continue to work well.

(I also run `make smoke-test` myself sometimes so it would be unfortunate to slow that one down).

PR #661, opened 2024-02-29
encoders: add atsamd21, atsamd51, atsame5x

sago35:
This PR adds support for the SAM. I've tested it on feather-m4 and feather-m0.


deadprogram:
Thanks @sago35 now merging.

PR #663, opened 2024-03-01
pixel: add support for Monochrome types such as the SSD1306

deadprogram:
This PR extends the `pixel` package to add support for  a new type `Monochrome` for displays such as the SSD1306.

### DIFF
--- a/pixel/image.go
+++ b/pixel/image.go
@@ -22,12 +22,13 @@ func NewImage[T Color](width, height int) Image[T] {
 	}
 	var zeroColor T
 	var data unsafe.Pointer
-	if zeroColor.BitsPerPixel()%8 == 0 {
+	switch {
+	case zeroColor.BitsPerPixel()%8 == 0:
 		// Typical formats like RGB888 and RGB565.
 		// Each color starts at a whole byte offset from the start.
 		buf := make([]T, width*height)
 		data = unsafe.Pointer(&buf[0])
-	} else {
+	default:
 		// Formats like RGB444 that have 12 bits per pixel.
 		// We access these as bytes, so allocate the buffer as a byte slice.
 		bufBits := width * height * zeroColor.BitsPerPixel()
@@ -80,10 +81,11 @@ func (img Image[T]) Len() int {
 func (img Image[T]) RawBuffer() []uint8 {
 	var zeroColor T
 	var numBytes int
-	if zeroColor.BitsPerPixel()%8 == 0 {
+	switch {
+	case zeroColor.BitsPerPixel()%8 == 0:
 		// Each color starts at a whole byte offset.
 		numBytes = int(unsafe.Sizeof(zeroColor)) * int(img.width) * int(img.height)
-	} else {
+	default:
 		// Formats like RGB444 that aren't a whole number of bytes.
 		numBits := zeroColor.BitsPerPixel() * int(img.width) * int(img.height)
 		numBytes = (numBits + 7) / 8 // round up (see NewImage)
@@ -99,7 +101,20 @@ func (img Image[T]) Size() (int, int) {
 func (img Image[T]) setPixel(index int, c T) {
 	var zeroColor T
 
-	if zeroColor.BitsPerPixel()%8 == 0 {
+	switch {
+	case zeroColor.BitsPerPixel() == 1:
+		// Monochrome.
+		x := int16(index) % img.width
+		y := int16(index) / img.width
+		offset := x + (y/8)*img.width
+		ptr := (*byte)(unsafe.Add(img.data, offset))
+		if c != zeroColor {
+			*((*byte)(ptr)) |= 1 << uint8(y%8)
+		} else {
+			*((*byte)(ptr)) &^= 1 << uint8(y%8)
+		}
+		return
+	case zeroColor.BitsPerPixel()%8 == 0:
 		// Each color starts at a whole byte offset.
 		// This is the easy case.
 		offset := index * int(unsafe.Sizeof(zeroColor))
@@ -147,7 +162,15 @@ func (img Image[T]) Get(x, y int) T {
 	var zeroColor T
 	index := y*int(img.width) + x // index into img.data
 
-	if zeroColor.BitsPerPixel()%8 == 0 {
+	switch {
+	case zeroColor.BitsPerPixel() == 1:
+		// Monochrome.
+		var c Monochrome
+		offset := x + (y/8)*int(img.width)
+		ptr := (*byte)(unsafe.Add(img.data, offset))
+		c = (*ptr >> uint8(y%8) & 0x1) == 1
+		return any(c).(T)
+	case zeroColor.BitsPerPixel()%8 == 0:
 		// Colors like RGB565, RGB888, etc.
 		offset := index * int(unsafe.Sizeof(zeroColor))
 		ptr := unsafe.Add(img.data, offset)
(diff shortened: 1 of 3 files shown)


pixel/image.go
```diff
@@ -22,12 +22,17 @@ func NewImage[T Color](width, height int) Image[T] {
 	}
 	var zeroColor T
 	var data unsafe.Pointer
-	if zeroColor.BitsPerPixel()%8 == 0 {
+	switch {
+	case zeroColor.BitsPerPixel() == 1:
+		// Monochrome.
+		buf := make([]T, width*height/8)
+		data = unsafe.Pointer(&buf[0])
```
aykevl: I'm pretty sure this is handled correctly in the default case below.

pixel/image.go
```diff
@@ -80,10 +85,14 @@ func (img Image[T]) Len() int {
 func (img Image[T]) RawBuffer() []uint8 {
 	var zeroColor T
 	var numBytes int
-	if zeroColor.BitsPerPixel()%8 == 0 {
+	switch {
+	case zeroColor.BitsPerPixel() == 1:
+		// Monochrome.
+		numBytes = int(img.width) * int(img.height) / 8
```
aykevl: Same here.

pixel/image.go
```diff
@@ -99,7 +108,21 @@ func (img Image[T]) Size() (int, int) {
...
-	if zeroColor.BitsPerPixel()%8 == 0 {
+	switch {
+	case zeroColor.BitsPerPixel() == 1:
+		// Monochrome.
+		x := int16(index) % img.width
+		y := int16(index) / img.width
+		offset := x + (y/8)*img.width
+		ptr := (*byte)(unsafe.Add(img.data, offset))
+		rgb := c.RGBA()
+		if rgb.R != 0 {
```
aykevl: Perhaps this is already optimized away by the compiler, but in general a comparison with zero will be most efficient:
```suggestion
		var zeroColor T
		if c != zeroColor {
```

pixel/image.go
```diff
@@ -99,7 +108,21 @@ func (img Image[T]) Size() (int, int) {
 func (img Image[T]) setPixel(index int, c T) {
 	var zeroColor T
 
-	if zeroColor.BitsPerPixel()%8 == 0 {
+	switch {
+	case zeroColor.BitsPerPixel() == 1:
+		// Monochrome.
+		x := int16(index) % img.width
+		y := int16(index) / img.width
+		offset := x + (y/8)*img.width
```
aykevl, line 109: I don't follow. Isn't the following more efficient, or is there a reason why you picked the above form?

```suggestion
		offset := index / 8
```

(Note that especially divisions are very slow, though divisions by a constant power of two are very fast).

pixel/image.go
```diff
@@ -99,7 +108,21 @@ func (img Image[T]) Size() (int, int) {
...
+		ptr := (*byte)(unsafe.Add(img.data, offset))
+		rgb := c.RGBA()
+		if rgb.R != 0 {
+			*((*byte)(ptr)) |= 1 << uint8(y%8)
+		} else {
+			*((*byte)(ptr)) &^= 1 << uint8(y%8)
```
aykevl, line 114: Here I'm pretty sure you can also use `uint8(index % 8)`.

pixel/image.go
```diff
@@ -181,8 +212,22 @@ func (img Image[T]) Get(x, y int) T {
...
+	case zeroColor.BitsPerPixel() == 1:
+		// Monochrome.
+		numBytes := int(img.width) * int(img.height) / 8
+		for i := 0; i < numBytes; i++ {
+			// TODO: this can be optimized a lot.
+			// - The store can be done as a 32-bit integer, after checking for
+			//   alignment.
+			// - Perhaps the loop can be unrolled to improve copy performance.
+			ptr := (*byte)(unsafe.Add(img.data, i))
+			*((*byte)(ptr)) = 0xff
```
aykevl: This is not correct. This code is supposed to fill the color `color` but instead it just fills the entire buffer with 1-bits (meaning that it would be impossible to make the screen black).
I would suggest something like this:
```go
var colorByte uint8
var zeroColor T
if color != zeroColor {
    colorByte = 0xff
}
```
and then you can use `colorByte` instead of 0xff.

pixel/pixel.go
```diff
@@ -202,6 +204,29 @@ func (c RGB444BE) RGBA() color.RGBA {
...
+func (c MonochromeVertical) RGBA() color.RGBA {
+	value := uint8(0)
+	if c {
+		value = 255
+	}
+	return color.RGBA{
+		R: value,
+		G: value,
+		B: value,
+		A: value,
```
aykevl: The alpha value is always 255 (displays generally aren't transparent).

```suggestion
		A: 255,
```

pixel/pixel.go
```diff
@@ -202,6 +204,29 @@ func (c RGB444BE) RGBA() color.RGBA {
 	return color
 }
 
+type MonochromeVertical bool
+
+func NewMonochromeVertical(r, g, b uint8) MonochromeVertical {
+	return MonochromeVertical((r > 0) || (g > 0) || (b > 0))
+}
```
aykevl: I think we should be using a consistent formula for this purpose, for example the one I used here: https://github.com/tinygo-org/drivers/pull/616. It makes dark-but-not-quite black colors actually black, instead of white.

pixel/pixel.go
```diff
@@ -202,6 +204,29 @@ func (c RGB444BE) RGBA() color.RGBA {
 	return color
 }
 
+type MonochromeVertical bool
```
aykevl: I'm not entirely sure about this name, to be honest. A single pixel doesn't have a direction (it's not horizontal or vertical). And the `Image[T]` type already makes the assumption that pixels are laid out in row major order (so, vertical).
Maybe just `Monochrome` or `Gray1` ("gray 1", because there could also be something like `Gray4` for 16-level grayscale images for example).

pixel/image.go
deadprogram: Updated.

pixel/image.go
deadprogram: Done.

pixel/image.go
deadprogram: Done.

pixel/image.go
deadprogram: At least from my unit testing this change did not pass my unit tests, so I left it as is.

pixel/image.go
deadprogram: Same as above.

pixel/image.go
deadprogram: Good call, done.

pixel/pixel.go
deadprogram: Changed to use this same formula.

pixel/pixel.go
deadprogram: As discussed in Slack, the format for the SSD1306 is in fact vertical. So should the name remain as it is, or be changed?

deadprogram: @aykevl what do you think?

deadprogram: In the end, I decided to go with `Monochrome`

pixel/pixel.go
deadprogram: Done.

deadprogram:
Now merging, thanks for all the feedback @aykevl

PR #665, opened 2024-03-02
ssd1306: Displayer interface improvements

deadprogram:
This PR is to implement the pixel package based interface for the SSD1306 display.

### DIFF
--- a/ssd1306/ssd1306.go
+++ b/ssd1306/ssd1306.go
@@ -11,12 +11,24 @@ import (
 
 	"tinygo.org/x/drivers"
 	"tinygo.org/x/drivers/internal/legacy"
+	"tinygo.org/x/drivers/pixel"
 )
+
+// Pixel formats supported by the ssd1306 driver.
+type Color interface {
+	pixel.MonochromeVertical
+
+	pixel.BaseColor
+}
 
 type ResetValue [2]byte
 
 // Device wraps I2C or SPI connection.
-type Device struct {
+type Device = DeviceOf[pixel.MonochromeVertical]
+
+// DeviceOf is a generic version of Device, which supports different pixel
+// formats.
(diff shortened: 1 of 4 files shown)


aykevl:
Does this display support any other pixel formats than the one monochrome pixel format?
If not, I would suggest not using generics and simply using `DrawBitmap(x, y int16, bitmap pixel.Image[pixel.MonochromeVertical]) error`. This will still match the generic interface in TinyGL.
(If needed, it's relatively simple to change the ssd1306 driver to be generic in the future without breaking backwards compatibility).

ssd1306/ssd1306.go
```diff
@@ -234,16 +260,47 @@ func (d *Device) SetPixel(x int16, y int16, c color.RGBA) {
...
+	d.Command(d.resetCol[0] + uint8(x))
+	d.Command(d.resetCol[0] + uint8(x) + uint8(w) - 1)
+
+	// TODO: handle when the bitmap is not aligned to 8 vertical pixels
+	d.Command(PAGEADDR)
+	d.Command((d.resetPage[0] + uint8(y)) / 8)
+	d.Command((d.resetPage[0] + uint8(y) + uint8(h)) / 8)
+
+	d.Tx(bitmap.RawBuffer(), false)
+	return nil
```
aykevl, line 295: For a prototype it may be easier to go through the bitmap and set it manually:

```go
width, height := bitmap.Size()
for by := 0; by < height; by++ {
    for bx := 0; bx < width; bx++ {
        d.SetPixel(x+bx, y+by, bitmap.Get(bx, by).RGBA())
    }
}
```

It'll be slower than necessary but it has a good chance of actually working.

deadprogram:
Closing since #663 and #671 are doing this better.

PR #672, opened 2024-04-19
ft6336: ignore bogus touch events

eliasnaur:
At least one ft6336 device reports all 255 values from the first read after reset or poweron, even after waiting the specified 300ms reset delay.


deadprogram:
Seems like a good idea to me. Thank you @eliasnaur now merging.

PR #673, opened 2024-04-20
epd2in66b: Waveshare 2.66inch E-Paper Display Module (B) for Raspberry Pi Pico

trichner:
A simple driver for the Waveshare 2.66in EPD for the Pico. Heavily inspired by the official python implementation and its magic timing.

See also: https://www.waveshare.com/Pico-ePaper-2.66-B.htm

Using `machine` package currently breaks the build.
**How should I go about removing the `machine` package from the unit tests?**

Example looks like this:
image


waveshare-epd/epd2in66b/dev.go
```diff
@@ -0,0 +1,329 @@
...
+
+type Device struct {
+	bus  drivers.SPI
+	cs   machine.Pin
+	dc   machine.Pin
+	rst  machine.Pin
+	busy machine.Pin
+
+	width  int16
+	height int16
```
HattoriHanzo031: Do these need to be in the struct since they are not being changed. Constants could be used directly, but I would rename them to something more unique (eg `displayWidth`) to avoid unintentional shadowing.

waveshare-epd/epd2in66b/dev.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+}
+
+func set(buf []byte, bytePos, bitPos int, v bool) {
+	if v {
+		buf[bytePos] |= 0x1 << bitPos
+	} else {
+		buf[bytePos] &^= 0x1 << bitPos
+	}
+}
```
HattoriHanzo031: Maybe split this function into two: `set` and `clear` so you don't have to have bool argument and `if`. It should be a bit faster that way.

deadprogram:
In an optimal scenario, we would add an `spi` test capabilty similar to https://github.com/tinygo-org/drivers/blob/release/tester/i2c.go

In the meantime, using the `drivers.SPI` interface instead of `machine.SPI` can also help.

waveshare-epd/epd2in66b/dev.go
trichner: Ah, good point. Done.

Makefile
```diff
@@ -19,7 +19,7 @@ rwildcard=$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $(subst
 # Recursively find all *_test.go files from cwd & reduce to unique dir names
 HAS_TESTS = $(sort $(dir $(call rwildcard,,*_test.go)))
 # Exclude anything we explicitly don't want to test for whatever reason
-EXCLUDE_TESTS = image
+EXCLUDE_TESTS = image waveshare-epd/epd2in66b
```
trichner, line 22: Alternatively I remove the tests for now, though I found them quite useful for myself.

deadprogram:
Looks like it is missing the example? https://github.com/tinygo-org/drivers/actions/runs/8860923503/job/24332227160?pr=673#step:8:308

trichner:
> Looks like it is missing the example? https://github.com/tinygo-org/drivers/actions/runs/8860923503/job/24332227160?pr=673#step:8:308

Too much copy paste, should be fixed. Sorry about that.

aykevl:
Why is this driver specific to the pico? It looks like a regular driver to me that should work on any chip?

trichner:
> Why is this driver specific to the pico? It looks like a regular driver to me that should work on any chip?

The driver 'should' already work with any other chip. The pins & SPI is all configurable.
I only have the pico HAT to test currently though: https://www.waveshare.com/pico-epaper-2.66.htm

If this looks too specific, I can also remove the pico HAT specific defaults.

What would you prefer? Either works for me.

On a side note: I did not manage to figure out whether this is a specific waveshare driver chip or something more generic.

deadprogram:
> If this looks too specific, I can also remove the pico HAT specific defaults.

It would be better to make them more generic.

To make life easier for devs, you could also add a file protected by build tag, with the specific values for the board you are using. That way the "default" case for your board "just works".

@trichner what do you think?

trichner:
> > If this looks too specific, I can also remove the pico HAT specific defaults.
> 
> It would be better to make them more generic.
> 
> To make life easier for devs, you could also add a file protected by build tag, with the specific values for the board you are using. That way the "default" case for your board "just works".
> 
> @trichner what do you think?

Sounds good! I gave it a try with build flags, was this what you had in mind?

examples/waveshare-epd/epd2in66b/main.go
```diff
@@ -0,0 +1,84 @@
...
+	time.Sleep(2 * time.Second)
+
+	machine.SPI1.Configure(machine.SPIConfig{
+		Frequency: epd2in66b.Baudrate,
+	})
+
+	println("started")
+
+	// in case you have a Pico module, you can directly use
+	// dev, err := epd2in66b.NewPicoModule()
```
trichner, line 31: I've kept the example for the generic setup with this hint for the specific `pico` use case.

waveshare-epd/epd2in66b/dev_pico.go
```diff
@@ -0,0 +1,34 @@
+//go:build pico
```
trichner, line 1: obviously only works for picos ;)

trichner:
@deadprogram let me know if anything else is missing or needs change.

waveshare-epd/epd2in66b/dev.go
```diff
@@ -0,0 +1,292 @@
...
+	// 0 == blank, 1 == red
+	if err := d.sendCommandByte(0x26); err != nil {
+		return err
+	}
+
+	if err := d.sendData(d.redBuffer); err != nil {
+		return err
+	}
+
+	err := d.turnOnDisplay()
```
deadprogram: This function could just end:

```go
return d.turnOnDisplay()
```

waveshare-epd/epd2in66b/dev.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	// set RAM Y-address start / end position
+	ystartLo := byte(ystart)
+	ystartHi := byte(ystart>>8) & 0x1
+
+	yendLo := byte(yend)
+	yendHi := byte(yend>>8) & 0x1
+
+	if err := d.sendCommandSequence([]byte{0x45, ystartLo, ystartHi, yendLo, yendHi}); err != nil {
```
deadprogram:
```go
return d.sendCommandSequence([]byte{0x45, ystartLo, ystartHi, yendLo, yendHi})
```

trichner:
Alright, all cleaned up!

trichner:
Anything left?

trichner:
Ping? :)

deadprogram:
Thanks for adding this @trichner now squash/merging.

PR #674, opened 2024-04-21
ssd1306: add Rotation(), SetRotation() and Sleep() functions to complete Displayer interface

deadprogram:
This PR extends the `ssd1306` to add `Rotation()`, `SetRotation()`, and `Sleep()` functions to complete the `Displayer` interface implementation.

### DIFF
--- a/ssd1306/ssd1306.go
+++ b/ssd1306/ssd1306.go
@@ -15,8 +15,9 @@ import (
 )
 
 var (
-	errBufferSize = errors.New("invalid size buffer")
-	errOutOfRange = errors.New("out of screen range")
+	errBufferSize     = errors.New("invalid size buffer")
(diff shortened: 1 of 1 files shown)


sago35 (approved):
LGTM

After this PR, it seems necessary to define a separate Displayer interface that includes Rotation() and SetRotation() methods. There are already many variations of SetRotation(). Additionally, it might be good to define a separate Displayer interface for Sleep() capability as well.

deadprogram:
The `Displayer` I am implementing here is https://github.com/aykevl/board/blob/main/common.go#L115-L143

And you are very correct, we need to sort this out in a better way in a future PR.

deadprogram:
Thanks for review @sago35 now merging.

PR #675, opened 2024-05-03
Simplify examples/ws2812

sago35:
The WS2812 example had variations due to its dependency on led and the different PIN names used.
With [this PR](https://github.com/tinygo-org/tinygo/pull/4231), boards with WS2812 now have `machine.WS2812` available.
This PR makes the following changes:

* Remove dependency on `machine.LED` (as some variations don't have `machine.LED`)
* Update to use `machine.WS2812` where it is defined


sago35:
My main motivation was to create an example of running WS2812 on waveshare-rp2040-zero. However, adding the new variation required many changes, so I created this PR. Going forward, this example will work without changes as long as machine.WS2812 is defined.

I have tested the following for verification and confirmed that at least the build is successful.

```
tinygo build -o /tmp/out.uf2 -target arduino               ./examples/ws2812/
tinygo build -o /tmp/out.uf2 -target digispark             ./examples/ws2812/
tinygo build -o /tmp/out.uf2 -target qtpy                  ./examples/ws2812/
tinygo build -o /tmp/out.bin -target m5stamp-c3            ./examples/ws2812/
tinygo build -o /tmp/out.uf2 -target thingplus-rp2040      ./examples/ws2812/
tinygo build -o /tmp/out.uf2 -target waveshare-rp2040-zero ./examples/ws2812/
```

examples/ws2812/main.go
```diff
@@ -12,11 +12,12 @@ import (
 	"tinygo.org/x/drivers/ws2812"
 )
 
-var leds [10]color.RGBA
+var (
+	noe  machine.Pin
```
aykevl: This looks like a typo.
Did you test this code?

aykevl (commented):
Seems like a good idea to me!

examples/ws2812/main.go
sago35: There was a typo. Also, the variable "noe" wasn't used at all. I've fixed it to make it more understandable. @aykevl

deadprogram:
Thank you for the improvements @sago35 and to @aykevl for review. Now merging.

PR #677, opened 2024-05-07
uc8151: update to support all functions needed by tinygl and board package Displayer interface

deadprogram:
This PR updates the `uc8151` display driver to support all functions needed by tinygl and board package Displayer interface.

Should allow using Badger2040 board with TinyGL and friends.

### DIFF
--- a/uc8151/registers.go
+++ b/uc8151/registers.go
@@ -1,5 +1,7 @@
 package uc8151
 
+import "tinygo.org/x/drivers"
+
 // Registers
 const (
 	// Display resolution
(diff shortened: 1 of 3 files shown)


uc8151/registers.go
```diff
@@ -141,11 +141,6 @@ const (
 	HZ_100 = 0b00111010
 	HZ_200 = 0b00111001
 
-	NO_ROTATION  Rotation = 0
-	ROTATION_90  Rotation = 1 // 90 degrees clock-wise rotation
-	ROTATION_180 Rotation = 2
-	ROTATION_270 Rotation = 3
-
```
aykevl: That's a breaking change. If you're fine with it that's okay. In other packages I've basically aliased these constants like `NO_ROTATION = drivers.Rotation0`.

uc8151/uc8151.go
```diff
@@ -287,32 +308,43 @@ func (d *Device) ClearBuffer() {
...
-func (d *Device) SetRotation(rotation Rotation) {
+func (d *Device) SetRotation(rotation drivers.Rotation) error {
 	d.rotation = rotation
+	return nil
+}
+
+// Set the sleep mode for this display. SInce it is e-ink, it is not really sleep mode. :)
+func (d *Device) Sleep(sleepEnabled bool) error {
+	return nil
 }
```
aykevl: There is in fact a deep sleep mode, which seems to reduce current quite a bit compared to being on all the time. It doesn't sleep the display, but it does put the controller chip to sleep (I think).
I think it would be a good idea to either implement this or note this in the comment.

aykevl (approved):

uc8151/registers.go
deadprogram: That is probably a good idea. I will do so.

deadprogram: Done.

uc8151/uc8151.go
deadprogram: I think I have implemented this now. PTAL.

deadprogram:
Thanks for review/feedback @aykevl now merging.

PR #680, opened 2024-05-13
Uc8151 Improvements  DrawBitmap and DrawBuffer

conejoninja:


deadprogram:
@conejoninja please rebase against latest `dev`. Thanks!

conejoninja:
Probably the first time I rebase locally, hope I didn't screw up.

deadprogram:
> Probably the first time I rebase locally, hope I didn't screw up.

That was perfect. :+1:

uc8151/uc8151.go
```diff
@@ -201,6 +201,31 @@ func (d *Device) DrawBitmap(x, y int16, bitmap pixel.Image[pixel.Monochrome]) er
 	return nil
 }
 
+// DrawBuffer copies the buffer directly to the screen at the given coordinates (no rotation taken into account). X coordinate and width need to be a multiple of 8.
```
deadprogram, line 204: > (no rotation taken into account)

That seems is a possible issue, since most of the time with this screen we operate at 270 degrees rotation.

aykevl:
`DrawBuffer` is a new interface, why didn't you use the conventional `DrawBitmap` interface? (Or even `DrawRGBBitmap8`)

conejoninja:
`DrawBitmap` already exists and it calls d.SetPixel for each pixel, but I already had the buffer with the data in the correct format, should should avoid a bunch of operations if just copy it instead. It's convenient (in my case)

deadprogram:
> DrawBitmap already exists and it calls d.SetPixel for each pixel

Perhaps you would like to improve the implementation of `DrawBitmap()` instead? :smile_cat:

conejoninja:
> > DrawBitmap already exists and it calls d.SetPixel for each pixel
> 
> Perhaps you would like to improve the implementation of `DrawBitmap()` instead? 😸

Not sure if possible or worthy. 
Each byte of the buffer represent a *row* of 1x8 pixels. Using arbitrary coordinates x,y (ie not a multiple of 8), size and different rotations could be quite complex.

deadprogram:
Just coming back around to this, since the badger2040 repo is using it.

What should we do?

deadprogram:
@conejoninja I think this PR will no longer be required after our efforts of earlier today.

PR #681, opened 2024-05-21
servo: add function SetAngle() to simplify API for most common use case

deadprogram:
This PR modifies the `servo` package to add a new function `SetAngle()` to simplify API for most common use case: move a servo to a particular angle.

### DIFF
--- a/servo/servo.go
+++ b/servo/servo.go
@@ -1,6 +1,12 @@
 package servo
 
-import "machine"
+import (
+	"machine"
+
+	"errors"
+)
+
+var ErrInvalidAngle = errors.New("servo: invalid angle")
 
 // PWM is the interface necessary for controlling typical servo motors.
 type PWM interface {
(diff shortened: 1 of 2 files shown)


servo/servo.go
```diff
@@ -1,6 +1,12 @@
 package servo
 
-import "machine"
+import (
+	"machine"
+
+	"errors"
+)
+
+var ErrInvalidAngle = errors.New("invalid angle")
```
aykevl: Maybe better:

```suggestion
var ErrInvalidAngle = errors.New("servo: invalid angle")
```

servo/servo.go
```diff
@@ -80,3 +86,17 @@ func (s Servo) SetMicroseconds(microseconds int16) {
...
+// SetAngle sets the angle of the servo in degrees. The angle should be between
+// 0 and 180, where 0 is the minimum angle and 180 is the maximum angle.
+func (s Servo) SetAngle(angle int) error {
+	if angle < 0 || angle > 180 {
+		return ErrInvalidAngle
+	}
+
+	// 0° is 1000µs, 180° is 2000µs
+	microseconds := angle*1000/180 + 1000
+	s.SetMicroseconds(int16(microseconds))
```
aykevl: Doesn't this vary between servo motors?
I think this needs some explanation that it varies by servo motor. For example, from Wikipedia:

> The parameters for the pulses are the minimal pulse width, the maximal pulse width, and the repetition rate. Given the rotation constraints of the servo, neutral is defined to be the center of rotation. Different servos will have different constraints on their rotation, but the neutral position is always around 1.5 milliseconds (ms) pulse width.

deadprogram: The explanation for `SetMicroseconds()` already does this, no?

servo/servo.go
deadprogram: Good idea. Done.

servo/servo.go
deadprogram: I added specific reference to `SetMicroseconds()` and some additional info as well. Should be good now.

deadprogram:
Any other feedback here before merge?

deadprogram:
I have addressed all of the comments on this PR, so now merging. Thanks for the feedback @aykevl

PR #682, opened 2024-05-25
pixel: add more tests

aykevl:
I'm trying to figure out a bug. While doing that, I added some tests to make sure all the pixel formats work as expected (the same data should be retrieved from the image as was put in).

Still haven't found the bug though...


aykevl:
Actually I can get this test to fail by increasing the test image size. New PR incoming...

PR #683, opened 2024-05-25
pixel: fix `Image[Monochrome].Set` for larger images

aykevl:
For bigger images, the pixel index might not fit in a int16. Therefore, int is needed during the calculation.

While fixing this bug, I've added a few tests that verify the Image implementation by creating images, filling them with random data, and then checking whether they still contain the same data. This test failed before the patch.


deadprogram:
Thanks for the fix @aykevl looks good. Now merging.

PR #695, opened 2024-07-20
servo: Add function `SetAngleWithMicroseconds`

PWND0U:
SetAngleWithMicroseconds: Adjust the angle by customizing the c ontrol pulse width.


deadprogram:
Seems good for higher precision servos than I normally work with. Now squash/merging, thanks @PWND0U

PR #700, opened 2024-08-14
removed what seems like an accidental home directory replace in go.mod

gherlein:
in go.mod was this line that I propose removing:

replace tinygo.org/x/drivers/mcp9808 => /home/kasterby/Documents/drivers/mcp9808

Looks like an accidental commit of a development local "replace" - probably should not be in the mainline.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module tinygo.org/x/drivers
 
 go 1.18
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Good catch @gherlein :+1: 

Switched branch to `dev` since we will be having a new release very soon anyhow.

PR #701, opened 2024-08-14
expanded README to discuss need to change variables in examples

gherlein:
Expanded the README to explain the need to edit the examples for the specific board being used.  It's obvious if you have some experience, but to someone new coming to tinygo they may not realize this prior to trying to test an example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ go get tinygo.org/x/drivers
 
 ## How to use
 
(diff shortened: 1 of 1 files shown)


deadprogram:
@gherlein good point on the pin mapping, and thanks for the clarifications.

I switched the branch for this PR to `dev` as mentioned here: 

Now merging. Thank you!

PR #702, opened 2024-08-16
ssd1306: Add function `SetFlip` and `GetFlip`

notpop:
# Implementation of the Flip Functions

Hello, I am new to contributing to OSS, so I apologize in advance if there are any mistakes.

## What I did
I added a Flip functions to the `ssd1306` driver.

## Background
While playing with `ssd1306`, I wanted to flip the display, but I found that there was no method implemented for it. I had to directly execute commands within the application. Thus, I thought it would be helpful to implement this functionality in the driver itself.

## Why I didn't extend `SetRotation`
Initially, I thought it would be best to integrate the flip functionality into `SetRotation` to align with the other existing implementations. However, I found that implementing 90° and 270° rotations is quite challenging. To help clarify my point, please refer to the images below:

- [0° image](https://pbs.twimg.com/media/GVFnBXoaEAMOFse?format=jpg&name=4096x4096)
- [90°? image](https://pbs.twimg.com/media/GVFnBXob0AAJ7kt?format=jpg&name=4096x4096)
- [180° image](https://pbs.twimg.com/media/GVFnBXraQAEQfHG?format=jpg&name=4096x4096)
- [270°? image](https://pbs.twimg.com/media/GVFnBXqaEAQhlnh?format=jpg&name=4096x4096)

From my current understanding, it seems that the `ssd1306` hardware does not support these kinds of rotation features directly, so a more complex implementation would be required. However, I believe that even just implementing the flip functionality is useful, so I decided to start with this.

Thank you for your time and consideration.

### DIFF
--- a/ssd1306/registers.go
+++ b/ssd1306/registers.go
@@ -38,4 +38,7 @@ const (
 
 	EXTERNALVCC  VccMode = 0x1
 	SWITCHCAPVCC VccMode = 0x2
+
+	NO_FLIP = false
+	FLIP    = true
 )
--- a/ssd1306/registers.go
+++ b/ssd1306/registers.go
@@ -1,5 +1,7 @@
 package ssd1306
 
+import "tinygo.org/x/drivers"
+
 // Registers
 const (
 	Address        = 0x3D
@@ -39,6 +41,6 @@ const (
 	EXTERNALVCC  VccMode = 0x1
 	SWITCHCAPVCC VccMode = 0x2
 
-	NO_FLIP = false
-	FLIP    = true
+	NO_ROTATION  = drivers.Rotation0
+	ROTATION_180 = drivers.Rotation180
 )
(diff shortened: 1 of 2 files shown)


sago35 (commented):
Thank you for the PR.

It is true that the SSD1306 does not seem to have 90-degree or 270-degree rotation capabilities, so the Flip API seems appropriate.
The only potential issue might be that the default value has been changed.
Originally, it was set so that IsFlip == true. I also found out that the register settings when IsFlip == true differ from the initial values of the SSD1306. However, in terms of compatibility with tinygo.org/x/drivers, I think it would be better not to change this. What do you think? @deadprogram

deadprogram:
I think we would be better off with `SetRotation` like other displays instead of adding another API. The implementation could return an error if you tried to set to a currently unsupported rotation for a particular display.

notpop:
Thank you for your feedback. Based on your suggestions, I have revised the SetRotation function.

deadprogram:
Thank you for making the changes @notpop looks good to me.

I changed the branch for this PR to `dev` as discussed here:

@sago35 or anyone else have any comments before I squash/merge?

sago35 (approved):
LGTM

deadprogram:
Thank you @notpop and @sago35

PR #706, opened 2024-10-02
Onewire improvements

HattoriHanzo031:
Some improvements to OneWire/DS18B20 driver.

Although external pull-up resistor for DS18B20 is recommended, internal pull-up is good enough in some cases (for example for single sensor on a short bus), so low level functions are changed to enable internal pull-up for high state. 

While testing DS18B20 sensor without internal or external pull-up, I got panic because OneWire `Search` function "found" more than 32 devices and overflowed the array. To prevent this, the check is added to stop the search (and return an error) when more than 32 devices are found. Also, CRC check was added to prevent adding invalid devices during search. 

Additionally PR contains suggestion for changing `Crc8` method and in turn `OneWireDevice` interface (which is a breaking change) to be a bit more idiomatic Go. This is only a suggestion and I would like to get other opinions on it.

### DIFF
--- a/ds18b20/ds18b20.go
+++ b/ds18b20/ds18b20.go
@@ -19,7 +19,7 @@ type OneWireDevice interface {
 	Write(uint8)
 	Read() uint8
 	Select([]uint8) error
-	Сrc8([]uint8, int) uint8
+	Сrc8([]uint8) uint8
 }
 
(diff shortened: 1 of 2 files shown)


deadprogram:
Thank you very much for the improvements @HattoriHanzo031 now merging.

PR #709, opened 2024-10-21
touch: add capacitive touch sensing on normal GPIO pins

aykevl:
It is possible to make a capacitive touch button out of any GPIO pin on most MCUs. The input may be somewhat unreliable compared to dedicated hardware, but it probably works well enough for most people. The way to do this is by connecting the GPIO pin to ground through a 1MΩ resistor, charging the pin by setting it high, switching to input, and measuring how long it takes to discharge (through the 1MΩ resistor).

Most of this is handled inside the driver. All you need to do to make this work is to connect an 1 mega-ohm resistor to the GPIO pin and to ground, and to connect some sort of metal surface to the GPIO pin. To avoid ESD issues, it's probably best to cover the metal surface with a thin non-conducting layer (Scotch tape for example).

Tested on the following chips/boards:

  * RP2040 (Raspberry Pi Pico)
  * ATSAMD21 (Adafruit PyBadge)
  * NRF52840 (PCA10056 developer board)
  * ESP8266 (NodeMCU)
  * ATmega328p (Arduino Uno)
  * ESP32C3 (WaveShare ESP-C3-32S-Kit, see https://github.com/tinygo-org/tinygo/pull/4538)
  * FE310 (SiFive HiFive1 rev B, see https://github.com/tinygo-org/tinygo/pull/4539)

There are loads more I could test, but with this list I'm fairly confident it's going to work on nearly any MCU that's running at a high enough clock rate.

The sensitivity threshold in the example may need to be adjusted per board though, the default value of 100 typically recognizes when a cable is being touched but the RP2040 for example is capable of doing much more precise measurements if the power supply is sufficiently noise-free (it can even detect a finger at ~1cm with a large enough pad!)

This capacitive touch driver does detect the amount of noise on the input and tries to adjust for it, but the algorithm can probably be improved still. I'm not all that familiar with DSP algorithms. I might make a PR in the future to improve this if I get around to it.

### DIFF
--- a/examples/touch/capacitive/main.go
+++ b/examples/touch/capacitive/main.go
@@ -1,0 +1,69 @@
+// Capacitive touch sensing example.
+//
+// This capacitive touch sensor works by charging a normal GPIO pin, then slowly
+// discharging it through a 1MΩ resistor and seeing how long it takes to go from
+// high to low.
+//
+// Use as follows:
+//   - Change touchPin below as needed.
+//   - Connect this pin to some metal surface, like a piece of aluminimum foil.
+//     Make sure this surface is covered (using paper, Scotch tape, etc).
+//   - Also connect this same pin to ground through a 1MΩ resistor.
+//
+// This sensor is very sensitive to noise on the power source, so you should
+// probably try to limit it by running from a battery for example. Especially
+// phone chargers can produce a lot of noise.
+package main
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/touch/capacitive"
+)
+
+const touchPin = machine.GP16 // Raspberry Pi Pico
+
+func main() {
+	time.Sleep(time.Second * 2)
+	println("start")
+
+	led := machine.LED
+	led.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	led.Low()
+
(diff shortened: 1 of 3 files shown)


deadprogram:
This PR is basically a better version of https://github.com/tinygo-org/drivers/tree/release/makeybutton which I am certainly willing to deprecate.

aykevl:
Oh I didn't realize we already had a capacitive touch driver!
This new driver is probably much more sensitive, with all the autocalibration built in. And of course with a configurable sensitivity. The one thing it doesn't have is the equivalent of `HighMeansPressed` (though it could be added of course).

deadprogram:
> This new driver is probably much more sensitive, with all the autocalibration built in. And of course with a configurable sensitivity.

I like that a lot better for sure.

> The one thing it doesn't have is the equivalent of HighMeansPressed (though it could be added of course).

Not that big a feature missing, plus I dislike the name I gave that feature anyhow.

Let's merge and mark the other as deprecated?

aykevl:
> Let's merge and mark the other as deprecated?

Sure, sounds good :)
We might even keep both because I think they serve a somewhat different purpose.

deadprogram:
Now merging, thanks for the addition @aykevl

PR #711, opened 2024-10-25
uc8151: add FillRectangle() and SetScroll() functions to satisfy tinyterm.Displayer interface

deadprogram:
This PR adds to the `uc8151` display by implementing the `FillRectangle()` and `SetScroll()` functions in order to satisfy the `tinyterm.Displayer` interface.

### DIFF
--- a/uc8151/uc8151.go
+++ b/uc8151/uc8151.go
@@ -478,3 +478,31 @@ func (d *Device) SetLUT(speed Speed, flickerFree bool) error {
 
 	return nil
 }
+
+// FillRectangle fills a rectangle at a given coordinates with a color
+func (d *Device) FillRectangle(x, y, width, height int16, c color.RGBA) error {
+	dw, dh := d.Size()
+
+	if x < 0 || y < 0 || width <= 0 || height <= 0 ||
+		x >= d.width || (x+width) > dw || y >= dh || (y+height) > dh {
+		return errOutOfRange
+	}
+
+	if x+width == dw && y+height == dh && c.R == 0 && c.G == 0 && c.B == 0 {
+		d.ClearDisplay()
+		return nil
+	}
+
+	for i := x; i < x+width; i++ {
+		for j := y; j < y+height; j++ {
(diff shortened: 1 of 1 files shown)


deadprogram:
Once this can get merged, I have some related changes to tinyterm that allows it to work on the badger2040 display.

ydnar (approved):
Do you want to add an assertion that it satisfies the interface?

deadprogram:
@ydnar what would be the best way to do that? The interface is defined in package that consumes this package here: https://github.com/tinygo-org/tinyterm/blob/release/tinyterm.go#L27

ydnar:
Ah. Nevermind then!

ydnar:
> @ydnar what would be the best way to do that? The interface is defined in package that consumes this package here: https://github.com/tinygo-org/tinyterm/blob/release/tinyterm.go#L27

Can you define it elsewhere and have tinyterm import that and alias it?

ydnar:
Seems like a big lift for little value...

deadprogram:
I'll think about a way to do it in a future PR. Thanks for review @ydnar now merging.

uc8151/uc8151.go
```diff
@@ -478,3 +478,31 @@ func (d *Device) SetLUT(speed Speed, flickerFree bool) error {
...
+		}
+	}
+
+	return nil
+}
+
+// SetScroll sets the vertical scrolling for the display, which is a NOP for this display.
+func (d *Device) SetScroll(line int16) {
+	return
+}
```
aykevl, line 508: You have a program that calls `SetScroll` without first calling `SetScrollArea`? I think that won't work on other displays like ST7789.

deadprogram: This is for interface compatibility with tinyterm which does not call `SetScrollArea()`. See https://github.com/tinygo-org/tinyterm/blob/release/tinyterm.go#L27-L31

Actually, in the case of the `uc8151` tinyterm only uses the software scroll, which is why the implementation here just returns without doing anything.

PR #715, opened 2024-10-28
Pixel from bytes

conejoninja:
Duplicate from #713 because it made the tests fail


deadprogram:
The random noise fails sometimes on the mono image. Might have been better to not revert, but too late now.

conejoninja:
Thanks, I tried about 20-30 times and it failed every time, I didn't know the cause, so I thought since I merged a couple of PRs in different repos, I messed up somewhere along the line. Thanks again for your work.

aykevl:
The random noise fails for monochrome images were in fact a bug, I've fixed it here: https://github.com/tinygo-org/drivers/pull/716

PR #716, opened 2024-10-28
pixel: fix Monochrome setPixel 

aykevl:
Set/setPixel and Get weren't using the same indices. I've taken the ones used in Get and applied them to setPixel too.
This fixes testImageNoise for the monochrome image.

I don't know which of the two behaviors was correct, but the one in `Get` certainly is simpler and more efficient.
@deadprogram do you have a specification of the format somewhere?

### DIFF
--- a/pixel/image.go
+++ b/pixel/image.go
@@ -138,14 +138,14 @@ func (img Image[T]) setPixel(index int, c T) {
 	switch {
 	case zeroColor.BitsPerPixel() == 1:
 		// Monochrome.
-		x := index % int(img.width)
 		offset := index / 8
+		bits := index % 8
 
 		ptr := (*byte)(unsafe.Add(img.data, offset))
 		if c != zeroColor {
-			*((*byte)(ptr)) |= (1 << (7 - uint8(x%8)))
+			*((*byte)(ptr)) |= (1 << (7 - uint8(bits)))
 		} else {
-			*((*byte)(ptr)) &^= (1 << (7 - uint8(x%8)))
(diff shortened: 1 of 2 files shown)


pixel/image.go
```diff
@@ -202,7 +202,7 @@ func (img Image[T]) Get(x, y int) T {
 		// Monochrome.
 		var c Monochrome
 		offset := index / 8
-		bits := index - (offset * 8)
+		bits := index % 8
```
aykevl, line 205: This code looks different, but it doesn't change behavior. It's just a simplification.

pixel/image.go
```diff
@@ -138,14 +138,14 @@ func (img Image[T]) setPixel(index int, c T) {
 	switch {
 	case zeroColor.BitsPerPixel() == 1:
 		// Monochrome.
-		x := index % int(img.width)
 		offset := index / 8
+		bits := index % 8
```
aykevl, line 142: The old behavior was most certainly incorrect.

Take for example an image of size 5x2. The first pixel at (0, 0) is `index` 0 and `x` 0 (make sense).
But at position (0, 1), which is index 5, you _also_ get `index` 0 and `x` 0. So the two pixels are stored at the same location.

conejoninja: exactly, took me a while to discover that the old behavior only works on images with width multiple of 8 :facepalm:

deadprogram:
This is great, thank you very much for fixing this @aykevl 

Tested on badger2040 display, works as expected. Now merging.

PR #717, opened 2024-10-29
bugfix: correct pcf8591 casting of return value from adc

deadprogram:
This PR corrects the `pcf8591` driver, by correcting the casting of return value from adc as it was overflowing the 8 bit result.


conejoninja:
Thanks, merged

PR #720, opened 2024-11-02
Adding support for the LSM303DLHC e-Compass

BryanSouza91:
I have added support for the LSM303DLHC e-Compass using the existing LSM303AGR driver as the starting point. I updated the registers as necessary from the information provided by the datasheet. I also fixed spelling in the connection error message in the LSM303AGR driver.


deadprogram:
Thanks for working on this @BryanSouza91 

Can you please add an example, and also then add that example to the smoke tests? https://github.com/tinygo-org/drivers/blob/dev/smoketest.sh

Thank you!

BryanSouza91:
@deadprogram I finally got around to adding the example and then adding that to the smoketest.

lsm303dlhc/lsm303dlhc.go
```diff
@@ -0,0 +1,232 @@
...
+type Configuration struct {
+	AccelPowerMode uint8
+	AccelRange     uint8
+	AccelDataRate  uint8
+	MagPowerMode   uint8
+	MagSystemMode  uint8
+	MagDataRate    uint8
+}
+
+// commented out "Connected" related lines since the DLHC sensor does not have the WHO_AM_I registers
```
deadprogram: In that case, these lines should be removed instead of commented.

lsm303dlhc/lsm303dlhc.go
```diff
@@ -0,0 +1,232 @@
...
+// This function only creates the Device object, it does not touch the device.
+func New(bus drivers.I2C) *Device {
+	return &Device{
+		bus:          bus,
+		AccelAddress: ACCEL_ADDRESS,
+		MagAddress:   MAG_ADDRESS,
+	}
+}
+
+// Connected returns whether both sensor on LSM303dlhc has been found.
```
deadprogram: Same for all the commented lines...

BryanSouza91: Removed in latest commit.

lsm303dlhc/lsm303dlhc.go
BryanSouza91: I have removed the commented lines that were related to unused functions.

deadprogram:
@BryanSouza91 due to having added the merge commits to this PR, it is not really possible to merge in its current state. Perhaps you can cherry-pick your commits related to the LSM303DLHC into a separate branch, squash them, and then either overwrite the branch in this PR, or create a new PR.

What do you think?

BryanSouza91:
I shouldn't have an issue squashing then overwriting this branch, if I do I will create a new PR.

BryanSouza91:
@deadprogram I don't think that last commit will work... I have [dev-PR branch](/BryanSouza91/drivers/tree/dev-PR) ready to go.

deadprogram:
Closing since was replaced by #783 thank you!

Issue #691, opened 2024-07-02
[HUB75]Leds do not light up as expected


ehime-iyokan:
I wrote program using hub75 package and raspberry pi pico. but, it didn't light up. So I have tried to fix hub75.
As a result. It light up as expected. And Environmental information is here. -> https://github.com/ehime-iyokan/tinygo_hub75_sample

![picture1](https://github.com/tinygo-org/drivers/assets/86559239/314c977d-55a1-4595-b675-c12286e2ff68)

First, I added a `time.sleep(1 * time.Microsecond)` after latch-High.
By doing so, the light came on. However, I don't know why, so if you know, please let me know.
Moreover, the red lights are on in different places.

![picture2](https://github.com/tinygo-org/drivers/assets/86559239/4283d00f-ef09-40bf-90f0-eaa0a4937c60)

Second, I fixed buffering of red data.
I guess it has to do left-shift first. 
`d.buffer[c][offsetR] = d.buffer[c][offsetR] &^ 1 << bitSelect`
-> `d.buffer[c][offsetR] &^= 1 << bitSelect`

![picture3](https://github.com/tinygo-org/drivers/assets/86559239/bda20b58-6380-4190-9c75-f0a4a665f91e)

ehime-iyokan:
`I added a time.sleep(1 * time.Microsecond) after latch-High.
By doing so, the light came on. However, I don't know why`
If I figured out why it worked correctly, I'll create a pull request.

sago35:
@ehime-iyokan 
The GPIO of the RP2040 might be too fast. Without adding time.Sleep(), the pulse width is 110ns, which might be too short. With time.Sleep(), it is around 5us.

I think the fix you are considering is necessary.

## with `time.Sleep(1 * time.Microsecond)` : 5.046us
![image](https://github.com/user-attachments/assets/72691496-cf55-4bd6-8e77-5aee8d52a651)


## without `time.Sleep(1 * time.Microsecond)` : 110ns
![image](https://github.com/user-attachments/assets/f827a7b9-a533-4072-91a6-f2ed2a1440c8)

sago35:
Including other parts as well, it is better to use `delay.Sleep()` for microsecond sleeps. I think it would be good if you could create a Pull Request with that change.

https://github.com/tinygo-org/drivers/blob/v0.28.0/delay/sleep.go


## with `delay.Sleep(1 * time.Microsecond)` : 1.16us

[...]

PR #722, opened 2024-11-04
hub75: fix data buffering

ehime-iyokan:
I fixed buffering of red data. ( https://github.com/tinygo-org/drivers/issues/691#issue-2386396643 )

The red data seems to be misaligned, so the display position is incorrect;
however, the other colors are fine.
```golang
// Sample code for order of operations
package main

import "fmt"

func main() {
	bitSelect := 2
	data_1 := 0x05
	data_2 := 0x05

	fmt.Println("before")
	fmt.Printf("%08b\n", data_1)
	data_1 = data_1 &^ 1 << bitSelect
	fmt.Printf("%08b\n", data_1)

	fmt.Println("after")
	fmt.Printf("%08b\n", data_2)
	data_2 &^= 1 << bitSelect
	fmt.Printf("%08b\n", data_2)
}
```
```
// before: it setting bits to 1 in unintended positions.
// after: it is clearing the bits in the intended positions.
$ go run .    
before
00000101
00010000
after
00000101
00000001
```

### DIFF
--- a/hub75/hub75.go
+++ b/hub75/hub75.go
@@ -167,7 +167,7 @@ func (d *Device) fillMatrixBuffer(x int16, y int16, r uint8, g uint8, b uint8) {
 		if r > colorTresh {
 			d.buffer[c][offsetR] |= 1 << bitSelect
 		} else {
-			d.buffer[c][offsetR] = d.buffer[c][offsetR] &^ 1 << bitSelect
+			d.buffer[c][offsetR] &^= 1 << bitSelect
 		}
 		if g > colorTresh {
 			d.buffer[(c+d.colorThirdStep)%d.colorDepth][offsetG] |= 1 << bitSelect


ehime-iyokan:
I have confirmed sample-program works without any problems. https://github.com/ehime-iyokan/DisplayWeatherInfoOnLEDmatrix/commit/4ec0054991783fed96f22bb0bc48d5af6b15a6dd

ledmatrix

deadprogram:
@conejoninja PTAL

conejoninja:
It's on my list, I need to check it doesn't have any unintended consecuences

conejoninja:
I'm having some hardware issues that doesn't allow me to test this, adding @aykevl just in case I can't review it before next release

ehime-iyokan:
> It's on my list, I need to check it doesn't have any unintended consecuences

I'll try to consider test patterns. @conejoninja 
However, I don't have much experience contributing, so I'd appreciate any advice.

ehime-iyokan:
I have summarized the code fix and test patterns. @conejoninja 
# 🔧 **Details of Fix**

- **Terminology:**
  - The **`if clause`** refers to the processing starting at line 168.
  - The **`else clause`** refers to the processing starting at line 170.

## **🛠️ Fix Implementation**

- The `else clause` should be corrected as follows:  
  ```go
  d.buffer[c][offsetR] &^= 1 << bitSelect
  ```

- This ensures that only the bit at the `bitSelect` position is cleared, leaving other bits unaffected.

## **🔍 Observations**

- **Expected Behavior:**  
  - The `if clause` sets the bit at the `bitSelect` position to `1`.
  - Correspondingly, the `else clause` should set the bit at the `bitSelect` position to `0`.

- **Current Implementation Issue:**  
  - The current code in the `else clause` is as follows:  
    ```go
    d.buffer[c][offsetR] = d.buffer[c][offsetR] &^ 1 << bitSelect
    ```
  - This produces the same result as the following incorrect operation:  
    ```go
    d.buffer[c][offsetR] = (d.buffer[c][offsetR] &^ 1) << bitSelect
    ```
  - This does **not** achieve the desired behavior of clearing the specific bit.

- **Analysis:**  
  - The goal is to set the bit at the position indicated by `bitSelect` to `0`.
  - The current implementation is incorrect and needs to be fixed.

## **🧪 Sample Code for Behavior Verification**

- [Go Playground Sample](https://go.dev/play/p/cVr8fXGKSSi)  
  - **`before`:** Demonstrates behavior before the fix.  
  - **`after`:** Demonstrates behavior after the fix.

---


# 📝 **Test Verification**

## **Verification of Red LED Operation**

- **Test Steps**:  
  - 1. Illuminate all red LEDs.  
  - 2. Turn off a single red LED while others remain on.  
- **Expected Result**:  
  - Only red LEDs should illuminate or turn off as expected, with no impact on green or blue LEDs.

## **No Interference with Other Colors**

- **Test Steps**:  
  - 1. Illuminate red and green LEDs simultaneously.  
  - 2. Illuminate red and blue LEDs simultaneously.  
- **Expected Result**:  
  - Red and green, or red and blue, should display correctly without color interference.

## **Note**
- Due to the complexity of the processing, such as differences in behavior depending on brightness (color depth), additional test patterns may be required.
- It might also be a good idea to check if there are any issues when using equipment other than the ones I used.
  - https://github.com/ehime-iyokan/DisplayWeatherInfoOnLEDmatrix?tab=readme-ov-file#environment
---

# 📋 **Test Results**

| **Test Case** | **Result** | **Notes** |
|------------------------------------|---------------|--------------------------------|
| Red LED Operation | Pending | Red LEDs illuminate correctly  |
| No Interference with Other Colors  | Pending | No impact on green or blue  |
| ? | Pending | ? |

sago35 (approved):
I believe the correctness of this PR is demonstrated by the following code. The intended behavior is to clear a specific bit, but the previous code cleared all bits below it as well.

https://go.dev/play/p/cVr8fXGKSSi

```
before 
11111110
11111100
11111000
11110000
11100000
11000000
10000000
00000000

after
11111110
11111101
11111011
11110111
11101111
11011111
10111111
01111111
```

This is evident when looking at the handling of other colors as well.

image

deadprogram:
Thanks for review @sago35 and to @ehime-iyokan for the fix. Now merging.

PR #725, opened 2024-12-09
Added TMC5160 support

amken3d:
Requesting review of the the PR to add support for TMC5160. The TMC5160 is a popular stepper driver and is feature packed. This implementation add support to using the TMC5160 using either the UART mode or the SPI mode. The code has been tested to work on rp2040. The smoke tests passed using "make test". The code is not fully tested for all available features of the TMC5160 as it is a pretty complicated piece of silicon.


go.mod
```diff
@@ -1,20 +1,23 @@
 module tinygo.org/x/drivers
 
-go 1.18
+go 1.22.1
```
deadprogram, line 3: It would be better to require the minimum version of Go supported by TinyGo here instead. Which would be current version minus 2 e.g. Go 1.21

deadprogram:
Hello @amken3d this is a very impressive PR! Thank you for working on it.

I changed the branch for it to `dev` as mentioned here:

Also, I made a comment, although did not have time to look deeply in the new code.

Also, wondering if you could please add a smoketest/example?

Thanks!

amken3d:
@deadprogram, thank you for the kind words. 
I have added example code and smoke test as per your suggestion. I hope this adds some value to users. Looking forward to many more adventures with Tinygo. 
My go and tinygo journey started about a 2 months ago and I am loving it. My primary reason for switching to go/tinygo is because I am building a Pick and Place machine using go (gocv). I initially wrote the microcontroller code in c++(platformio), but I found it much more ergonomic to use tinygo instead. I do have a couple of gripes we can discuss offline, but overall, I feel like Tinygo is  much easier alternative to Arduino/C/C++. 

Regards
H Keni

go.mod
```diff
@@ -1,20 +1,23 @@
...
+
+toolchain go1.23.3
```
deadprogram: Same comment here as above.

amken3d:
When i do go mod tidy, it automatically upgrades to
go 1.22.1

toolchain go1.23.1

Is there a way to pin it to 1.21? I've tried everything that google suggested and failed

deadprogram:
I think the `toolchain` here does not matter to user, so you were fine to run go mod tidy and commit the result. I was incorrect.

> The toolchain directive only has an effect when the module is the main module and the default toolchain’s version is less than the suggested toolchain’s version.

via https://go.dev/ref/mod#go-mod-file-toolchain

amken3d:
@deadprogram : Updated

tmc5160/tmc5160.go
```diff
@@ -0,0 +1,194 @@
...
+		TSTEP:        "TSTEP",
+	}
+	// Loop through the register list and read each register
+	for _, reg := range registers {
+		// Fetch the register name from the map
+		regName, exists := registerNames[reg]
+		if !exists {
+			regName = "Unknown Register"
+		}
+		//println("Reading reg:", regName)
```
deadprogram: Please remove commented out lines.

tmc5160/SPIcomm.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	//println("Sending", tx[0], tx[1], tx[2], tx[3], tx[4])
+	rx := make([]byte, 5)
+
+	// Perform the SPI transaction
+	err := spi.Tx(tx, rx)
+	if err != nil {
+		return 0, err
+	}
+	//println("Received", rx[0], rx[1], rx[2], rx[3], rx[4])
```
deadprogram: Please remove commented out lines.

amken3d: Done

tmc5160/tmc5160.go
amken3d: Done

examples/tmc5160/main.go
```diff
@@ -0,0 +1,64 @@
...
+	"machine"
+
+	"tinygo.org/x/drivers/tmc5160"
+)
+
+func main() {
+	// Step 1. Setup your protocol.  SPI setup shown below
+	spi := machine.SPI1
+	spi.Configure(machine.SPIConfig{
+		SCK:       machine.GPIO10,
```
deadprogram: You should not need to specify `SCK`, `SDI`, and `SDO` here since they are already the defaults for `SPI1` on Pico. See https://github.com/tinygo-org/tinygo/blob/release/src/machine/board_pico.go#L60-L64 and https://github.com/tinygo-org/tinygo/blob/release/src/machine/machine_rp2040_spi.go#L176-L180

tmc5160/tmc5160.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+type Driver struct {
+	comm      RegisterComm
+	address   uint8
+	enablePin machine.Pin
+	stepper   Stepper
+}
+
+func NewDriver(comm RegisterComm, address uint8, enablePin machine.Pin, stepper Stepper) *Driver {
```
deadprogram, line 17: For other drivers that support more than 1 way to connect to it, we have generally had separate functions to use them e.g. `NewSPI()` and `NewUART()`

See https://github.com/tinygo-org/drivers/blob/release/ssd1306/ssd1306.go#L74-L97 for an example.

Does this seem like a good pattern to use?

deadprogram:
The file named `SPIcomm.go` might better named `spicomm.go` to avoid capitals and also to match `uartcomm.go` naming.

tmc5160/tmc5160.go
amken3d: The reason for my pattern is that it gives you a flexible way to implement shadow interfaces. In my application, I have a host which does a lot of the motor processing, and provides commands to the realtime unit (rp2040/rp2350) which is connected to the physical motor. The host go application uses the same package but implements just the base RegisterComm interface. That way , the host never needs to know what the lower level implementation is. The Host then issues commands to the realtime unit by just telling it what to write/read from which register and which driver to address it to.. The realtime unit then implements the appropriate SPI or UART commands accordingly.  Hope that makes sense.

deadprogram: Thanks for the clarification on that @amken3d

tmc5160/func_test.go
```diff
@@ -0,0 +1,61 @@
+//go:build test
```
deadprogram: I think due to this directive, the tests are in this file are not being run by the CI.

See https://github.com/tinygo-org/drivers/pull/725/checks#step:7:182

If you want to exclude these tests from being run by the GH action you should add it here https://github.com/tinygo-org/drivers/blob/dev/Makefile#L22

It would be better if the tests can run, however. :smile_cat:

amken3d: I removed the file as it was not relevant.

tmc5160/spicomm.go
```diff
@@ -0,0 +1,134 @@
...
+func spiTransfer40(spi *machine.SPI, register uint8, txData uint32) (uint32, error) {
+	// Prepare the 5-byte buffer for transmission (1 byte address + 4 bytes data)
+	tx := []byte{
+		register,           // Address byte
+		byte(txData >> 24), // Upper 8 bits of data
+		byte(txData >> 16), // Middle 8 bits of data
+		byte(txData >> 8),  // Next 8 bits of data
+		byte(txData),       // Lower 8 bits of data
+	}
+	//println("Sending", tx[0], tx[1], tx[2], tx[3], tx[4])
```
deadprogram: Please remove commented out line here.

deadprogram:
Thanks very much for this addition @amken3d now squash/merging.

PR #729, opened 2024-12-18
chore: update microphone driver to use latest i2s interface

deadprogram:
This PR is to update the microphone driver to use the latest i2s interface. See https://github.com/tinygo-org/tinygo/pull/4651


deadprogram:
Merging since it fixes CI.

PR #735, opened 2025-01-13
aht20: Fixed bug in driver read command

crazyorc:
The driver was checking reserved bit 2 (0x4) for calibration status instead of the calibration bit.  Use the package constants (which are correct) instead of magic numbers.

From the datasheet the status bits are defined this way:
Screenshot 2025-01-12 at 6 58 14 PM


deadprogram:
> Use the package constants (which are correct) instead of magic numbers.
:100: 

Now merging, thanks @crazyorc

PR #736, opened 2025-02-10
fix(dhcp): fix typo in DHCP error message

henryleduc:
While reviewing the `netdev` package I noticed the DHCP error messages both have DHCP misspelled.


deadprogram:
Thanks for the correction @henryleduc now merging.

PR #738, opened 2025-02-12
Add MAX6675 Cold-Junction-Compensated K-Thermocouple-to-Digital Converter 

jmhobbs:
Small SPI device for reading temperatures from a probe.

Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/max6675.pdf

Tested with feather-nrf52840 and a generic breakout board.

Happy to make any changes to match conventions, I looked at some of the other devices in here and tried to conform to the style.  I did not see a test framework for SPI so did not end up writing any tests.

Thanks!

### DIFF
--- a/examples/max6675/main.go
+++ b/examples/max6675/main.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"fmt"
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/max6675"
+)
+
+// example for reading temperature from a thermocouple
+func main() {
+	// Pins are for an Adafruit Feather nRF52840 Express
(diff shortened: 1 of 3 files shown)


deadprogram:
Thank you for the new device @jmhobbs 

Can you please add to the smoke tests here? https://github.com/tinygo-org/drivers/blob/dev/smoketest.sh

> I did not see a test framework for SPI so did not end up writing any tests.

That needs to get added by someone, at some point :smile_cat:

jmhobbs:
> Can you please add to the smoke tests here? https://github.com/tinygo-org/drivers/blob/dev/smoketest.sh

Done!

Thanks!

deadprogram:
Thanks you very much @jmhobbs now merging.

PR #741, opened 2025-02-18
Add wifi driver comboat for Elecrow W5 rp2040 and rp2350 devices

scottfeldman:
Combo-AT driver uses AT command set to talk to onboard rtl8720d wifi device.  The driver supports UDP/TCP/TLS clients in Wifi STA mode. Support for UDP/TCP servers is not supported, currently.

https://aithinker-combo-guide.readthedocs.io/en/latest/docs/instruction/index.html https://aithinker-combo-guide.readthedocs.io/en/latest/docs/command-set/index.html https://aithinker-combo-guide.readthedocs.io/en/latest/docs/command-examples/index.html

### DIFF
--- a/comboat/comboat.go
+++ b/comboat/comboat.go
@@ -1,0 +1,711 @@
+// Package comboat implements WiFi driver for the Aithinker-Combo-AT WiFi
+// device found on the Elecrow W5 rp2040 and rp2350 devices.  Ths WiFi device
+// is a RTL8720d variant.  The driver interface is via AT command set over UART
+// (see reference docs below).
+//
+// NOTE: the driver doesn't support UDP/TCP server connections in STA mode,
+// currently.  UDP/TCP/TLS client connections are supported in STA mode.
+//
+// https://aithinker-combo-guide.readthedocs.io/en/latest/docs/instruction/index.html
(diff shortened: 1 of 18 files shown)


scottfeldman:
I don't understand this build failure.

deadprogram:
@scottfeldman the error message is not very clear here
https://github.com/tinygo-org/drivers/actions/runs/13386633341/job/37384751972?pr=741#step:8:596

but the failure is because you are missing the correct build tags for the new board here
https://github.com/tinygo-org/drivers/blob/dev/examples/net/tlsclient/main.go#L8

you will also need it here
https://github.com/tinygo-org/drivers/blob/dev/examples/net/ntpclient/main.go#L6
even though you did not hit that error yet on the CI due to the previous error.

deadprogram:
@scottfeldman I think we would be better off creating a specific build tag for `comboat` and adding it to the boards which have this setup, similar to how the `ninafw` build tag is used here: https://github.com/tinygo-org/tinygo/blob/dev/targets/pyportal.json#L3

That way new boards which have that tag will "just build". What do you think?

scottfeldman:
@deadprogram sounds good.  I'll update PR once I get a work dev env.

deadprogram:
Thanks for making the requested changes here @scottfeldman now squash/merging.

PR #742, opened 2025-02-18
Expectation driven SPI test helper (RFC)

jmhobbs:
This is my WAG at an SPI test helper after I wrote #738

It's based on ordered `Tx` expecations.  I also had to make a partial `Pin` interface as input to my MAX6675, I'm not sure how to finish that out in a meaningful way without `machine`.  I'm not sure if the chip select pin should be part of the test helper or an independent concern.

I'll leave this here as a discussion point, but happy to adjust or expand this as well as try my hand and testing some of the other SPI based drivers to get a feel for the API here.

### DIFF
--- a/pin.go
+++ b/pin.go
@@ -4,5 +4,5 @@ type Pin interface {
 	Get() bool
 	High()
 	Low()
-	Set(high bool)
+	Set(value bool)
 }
--- a/pin.go
+++ b/pin.go
@@ -1,0 +1,8 @@
+package drivers
+
+type Pin interface {
+	Get() bool
+	High()
+	Low()
+	Set(high bool)
+}
--- a/tester/spi.go
+++ b/tester/spi.go
@@ -1,0 +1,48 @@
+package tester
+
+import (
+	"bytes"
+)
+
+type spiExpectation struct {
+	write []byte
+	read  []byte
+}
+
+type SPIBus struct {
+	c            Failer
+	expectations []spiExpectation
+}
+
+func NewSPIBus(c Failer) *SPIBus {
+	return &SPIBus{
+		c:            c,
+		expectations: []spiExpectation{},
+	}
+}
+
+func (s *SPIBus) Tx(w, r []byte) error {
+	if len(s.expectations) == 0 {
+		s.c.Fatalf("unexpected SPI exchange")
+	}
+	ex := s.expectations[0]
+	if !bytes.Equal(ex.write, w) {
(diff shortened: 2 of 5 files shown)


deadprogram:
Very cool, I will take a look!

tester/spi.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (s *SPIBus) Expect(in []byte, out []byte) {
+	if len(in) != len(out) {
+		s.c.Fatalf("Expect: input and output slices must be the same length")
+	}
+	s.expectations = append(s.expectations, spiExpectation{in, out})
+}
+
+type Pin struct {
```
deadprogram: Shouldn't this be in `tester/pin.go`?

pin.go
```diff
@@ -0,0 +1,12 @@
+package drivers
+
+type Pin interface {
+	//  Configure(config PinConfig)
```
deadprogram: Remove commented lines?

deadprogram:
> I also had to make a partial Pin interface

We were needing to do something about that probably, however since some drivers do very specific things with `Pin` we will have to see how far we can take that.

> I'm not sure if the chip select pin should be part of the test helper or an independent concern.

A very good question. Since we do that CS "manually" in the drivers, it probably needs to be somehow separated.

jmhobbs:
I added another, very basic expectation based `Pin` implementation as well.  Works for CS at least.

jmhobbs:
Using something like `testify/mock` and `mockery` might be a better direction for these.

pin.go
```diff
@@ -0,0 +1,8 @@
+package drivers
+
+type Pin interface {
+	Get() bool
+	High()
+	Low()
+	Set(high bool)
```
gandarez: Maybe `Set(value bool)` instead?

jmhobbs: Looks like different boards have a different name there.

https://github.com/tinygo-org/tinygo/blob/8c54e3dd885824638015f319907f75c01570cdad/src/machine/machine_esp8266.go#L141
```go
func (p Pin) Set(value bool) {
```

https://github.com/tinygo-org/tinygo/blob/8c54e3dd885824638015f319907f75c01570cdad/src/machine/machine_nrf.go#L70
```go
func (p Pin) Set(high bool) {
```

I feel like `high` makes sense since a `true` value is high there, but it's not a strong opinion.

gandarez: Can't this function be used to set false value?

jmhobbs: It can, I see your point.

PR #746, opened 2025-03-11
chore: update all SPI usage to use either *machine.SPI or drivers.SPI

deadprogram:
This PR is to update all SPI usage to use either `*machine.SPI` or `drivers.SPI` to accommodate recent switch in machine package to always use `*machine.SPI`.

### DIFF
--- a/examples/flash/console/spi/main.go
+++ b/examples/flash/console/spi/main.go
@@ -10,7 +10,7 @@ import (
 func main() {
 	console_example.RunFor(
 		flash.NewSPI(
-			&machine.SPI1,
+			machine.SPI1,
 			machine.SPI1_SDO_PIN,
 			machine.SPI1_SDI_PIN,
 			machine.SPI1_SCK_PIN,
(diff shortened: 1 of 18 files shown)


gandarez:
Wouldn't this introduce a breaking change since we are changing the `NewXXX()` funcs to accept an interface as param?

deadprogram:
@gandarez that is potentially true, however it is already something we have been moving towards with the I2C interface, for example.

The benefit is greater compatibility in the future, and also ability to write tests such as that proposed in PR #742

gandarez:
What's the rationale to decide between `machine` or `drivers` package?

deadprogram:
I would like to switch entirely to use `drivers.SPI` so that it is testable interface and also no dependency on the TinyGo `machine` package.

However, some current drivers are calling specific SPI methods and I did not want to have to address that problem at the same time, which is why I did not change all of them to use `drivers.SPI` in this PR.

gandarez (approved):

deadprogram:
Thanks for review @gandarez now merging.

PR #747, opened 2025-03-13
Add basic driver for Waveshare 7.5" e-ink

jmhobbs:
This is an incomplete, but functional driver for the [Waveshare 7.5" e-ink panel, v2](https://www.waveshare.com/product/displays/e-paper/epaper-1/7.5inch-e-paper-hat.htm)

All of their documentation and code refers to it as `epd7in5_V2` so I went with roughly that for the names of things.

Currently this only supports full screen draws, and two color (it can handle partial refreshes and a 4-grey mode)

I do intend to come back to those features. I can open a new PR to add features later, or leave this PR open until I get them implemented.

Thanks!


jmhobbs:
I totally missed the whole `waveshare-epd` directory 😂 

I will shift my work there since there is not a 7.5" version yet.

jmhobbs:
Updated!  The API is different from the other `waveshare-epd` drivers, as it doesn't implement `drivers.Displayer`.  Would you prefer it to implement that and handle buffers internally?

deadprogram:
> The API is different from the other waveshare-epd drivers, as it doesn't implement drivers.Displayer. Would you prefer it to implement that and handle buffers internally?

In the interest of supporting other packages that use that interface, yes, please!

b0ch3nski:
@jmhobbs Are you planning to continue the work on this driver? I'd be interested too! :slightly_smiling_face:

jmhobbs:
@b0ch3nski yes, I need to push it to the front of my queue, I will do so.

PR #751, opened 2025-04-07
Add support for bme68x sensor

gandarez:
This PR adds support for BME68X (BME680/BME688) a low power gas, pressure, temperature & humidity sensor.

It includes the support for Sleep and Forced modes as well I2C and SPI interfaces. It introduces the `Options Pattern` to make the initialization of `bme68x` package more flexible.

Simple example:
```golang
tsensor := bme68x.NewI2C(machine.I2C0,
    bme68x.WithIIRFilter(bme68x.Coeff4),
    bme68x.WithTemperatureOversampling(bme68x.Sampling8X),
    bme68x.WithPressureOversampling(bme68x.Sampling4X),
    bme68x.WithHumidityOversampling(bme68x.Sampling2X),
    bme68x.WithHeatrDuration(150),
    bme68x.WithHeatrTemperature(320),
)
if err := tsensor.Configure(); err != nil {
     logger.Error(fmt.Sprintf("error configuring sensor: %s", err.Error()))
     return
}

connected, err := tsensor.Connected()
if err != nil {
    logger.Error(fmt.Sprintf("error checking sensor connection: %s", err.Error()))
    return
}

if !connected {
    logger.Error("sensor not connected")
    return
}

if err := tsensor.SetMode(bme68x.ModeForced); err != nil {
    logger.Error(fmt.Sprintf("error setting sensor mode: %s", err.Error()))
    return
}

for {
    if err := tsensor.Read(); err != nil {
        logger.Error(fmt.Sprintf("error reading sensor: %s", err.Error()))

	time.Sleep(2 * time.Second)
	continue
    }

     logger.Info(strings.Repeat("-", 40))

     logger.Info(fmt.Sprintf("    Temperature: %.2f°C", tsensor.Temperature))
     logger.Info(fmt.Sprintf("    Pressure: %.fhPa", tsensor.Pressure/100))
     logger.Info(fmt.Sprintf("    Gas: %.1fKOhms", tsensor.GasResistance/1000))
     logger.Info(fmt.Sprintf("    Approx. Altitude: %.1fm", bme68x.CalcAltitude(seaLevelPressurehPa, tsensor.Pressure)))
     logger.Info(fmt.Sprintf("    Humidity: %.1f%% (%s)", tsensor.Humidity, humidityDescription))
     logger.Info(strings.Repeat("-", 40))

     time.Sleep(2 * time.Second)
}
```

### DIFF
--- a/bme68x/bme68x.go
+++ b/bme68x/bme68x.go
@@ -1,0 +1,864 @@
+// Package bme68x provides a driver for the BME680/BME688 low power gas,
+// pressure, temperature and humidity sensor by Bosch.
+//
+// Datasheet:
+// https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bme680-ds001.pdf
+package bme68x
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"time"
+
+	"tinygo.org/x/drivers"
+)
+
+const (
+	// The default I2C address which this device listens to.
+	Address = 0x77
+	// AmbientTemperature is the ambient temperature in deg C used for defining the heater temperature.
+	AmbientTemperature = 25.0
+	// TargetTemperature is the target temperature in deg C used for defining the heater temperature.
+	TargetTemperature uint16 = 320
+	// TargetHeatrDuration is the target heater duration in ms used for defining the heater temperature.
+	TargetHeatrDuration uint16 = 150
+	// MeasOffset is the offset for the measurement phase
+	MeasOffset uint32 = 1963
+	// MeasDur is the duration of the TPH switching phase
+	MeasDur uint32 = 1908
+	// GasDur is the duration of the gas measurement phase
+	GasDur uint32 = 2385
+	// WakeUpDur is the duration of the wake up phase in ms
+	WakeUpDur = 1000
+	// MaxDuration is the maximum duration for gas wait
+	MaxDuration = 0xFF
+	// PeriodPoll is thed default period for polling the sensor in µs
+	PeriodPoll uint32 = 10000
+	// PeriodReset is the period for resetting the sensor in µs
+	PeriodReset uint32 = 10000 // 10000 µs
+)
+
+var (
+	osToMeasCycles = [6]uint8{0, 1, 2, 4, 8, 16}
+	lookupK1Range  = [16]float32{
+		0.0, 0.0, 0.0, 0.0, 0.0, -1.0, 0.0, -0.8, 0.0, 0.0, -0.2, -0.5, 0.0, -1.0, 0.0, 0.0,
+	}
+	lookupK2Range = [16]float32{
+		0.0, 0.0, 0.0, 0.0, 0.1, 0.7, 0.0, -0.8, -0.1, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
+	}
+)
+
+type (
+	// calibrationCoefficients reads at startup and stores the calibration coefficients.
+	calibrationCoefficients struct {
+		// temperature related coefficients
+		t1 uint16
+		t2 int16
+		t3 int8
+
+		// pressure related coefficients
+		p1  uint16
+		p2  int16
+		p3  int8
+		p4  int16
+		p5  int16
+		p6  int8
+		p7  int8
+		p8  int16
+		p9  int16
+		p10 int8
+
+		// humidity related coefficients
+		h1 uint16
+		h2 uint16
+		h3 int8
+		h4 int8
+		h5 int8
+		h6 uint8
+		h7 int8
+
+		// gas related coefficients
+		g1 int8
+		g2 int16
+		g3 int8
+
+		// other coefficients
+		resHeatRange uint8
+		resHeatVal   int8
+		rangeSwErr   int8
+	}
+
+	Config struct {
+		Pressure    Oversampling
+		Temperature Oversampling
+		Humidity    Oversampling
+		IIR         FilterCoefficient
+		ODR         ODR
+		// HeatrTemp is the target temperature in degree Celsius.
+		HeatrTemp uint16
+		// HeatrDur is the gas wait period.
+		HeatrDur uint16
+		// HeatrEnable enables gas measurement.
+		HeatrEnable        bool
+		AmbientTemperature int8
+		PeriodPoll         uint32
+		mode               Mode
+	}
+
+	Device struct {
+		bus                     bus
+		address                 uint16
+		chipID                  byte
+		calibrationCoefficients calibrationCoefficients
+		config                  *Config
+		measStart               int64
+		measPeriod              uint16
+
+		// Status contains new_data, gasm_valid and heat_stab bits.
+		Status byte
+		// GasIndex is the index of the heater profile used.
+		GasIndex uint8
+		// MeasIndex is the measurement index to track order.
+		MeasIndex uint8
+		// ResHeat is the heater resistance.
+		ResHeat uint8
+		// GasWait is the gas wait period.
+		GasWait uint8
+		// TemperatureFine is the intermediate temperature coefficient.
+		TemperatureFine float32
+		// Temperature is the temperature in degree Celsius.
+		Temperature float32
+		// Pressure is the pressure in Pascal.
+		Pressure float32
+		// Humidity is the relative humidity in percent (x1000).
+		Humidity float32
+		// GasResistance is the gas resistance in Ohms.
+		GasResistance float32
+		// Idac is the current DAC
+		Idac uint8
+		// VariantID is the variant ID.
+		VariantID uint8
+	}
+
+	// bus is the interface for the I2C and SPI bus.
+	bus interface {
+		Reset(addr uint16) error
+		Read(addr uint16, reg uint8, data []byte) error
+		Write(addr uint16, reg []uint8, data []byte) error
+	}
+)
+
+// NewI2C creates a new BME68x connection. The I2C bus must already be
+// configured.
+//
+// This function only creates the Device object, it does not touch the device.
+func NewI2C(bus drivers.I2C, opts ...Option) *Device {
+	return new(&i2c{
+		bus: bus,
+	}, opts...)
+}
+
+// NewSPI creates a new BME68x connection. The SPI bus must already be
+// configured. It also requires a CS pin to be used as the chip select.
+//
+// This function only creates the Device object, it does not touch the device.
+func NewSPI(bus drivers.SPI, opts ...Option) *Device {
+	return new(&spi{
+		bus: bus,
+	}, opts...)
+}
+
+func new(bus bus, opts ...Option) *Device {
+	device := &Device{
+		address: Address,
+		bus:     bus,
+		config: &Config{
+			mode:               ModeForced,
+			Temperature:        Sampling8X,
+			Humidity:           Sampling2X,
+			Pressure:           Sampling4X,
+			IIR:                Coeff4,
+			ODR:                ODR_NONE,
+			HeatrEnable:        true,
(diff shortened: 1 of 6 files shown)


bme68x/bme68x.go
```diff
@@ -0,0 +1,864 @@
...
+	"fmt"
+	"math"
+	"time"
+
+	"tinygo.org/x/drivers"
+)
+
+const (
+	// The default I2C address which this device listens to.
+	Address = 0x77
```
soypat, line 19: Prefer `DefaultAddress`

bme68x/bme68x.go
```diff
@@ -0,0 +1,864 @@
...
+	// AmbientTemperature is the ambient temperature in deg C used for defining the heater temperature.
+	AmbientTemperature = 25.0
```
soypat, line 21: Unexport these, leave them as defaults if not provided

bme68x/bme68x.go
soypat: If driver provides a way of modifying address then let user calculate it like https://github.com/tinygo-org/drivers/blob/156d6e7c9ce473be185e1554c9da3c2983e6e26a/ndir/ndir.go#L15

If not simply avoid exporting it for now, we can add it later :)

bme68x/option.go
```diff
@@ -0,0 +1,74 @@
+package bme68x
+
+type Option func(*Device)
```
soypat, line 3: See the following pattern 
```go
var config Config = bme68x.DefaultConfig()
config.WithAmbientTemp(30)
config.WithOversampling(bme68x.OverSample ....)
dev := bme68x.NewSPI(spi)
err := dev.Configure(config)
config.WithAmbientTemp(25) // maybe user confused.
err := config.Validate()
```

maybe worth seeing a way to avoid heap allocations. I want long lived Go programs!

soypat: -https://tinygo.org/docs/concepts/compiler-internals/heap-allocation/
- https://github.com/tinygo-org/tinygo-site/pull/451

bme68x/bme68x.go
```diff
@@ -0,0 +1,864 @@
...
+		// Pressure is the pressure in Pascal.
+		Pressure float32
+		// Humidity is the relative humidity in percent (x1000).
+		Humidity float32
+		// GasResistance is the gas resistance in Ohms.
+		GasResistance float32
+		// Idac is the current DAC
+		Idac uint8
+		// VariantID is the variant ID.
+		VariantID uint8
```
soypat, line 141: If users want to access this then this sounds like an oppportunity for a `type Result struct` abstraction which can be returned by `LastResult()` method.

soypat: If any of these fields indicates an error then use standard Go abstraction of returning value with error.

bme68x/bme68x.go
```diff
@@ -0,0 +1,864 @@
...
+}
+
+// Reset does a soft reset by writing 0xB6 to the reset register.
+func (d *Device) Reset() error {
+	return d.bus.Reset(d.address)
+}
+
+// Connected checks if the device is connected by reading the chip ID.
+// It returns true if the chip ID matches the expected value.
+func (d *Device) Connected() (bool, error) {
```
soypat, line 314: `IsConnected() bool` and return false on error.

bme68x/bme68x.go
```diff
@@ -0,0 +1,864 @@
...
+	}
+
+	return nil
+}
+
+func (d *Device) SetGasHeater(temp, dur uint16, enable bool) error {
+	d.config.HeatrTemp = temp
+	d.config.HeatrDur = dur
+	d.config.HeatrEnable = enable
+
```
soypat, line 431: REmove config from Device. If user wants actual config ideally device should query the peripheral and get the config as stored on the peripheral. This is the most-true-to-life reflection of the config because you know, sometimes hardware is faulty and config can be set incorrectly.

bme68x/bme68x.go
```diff
@@ -0,0 +1,864 @@
...
+	// configure only in the sleep mode
+	if err := d.SetMode(ModeSleep); err != nil {
+		return err
+	}
+
+	// read the current configuration
+	var data [5]byte
+	if err := d.bus.Read(d.address, REG_CTRL_GAS_1, data[:]); err != nil {
+		return err
+	}
```
soypat, line 455: If these configurations can be done out-of-order (does datasheet allow it? someimes hardware fails if not done in order it likes) then maybe extract to their own method. Or just leave as one Config method, that way we're sure it works and user interacts with only one point of contact.

bme68x/bme68x.go
```diff
@@ -0,0 +1,864 @@
...
+
+	return nil
+}
+
+// applyHeatrConfig sets the heater configurations.
+func (d *Device) applyHeatrConfig() error {
+	rhRegData := make([]uint8, 1)
+	gwRegData := make([]uint8, 1)
+
+	rhRegData[0] = d.calcResistanceHeat(d.config.HeatrTemp)
```
soypat, line 548: parameter and export this method.

bme68x/bme68x.go
```diff
@@ -0,0 +1,864 @@
...
+
+	if err := d.bus.Write(d.address, []uint8{REG_GAS_WAIT0}, gwRegData[:]); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Read reads all sensor data and store it in the Device struct.
+func (d *Device) Read() error {
```
soypat, line 564: `Update(which drivers.Measurement) error` is new `drivers.Sensor` interface. See https://deploy-preview-451--tinygo.netlify.app/docs/guides/driver-design/, in particular this PR: https://github.com/tinygo-org/drivers/pull/345

bme68x/bme68x.go
```diff
@@ -0,0 +1,864 @@
...
+// calcResistanceHeat calculates the heater resistance value. It takes the target
+// temperature in degree Celsius and returns the calculated heater resistance
+// value.
+func (d *Device) calcResistanceHeat(target uint16) uint8 {
+	// cap temperature to 400°C
+	if target > 400 {
+		target = 400
+	}
+
+	var1 := (float32(d.calibrationCoefficients.g1) / 16.0) + 49
```
soypat, line 798: Beware floating point operatiuons slow on hardware.

soypat: again read avoidances here https://deploy-preview-451--tinygo.netlify.app/docs/guides/driver-design/

bme68x/i2c.go
```diff
@@ -0,0 +1,46 @@
...
+func (i *i2c) Write(addr uint16, reg []uint8, data []byte) error {
+	buf := make([]uint8, LEN_INTERLEAVE_BUFF)
+
+	if len(data) > 0 && len(data) <= int(LEN_INTERLEAVE_BUFF/2) {
+		for i := 0; i < len(data); i++ {
+			buf[2*i] = reg[i]
+			buf[2*i+1] = data[i]
+		}
+
+		if err := i.bus.Tx(uint16(addr), buf, nil); err != nil {
```
soypat, line 40: buf escapes because Tx is an interface and compiler cant prove it does not retain reference to buf.

soypat: Lots of ways of mitigating, it looks pretty straightforward in your case, just add static buffer to `i2c` struct :)
https://deploy-preview-451--tinygo.netlify.app/docs/guides/driver-design/

bme68x/spi.go
```diff
@@ -0,0 +1,122 @@
...
+	// MEM_PAGE_MSK is mask for SPI memory page
+	MEM_PAGE_MSK uint8 = 0x10
+	// SPI_RD_MSK is the mask for reading a register in SPI
+	SPI_RD_MSK uint8 = 0x80
+	// SPI_WR_MSK is the mask for writing a register in SPI
+	SPI_WR_MSK uint8 = 0x7F
+	// MEM_PAGE0 is the SPI memory page 0
+	MEM_PAGE0 uint8 = 0x10
+	// MEM_PAGE1 is the SPI memory page 1
+	MEM_PAGE1 uint8 = 0x00
```
soypat, line 22: maybe unexport

bme68x/spi.go
```diff
@@ -0,0 +1,122 @@
...
+		return fmt.Errorf("failed to set memory page: %w", err)
+	}
+
+	return s.read(reg, data[:])
+}
+
+func (s *spi) read(reg uint8, data []byte) error {
+	reg |= SPI_RD_MSK
+
+	return s.bus.Tx([]byte{reg}, data)
```
soypat, line 64: beware heap alloc here.

bme68x/spi.go
```diff
@@ -0,0 +1,122 @@
...
+		for i := 0; i < len(data); i++ {
+			if err := s.setMemoryPage(reg[i]); err != nil {
+				return fmt.Errorf("failed to set memory page: %w", err)
+			}
+
+			buf[2*i] = reg[i] & SPI_WR_MSK
+			buf[2*i+1] = data[i]
+		}
+
+		if err := s.bus.Tx(buf, nil); err != nil {
```
soypat, line 81: heap

bme68x/spi.go
```diff
@@ -0,0 +1,122 @@
...
+
+	var data [1]byte
+	if err := s.read(REG_MEM_PAGE|SPI_RD_MSK, data[:]); err != nil {
+		return fmt.Errorf("failed to read memory page: %w", err)
+	}
+
+	data[0] &^= MEM_PAGE_MSK
+	data[0] |= (memoryPage & MEM_PAGE_MSK)
+
+	return s.bus.Tx([]byte{REG_MEM_PAGE | SPI_WR_MSK}, []byte{data[0]})
```
soypat, line 110: heap

soypat (commented):
lots of good stuff and also some stuff to improve. hope this helps :)

bme68x/bme68x.go
gandarez: @soypat Only the heater duration can't be simply read from the device because the calculated value is irreversible. More info in the coming commit.

bme68x/spi.go
gandarez: This is specific to SPI but it defines the register address of SPI interface as well the other registers which are all exported. I don't think it's such a bit problem having them exported.

bme68x/spi.go
gandarez: how to avoid heap in this case?

bme68x/spi.go
gandarez: how to avoid heap in this case?

bme68x/spi.go
soypat: Does the user need them?

soypat: If the user already has access to the Read/Write layer, they don't really need them. They might need the registers but not this...

bme68x/spi.go
soypat: Key is always pass in the spi's in-struct array to Tx. All you gotta do is fill it with the correct data.
```go
s.bus.wbuf[0] = reg | SPI_RD_MSK
n := copy(s.bus.wbuf[:], data)
return s.bus.Tx(s.bus.wbuf[:n], s.bus.rbuf[:n])
```

soypat: Also, your driver is technically incorrect, you should always pass in equal length buffers to Tx, or one of them should be nil

bme68x/spi.go
soypat: Same rules as above. use in-struct arrays. please refer to https://deploy-preview-451--tinygo.netlify.app/docs/guides/driver-design/

If the path to avoiding heap allocations is not clear please leave a comment here with what is missing or could be improved: https://github.com/tinygo-org/tinygo-site/pull/451

PR #755, opened 2025-05-04
tmc2209 bug fixes

cswank:
I guess this means nobody has ever used this driver for controlling a TMC2209 via uart because it panics because the buffer is too small.

Even with this fix I'm unable to turn the motor via uart, but I'm not sure why.  If this PR piques someones interest in making a working example I'd be very grateful.  

EDIT:  I got the motor to turn.  The reason turned out to be due to calculating the crc incorrectly.  I pushed another commit that calculates crc according to the datasheet.


deadprogram:
Thank you for the fixes here @cswank now going to squash/merge.

PR #759, opened 2025-05-15
ws2812: add 200MHz support for the Cortex-M0/rp2040

### LINKED ISSUE #754: WS2812 LED stopped changing colour on Pico 2040 with 200MHz clock
My blink program for built-in WS2812 LED on Waveshare RP2040 Zero stopped working after upgrading TinyGo from 0.36.0 to 0.37.0.

I see from the changelog that the CPU frequency for the RP2040 has been boosted from 125 to 200MHz (nice!), but the WS2812 driver doesn't seem to support such lofty frequencies. The switch statement in [ws2812/ws2812_cortexm.go](https://github.com/tinygo-org/drivers/blob/release/ws2812/ws2812_cortexm.go) has a case for 125MHz but tops out at 168MHz - there is no case for the new 200MHz CPU frequency.

My next step was to read through the docs in the `machine` package to see if I could set the CPU back to 125MHz, but could not see any way to do that.

leon-matthews:
My blink program for built-in WS2812 LED on Waveshare RP2040 Zero stopped working after upgrading TinyGo from 0.36.0 to 0.37.0, after we bumped the CPU clock from 125MHz to 200MHz.

Closes #754


deadprogram:
Thanks for the addition here @leon-matthews :speedboat: now merging.

PR #766, opened 2025-06-19
lsm6ds3tr: avoid unnecessary heap allocations

ysoldak:
This re-uses internal buffer to avoid unnecessary heap allocations in lsm6ds3tr driver; switches away from the legacy I2C interface.

### DIFF
--- a/lsm6ds3tr/lsm6ds3tr.go
+++ b/lsm6ds3tr/lsm6ds3tr.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 
 	"tinygo.org/x/drivers"
-	"tinygo.org/x/drivers/internal/legacy"
 )
 
 type AccelRange uint8
@@ -26,7 +25,7 @@ type Device struct {
 	accelSampleRate AccelSampleRate
 	gyroRange       GyroRange
 	gyroSampleRate  GyroSampleRate
-	buf             [6]uint8
+	buf             [7]uint8 // cmd + up to 6 bytes of data
 }
 
 // Configuration for LSM6DS3TR device.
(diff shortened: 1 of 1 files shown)


lsm6ds3tr/lsm6ds3tr.go
```diff
@@ -84,30 +83,32 @@ func (d *Device) doConfigure(cfg Configuration) (err error) {
...
 	if err != nil {
 		return
 	}
 
 	// Configure gyroscope
-	data[0] = uint8(d.gyroRange) | uint8(d.gyroSampleRate)
-	err = legacy.WriteRegister(d.bus, uint8(d.Address), CTRL2_G, data)
+	d.buf[0] = CTRL2_G
+	d.buf[1] = uint8(d.gyroRange) | uint8(d.gyroSampleRate)
+	err = d.bus.Tx(d.Address, d.buf[0:2], nil)
```
soypat: Could we abstract these calls to something like `d.writeReg8/readReg8` ?

```go
func (d *Device) writeReg8(addr, value byte) error {
   d.buf[0] = addr
   d.buf[1] = value
   return err = d.bus.Tx(d.Address, d.buf[0:2], nil)
}

func (d *Device) readReg8(addr byte) (byte,error) {
   d.buf[0] = addr
   err := d.bus.Tx(d.Address, d.buf[0:1], d.buf[1:2])
   return d.buf[1], err
}
```

soypat (changes requested):
Nit that I think would do wonders for readability. I'd love to feature this in a youtube video once it is done if y'all don't mind!

aykevl (commented):
LGTM, but I agree with @soypat some helper methods would make the code a bit easier to read.

lsm6ds3tr/lsm6ds3tr.go
ysoldak: This is just one of many drivers that scream for such improvement, a trial balloon if you wish.
Each of these drivers will have to repeat the pattern.

Any ideas how could we implement these helper functions once and re-use them, instead of implementing in every driver anew?

ysoldak:
> Nit that I think would do wonders for readability. I'd love to feature this in a youtube video once it is done if y'all don't mind!

Nothing against from my side. All kudos to @aykevl for the idea, btw.

lsm6ds3tr/lsm6ds3tr.go
ysoldak: @soypat @aykevl I'm fine with adding helper functions. Shall I just add them and replicate the pattern in other drivers? Or is there a better way?

soypat: I've created [regmap](https://github.com/tinygo-org/drivers/pull/768/files). Thanks for the suggestion

soypat: Maybe add them for now and we can port it to use regmap after regmap has been merged

ysoldak:
Alright, I've implemented the helper functions. Had to implement not only pure read and write, but `setBits()` too -- for convenience.

Also, I've decided to simply pass value size to `readValue()` with contract being the data is written into shared buffer with offset 1 -- first byte is needed for the register number.
We could probably pass a destination slice, but we still need to source it somewhere, and in the scope of this driver it always be a part of the shared buffer anyway. And also anyway, we have to know the internals of the helper method and take into account the size of the shared buffer to chop the destination slice correctly.

An alternative could be we return read data w/o offset, i.e. read into the shared buffer from index 0.
Probably this is a cleaner interface but can be fragile: this assumes too much about I2C implementation, what it does not clear the destination slice first, for example, effectively wiping the register number.
~**UPD: I've tried, this does not work**~ ☝️
**UPD2: This actually works, while still being fragile and assuming perhaps too much about I2C implementation**, see #769

soypat:
> An alternative could be we return read data w/o offset, i.e. read into the shared buffer from index 0.

It is unclear to me what is gained from sharing the first byte, other than 1 byte of memory (likely not gained due to alignment requirements). 

If for some reason you want your data read to be at the first byte of the buffer you can declare two static buffers:

```go
type Dev struct { cmdbuf [1]byte; databuf [N]byte }

func (d *Dev) read(n int) error {
   return d.bus.tx(addr, d.cmdbuf[:], d.databuf[:n])
}
```
Now the intention is clear, there is a databuf which accumulates data read which is already quite self-documenting

ysoldak:
> It is unclear to me what is gained from sharing the first byte, other than 1 byte of memory

Gained is abstraction in the rest of code from the ways these helper functions work, i.e. when you _read_, you expect result be in some buffer starting from 0 index, and not index 1, because how I2C works...

But I agree, can basically make another "data" slice be a view of `buf[1:7]`. Can not make it this straightforward as you proposed though, since we have `writeValue()` function that expects write/command buffer be size of 2.

ysoldak:
@soypat PTAL
I'm closing #769, as I was able to achieve the same internal api clearness (or even better) in this PR already.

ysoldak:
Let me know if you prefer me squash commits manually or we can do squash on PR merge.

soypat:
LGTM! Thank you for all the effort you put in Yurii

soypat (approved):

deadprogram:
Switched target branch to `dev` and now squash/merging commits. Thank you @ysoldak @soypat and @aykevl

PR #769, opened 2025-06-21
lsm6ds3tr: avoid unnecessary heap allocations (alternative)

ysoldak:
This is an alternative implementation of #766 

The difference is here we store read result into the buffer w/o an offset -> easier to work with. 🤩 
Also, this implementation saves one byte, since we don't need one extra in the buffer anymore.

💣  This assumes however that I2C implementation does not clean the destination buffer prior to filling it.

💡 We can still have the read data be stored w/o offset in the buffer, but that requires making the buffer bigger again and then `copy(d.buf[0:size), d.buf[1:size+1))` which is probably suboptimal too.

### DIFF
--- a/lsm6ds3tr/lsm6ds3tr.go
+++ b/lsm6ds3tr/lsm6ds3tr.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 
 	"tinygo.org/x/drivers"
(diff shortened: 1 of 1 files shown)


ysoldak:
Closing in favor of the original #766 as it now achieved same level of internal api clearness w/o extra assumptions about I2C implementation.

PR #770, opened 2025-06-30
Allow GPS with any i2c address (support PA1010D)

jphastings:
[Adafruit's Mini GPS PA1010D Module](https://learn.adafruit.com/adafruit-mini-gps-pa1010d-module) works with this device driver (personally confirmed), but requires `0x10` as the i2c address, rather than `0x42`.

This change allows the device to be initialised with whatever i2c address is needed, while maintaining backward compatibility.

(If backward compatibility isn't needed, please let me know, and I'll remove the function that assumes a default, and force the use of an i2c address to initialise!)

Also adds new constants to allow easy configuration of both the ublox device and the PA1010D.

### DIFF
--- a/examples/gps/i2c/main.go
+++ b/examples/gps/i2c/main.go
@@ -10,7 +10,7 @@ import (
 func main() {
 	println("GPS I2C Example")
 	machine.I2C0.Configure(machine.I2CConfig{})
-	ublox := gps.NewI2C(machine.I2C0)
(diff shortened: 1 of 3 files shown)


deadprogram:
Looks good, thanks for working on this @jphastings 

I switched the target branch to `dev` as mentioned here: https://github.com/tinygo-org/drivers/blob/release/CONTRIBUTING.md#how-to-use-our-github-repository

Now merging, thanks again!

jphastings:
Ahh, I didn't spot that part of the contributing guide — thank you!

PR #774, opened 2025-07-06
drivers.Pin with function pointers example

ysoldak:
This showcases how we can split `drivers.Pin` to smaller `drivers.PinIn` and `drivers.PinOut` interfaces AND use function pointer to `.Set(bool)` function in performance critical places, like `softspi` implementation.

This still breaks users though, as it is based on a PR #749  that mandates pins to be configured outside the drivers.

### DIFF
--- a/pin.go
+++ b/pin.go
@@ -1,10 +1,18 @@
 package drivers
 
-// Pin is a digital pin interface. It is notably implemented by the
-// machine.Pin type.
+// Pin is a digital pin interface.
+// It is notably implemented by the machine.Pin type.
 type Pin interface {
+	PinIn
+	PinOut
+}
+
+type PinIn interface {
 	Get() bool
-	High()
-	Low()
+}
+
+type PinOut interface {
+	High() // deprecated: use Set(true)
+	Low()  // deprecated: use Set(false)
 	Set(high bool)
 }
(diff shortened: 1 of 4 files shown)


pin.go
```diff
@@ -1,10 +1,18 @@
...
+
+type PinIn interface {
 	Get() bool
-	High()
-	Low()
+}
+
+type PinOut interface {
+	High() // deprecated: use Set(true)
+	Low()  // deprecated: use Set(false)
```
ysoldak, line 16: I'd prefer we do not have these in the interface and sweep and eradicate call to them everywhere in drivers.

waveshare-epd/epd2in66b/dev.go
```diff
@@ -18,18 +18,18 @@ const (
...
 
 type Config struct {
-	ResetPin      drivers.Pin
-	DataPin       drivers.Pin
-	ChipSelectPin drivers.Pin
-	BusyPin       drivers.Pin
+	ResetPin      drivers.PinOut
+	DataPin       drivers.PinOut
+	ChipSelectPin drivers.PinOut
+	BusyPin       drivers.PinIn
```
ysoldak, line 24: Showcase use of `PinOut` and `PinIn` interfaces.

soypat:
This is pulled from a PR we agreed to not continue with at the moment. The design of https://github.com/tinygo-org/drivers/pull/753 leaves us open for this design in the future if we so desire to go on with it.

deadprogram:

PR #777, opened 2025-07-14
add regmap package to facilitate heapless driver development

soypat:
Redoing the botched https://github.com/tinygo-org/drivers/pull/768 PR

### DIFF
--- a/internal/regmap/devices.go
+++ b/internal/regmap/devices.go
@@ -1,0 +1,121 @@
+package regmap
+
+import (
+	"encoding/binary"
+	"io"
+
+	"tinygo.org/x/drivers"
+)
+
+// Device8 implements common logic to most 8-bit peripherals with an I2C or SPI bus.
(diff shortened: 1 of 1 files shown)


soypat:
@deadprogram Sorry bout #768, this one should be OK

deadprogram:
@soypat what do you think about adding some test coverage?

deadprogram:
@soypat your thoughts?

soypat:
I'm not sure of the value of adding tests. These are meant to be used by drivers which are themselves tested on the hardware. Same logic goes for this package. I won't trust it works, even with tests, until I see hardware using it. I say add it and let users use it and get some real testing going. It is an internal package so no risk of adding it.

soypat:
Can we merge this? Would like to have it for a class I'll be teaching

deadprogram:
Now merging due to popular demand. :smile_cat:

PR #786, opened 2025-08-18
fix: correct smoke tests for Adafruit Seesaw

deadprogram:
This PR corrects the smoke tests for the Adafruit Seesaw.


deadprogram:
Merging to correct tests...

PR #790, opened 2025-09-06
hx711: add driver for 24 bit ADC

gen2thomas:
Add a driver for the weight cell chip HX711.

TODO:
* [x] add example
* [x] use functional interface according #753 
* [x] introduce a PR for gobot as a reference implementation, see https://github.com/hybridgroup/gobot/pull/1164
* [x] try to switch from float32 to integer, at least for the exposed functions
* [x] check finally the heap allocations
* [x] check finally the conformance to https://tinygo.org/docs/guides/driver-design/


gpio.go
```diff
@@ -0,0 +1,19 @@
...
+	Configure() error
+
+	// Set sets the output to the given state.
+	Set(state bool) error
+}
+
+// InputPinner represents a pin used for digital input.
+type InputPinner interface {
+	// Configure sets the pin as input. Additionally mode can be set, such as an internal pull up or pull down resistor.
+	Configure() error
```
gen2thomas: we could alternatively define ConfigurePullUp(), ConfigurePullDown() function
OR lets the caller configure the pins and only define functional interfaces, according https://tinygo.org/docs/guides/driver-design/

```go
type PinInput func() bool
```
but IMO, if there are many pins to configure the interface with a "Configure()" function is more intuitive, see [ssd1289](https://github.com/tinygo-org/drivers/blob/228e57cf9861a763e79c0f9be44ff23365d5cc95/ssd1289/ssd1289.go#L36) as an example

hx711/hx711.go
```diff
@@ -0,0 +1,407 @@
...
+	B32     GainAndChannelCfg = 2 // channel B, gain factor 32 - 26 pulses
+	A64     GainAndChannelCfg = 3 // channel A, gain factor 64 - 27 pulses
+	A128B32 GainAndChannelCfg = 4 // channel A@128 and channel B@32 - after first read 26 pulses, after second 25 pulses
+	A64B32  GainAndChannelCfg = 5 // channel A@64 and channel B@32 - after first read 26 pulses, after second 27 pulses
+)
+
+const (
+	defaultResetDuration = 100 * time.Microsecond
+	defaultSettlingTime  = 400 * time.Millisecond // according the data sheet: RATE=0 (10Hz), otherwise use 50 ms
+	// readCycleTime = 100 * time.Millisecond // for RATE=0 (10Hz), otherwise 12.5 ms (80Hz)
```
gen2thomas: this is not used internally, e.g. for blocking fast updates, but a caller could have a benefit if we add a "MaxCycleTime()" function to the drivers.Sensor interface

gpio.go
gen2thomas: This is already discussed in #749. A recent proposal is given in #753 . So we will follow the consent of that PR.

PR #791, opened 2025-09-12
Add more flash device support

BryanSouza91:
I have added the definition for the W25Q80DV QSPI flash chip. The max clock speed was also changed for the W25Q80DL to the appropriate speed according to the datasheet.


BryanSouza91:
Added support for the P25Q16H flash chip as well.

deadprogram:
Thank you for the addition @BryanSouza91 now merging.

Issue #733, opened 2025-01-08
Digital Motion Processor device driver extension  for mpu6050


brucebotes:
This is my first ever contribution to the open source community. So please  bear with me and help with the contribution if it does add benefit/ seem fit.

It is a tinygo transaltion of the Jeff Rowberg C/C++ arduinno library for the DMP firmware. I am not sure if this is acceptable and/or legit. If there are no issues,  the team is welcome to add the extension to the MPU6050 library. 

If someone is prepared to guide me through the process I am prepared to give it a shot. 

Of note is the video and  python animation as provided by Maker's Wharf. Please see the referenced youtube link. 
How would we incorporate that within the existing setup?

Please find the repo here: https://github.com/brucebotes/tinygo_mpu6050_dmp

Ps. I tried this Slack link (https://gophers.slack.com/messages/CDJD3SUP6.)to raise this quesion, but that does not seem to work. So here goes...

aykevl:
Yes, supporting this in the TinyGo drivers repo would be great! Feel free to make a PR :)

brucebotes:
Ok great. I looked at the drivers & examples folders and realised that my current folder design does not quite fit in :-( ,  Ie do we create an other mpu6050_dmp driver folder next ro the mpu6505 folder as a sibling and place some of  library stuff there and displace the remainder  in examples?  Or should we place the DMP suff as a sub folder within the existing mpu6050 folder? Do you have any suggestions? Assistance would be appreciated.

aykevl:
Ideally we'd just put it all in the same driver folder - the mpu6050 folder itself. I think that will make the package easier to use.

brucebotes:
Hi Ayke,
Done. I hope and cross my fingers that all is well with the PR. This was my
first PR contribution to open source.
Cheer
Bruce

PR #796, opened 2025-10-02
add Digital Motion Processor for MPU6050 - see issue #733

brucebotes:
Extend the MPU6050 go drivers with Digital Motion Processing functionality. This version does not have a gimble lock.  

See issue 733: Please submit a Pull Request.


mpu6050/digital_motion_processor/main.go
aykevl, line 1: If this is an example, it should be in examples/mpu6050.

mpu6050/digital_motion_processor/python-animation/.DS_Store
aykevl, line 1: Please don't include files like these.

mpu6050/digital_motion_processor/python-animation/Animation.py
aykevl, line 1: I don't think Python examples belong in the drivers repo.

mpu6050/digital_motion_processor/mpu6050/DMP.go
aykevl, line 1: You are essentially recreating the MPU6050 driver here. I really don't want to see two drivers for the same hardware. So my suggestion would be to integrate the DMP changes you made into the existing MPU6050 driver, instead of adding a new driver.

aykevl (changes requested):
See the comments. Essentially, this needs a bit more integration into the drivers repo to be mergeable, meaning the PR should modify the existing mpu6050 driver instead of creating a whole new driver from scratch that duplicates functionality with the existing driver.

brucebotes:
Hi Ayke,
Yea I figured that much, hence my initial email.
Sorry about the .DSstore - i thought I killed it..
With regards to the topology I will have to think it through and give it a
try. In my repo the code is obviously coupled close to the application and
purpose. The tinygo driver layout is more of a challenge to me and I need
to apply some refactoring. I do agree with the example stuff - it was my
initial concern -  but the only way to context with the original Youtube
explanation. Theory is important, you know...

I will have a go at it later when I manage to wrap my mind around the
structure and merger and still keep the context intact. Nevertheless, I get
the gist of it.

If you have more ideas, I am all ears. I will get back to you.

B

PR #797, opened 2025-10-09
Suggestion to make function HAL internal and Interface HAL public

HattoriHanzo031:
This PR is based on #795 with function HAL being moved to internal and interface HAL being made public. onewire, dht and uc8151 drivers are refactored to demonstrate how would the drivers be refactored to eliminate dependency on machine package while preserving backward compatibility and performance. Also onewire_v2 mock driver was added to simulate adding new drivers in future that use the same pin as both input and output.


deadprogram:

PR #799, opened 2025-10-14
Add Honeywell HSC TruStability SPI+I2C pressure sensor driver

soypat:
- https://prod-edam.honeywell.com/content/dam/honeywell-edam/sps/siot/zh-cn/localized/datasheets/sps-siot-trustability-hsc-series-high-accuracy-board-mount-pressure-sensors-50099148-a-ciid-151133-cn.pdf

### DIFF
--- a/examples/honeyhsc/example.go
+++ b/examples/honeyhsc/example.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers"
+	"tinygo.org/x/drivers/honeyhsc"
+)
+
+// Data taken from https://github.com/rodan/honeywell_hsc_ssc_i2c/blob/master/hsc_ssc_i2c.cpp
+// these defaults are valid for the HSCMRNN030PA2A3 chip
+const (
+	i2cAddress = 0x28
+	// 10%
+	outputMinimum = 0x666
+	// 90% of 2^14 - 1
+	outputMax = 0x399A
+	// min is 0 for sensors that give absolute values
+	pressureMin = 0
+	// 30psi (and we want results in millipascals)
+	// pressureMax = 206842.7
+	pressureMax = 206843 * 1000
+)
+
+func main() {
+	bus := machine.I2C0
+	err := bus.Configure(machine.I2CConfig{
+		Frequency: 10_000,
+		SDA:       machine.I2C0_SDA_PIN,
+		SCL:       machine.I2C0_SCL_PIN,
+	})
+	if err != nil {
+		panic(err.Error())
+	}
+	sensor := honeyhsc.NewDevI2C(bus, i2cAddress, outputMinimum, outputMax, pressureMin, pressureMax)
+	for {
+		time.Sleep(time.Second)
+		const measuremask = drivers.Pressure | drivers.Temperature
+		err := sensor.Update(measuremask)
+		if err != nil {
+			println("error updating measurements:", err.Error())
+			continue
+		}
+		P := sensor.Pressure()
+		T := sensor.Temperature()
+		println("pressure:", P, "temperature:", T)
+	}
+}
--- a/examples/honeyhsc/example.go
+++ b/examples/honeyhsc/example.go
@@ -26,7 +26,7 @@ const (
 func main() {
 	bus := machine.I2C0
 	err := bus.Configure(machine.I2CConfig{
-		Frequency: 10_000,
+		Frequency: 400_000, // 100kHz minimum and 400kHz I2C maximum clock. 50 to 800 for SPI.
 		SDA:       machine.I2C0_SDA_PIN,
 		SCL:       machine.I2C0_SCL_PIN,
 	})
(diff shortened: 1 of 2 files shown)


examples/honeyhsc/example.go
```diff
@@ -0,0 +1,49 @@
...
+	pressureMin = 0
+	// 30psi (and we want results in millipascals)
+	// pressureMax = 206842.7
+	pressureMax = 206843 * 1000
+)
+
+func main() {
+	bus := machine.I2C0
+	err := bus.Configure(machine.I2CConfig{
+		Frequency: 10_000,
```
aykevl: 10kHz is an unusually low frequency, is this intentional? (Many chips don't even support such low frequencies).

honeyhsc/hsc.go
```diff
@@ -0,0 +1,135 @@
...
+// DevI2C is the TruStability® High Accuracy Silicon Ceramic (HSC) Series is a piezoresistive silicon pressure sensor offering a ratiometric
+// analog or digital output for reading pressure over the specified full scale pressure span and temperature range.
+type DevI2C struct {
+	bus drivers.I2C
+	dev
+	addr uint8
+	buf  [4]byte
+}
+
+func NewDevI2C(bus drivers.I2C, addr uint8, outMin, outMax uint16, pMin, pMax int32) *DevI2C {
```
aykevl: Some I2C devices have 16-bit addresses, although I don't think we have any support for them at the moment. Perhaps it would be a good idea to use `uint16` anyway?

```suggestion
func NewDevI2C(bus drivers.I2C, addr uint16, outMin, outMax uint16, pMin, pMax int32) *DevI2C {
```

Also, can you add some doc comments?

honeyhsc/hsc.go
```diff
@@ -0,0 +1,135 @@
...
+			cmin: outMin,
+			cmax: outMax,
+			pmin: pMin,
+			pmax: pMax,
+		},
+	}
+	return h
+}
+
+func (d *DevI2C) Update(which drivers.Measurement) error {
```
aykevl, line 65: Same here, please add documentation to this method (especially since it is exported).

honeyhsc/hsc.go
```diff
@@ -0,0 +1,135 @@
...
+	if which&measuremask == 0 {
+		return nil
+	}
+	buf := &d.buf
+	const reg = 0
+	value := (d.addr << 1) | 1
+	err := d.bus.Tx(uint16(d.addr), []byte{reg, value}, buf[:])
```
aykevl: Note that the `[]byte{reg, value}` will likely cause a heap allocation depending on the I2C implementation.

honeyhsc/hsc.go
```diff
@@ -0,0 +1,135 @@
...
+// DevI2C is the TruStability® High Accuracy Silicon Ceramic (HSC) Series is a piezoresistive silicon pressure sensor offering a ratiometric
+// analog or digital output for reading pressure over the specified full scale pressure span and temperature range.
+type DevSPI struct {
+	spi drivers.SPI
+	cs  pinout
+	dev
+	buf [4]byte
+}
+
+func NewDevSPI(conn drivers.SPI, cs pinout, outMin, outMax uint16, pMin, pMax int32) (*DevSPI, error) {
```
aykevl, line 107: Same here, this should get some explanation. At least clarifying with parameters like `outMin` mean.

honeyhsc/hsc.go
```diff
@@ -0,0 +1,135 @@
...
+			cmin: outMin,
+			cmax: outMax,
+			pmin: pMin,
+			pmax: pMax,
+		},
+	}
+	return h, nil
+}
+
+// Update implements the sensor interface.
```
aykevl: That's technically correct, but not super helpful to users.
What about something like

> Update reads the sensor values from the device. The values are stored in the driver object, and can be read with the `Pressure` and `Temperature` methods.

honeyhsc/hsc.go
```diff
@@ -0,0 +1,135 @@
...
+	pmin, pmax int32
+}
+
+// Pressure returns pressure in millipascals [mPa].
+func (d *dev) Pressure() int32 {
+	return d.pressure
+}
+
+// Temperature returns temperature in milliKelvin [mK].
+func (d *dev) Temperature() int32 {
```
aykevl, line 170: By far most drivers use milli-degrees Celsius. See for example:

https://github.com/tinygo-org/drivers/blob/228e57cf9861a763e79c0f9be44ff23365d5cc95/bma42x/bma42x.go#L277-L279

And:

https://github.com/tinygo-org/drivers/blob/228e57cf9861a763e79c0f9be44ff23365d5cc95/bme280/bme280.go#L180-L181

I would like to keep that as a convention.

Also see https://github.com/tinygo-org/drivers/pull/332 which I still think is a good idea. It's somewhat out of scope for this PR though.

examples/honeyhsc/example.go
soypat: Very good catch. 100kHz is minimum for this IC.

honeyhsc/hsc.go
soypat: thanks for the catch!

soypat:
@aykevl Thanks for the suggestions! good stuff :)

deadprogram:
Thanks for the addition @soypat and to @aykevl for review. Now merging.

PR #800, opened 2025-10-17
lsm9ds1: avoid unnecessary heap allocations

ysoldak:
This re-uses internal buffer to avoid unnecessary heap allocations in `lsm9ds1` driver; switches away from the legacy I2C interface.

Same as #766 but for `lsm9ds1` driver.

### DIFF
--- a/lsm9ds1/lsm9ds1.go
+++ b/lsm9ds1/lsm9ds1.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 
 	"tinygo.org/x/drivers"
-	"tinygo.org/x/drivers/internal/legacy"
 )
 
 type AccelRange uint8
@@ -28,7 +27,7 @@ type Device struct {
 	accelMultiplier int32
 	gyroMultiplier  int32
 	magMultiplier   int32
-	buf             [6]uint8
+	buf             [7]uint8 // up to 6 bytes for read + 1 byte for the register address
 }
 
 // Configuration for LSM9DS1 device.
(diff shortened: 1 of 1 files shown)


lsm9ds1/lsm9ds1.go
```diff
@@ -61,19 +60,23 @@ func New(bus drivers.I2C) *Device {
...
 func (d *Device) Connected() bool {
-	data1, data2 := d.buf[:1], d.buf[1:2]
-	legacy.ReadRegister(d.bus, d.AccelAddress, WHO_AM_I, data1)
-	legacy.ReadRegister(d.bus, d.MagAddress, WHO_AM_I_M, data2)
-	return data1[0] == 0x68 && data2[0] == 0x3D
+	data, err := d.readBytes(d.AccelAddress, WHO_AM_I, 1)
+	if err != nil || data[0] != 0x68 {
+		return false
+	}
+	data, err = d.readBytes(d.MagAddress, WHO_AM_I_M, 1)
```
soypat, line 67: not required, but this could be rewritten as d.readByte (single byte return)

soypat: Also heads up: regmap package implements read/write logic related to heapless register mapped peripherals, like this one https://github.com/tinygo-org/drivers/blob/dev/internal/regmap/devices.go

May help out in future modernizations :)

soypat (approved):
Looks great! Thank you Yurii for the heap allocation removals!

Also keep on the lookout for further improvements in future refactors, like fulfilling Sensor interface like Ayke implemented in this PR https://github.com/tinygo-org/drivers/pull/798

deadprogram:
Thank you for the improvements @ysoldak and to @soypat for review. Now merging.

PR #801, opened 2025-10-18
regmap: Add Device8I2C/SPI types and their logic

soypat:
@aykevl  Stemmed from conversation over in driver rewrite https://github.com/tinygo-org/drivers/pull/798#discussion_r2435976562


deadprogram:
Oops, just noticed this. A very good idea! Now squash/merging. Thanks @soypat

Issue #789, opened 2025-09-05
what do you think about golangci-lint?


gen2thomas:
At least all drivers without the machine package import could be checked. I have tested this with the default settings of linter for the ds18b20 folder as an fast example and got 3 important issues regarding a missing check of error return value:

```
ds18b20.go:47:15: Error return value of `d.owd.Select` is not checked (errcheck)
		d.owd.Select(romid)
		            ^
ds18b20.go:57:14: Error return value of `d.owd.Select` is not checked (errcheck)
	d.owd.Select(romid)
	            ^
ds18b20.go:67:14: Error return value of `d.owd.Select` is not checked (errcheck)
	d.owd.Select(romid)
	            ^
ds18b20.go:32:23: ST1005: error strings should not end with punctuation or newlines (staticcheck)
	errReadTemperature = errors.New("Error: DS18B20. Read temperature error: CRC mismatch.")
	                     ^
4 issues:
* errcheck: 3
* staticcheck: 1
```

Currently [gofmt](https://github.com/tinygo-org/drivers/blob/228e57cf9861a763e79c0f9be44ff23365d5cc95/Makefile#L7C1-L7C11) is used in the Makefile.

aykevl:
I think it's a great idea. Though, it may be best to limit checking to new and modified files if possible.

Would you like to make a PR?

gen2thomas:
Hi @aykevl , I have prepared 3 different approaches to run the linter locally and in CI. Please can you have a look and make a decision? Its best to start with reading on #802 . I tried to add some background information to each PR, but maybe there are open questions...

Once we have made a decision I can optimize more regarding namings etc.

PR #802, opened 2025-10-19
fmt: introduce golangci-lint and go vet

gen2thomas:
see #789 

Some notes:

Install the linter locally by: `curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.5.0` or by a way of your choice: https://golangci-lint.run/docs/welcome/install/#local-installation.

The current linter configuration is quite long, because we drop many linters caused by a single or only a few findings. This will usually decreased by the time, when those findings are fixed.

Because we have to skip some packages and this is not supported by the linters, we have to apply the linter to each allowed package (folder) separately. 

This leads to following behavior for local run:
* all directories needs to be given as space separated list

This leads to following behavior for CI run when using the make file in docker container or plain setup in runner (currently implemented in this PR):
* provided [GitHub Action](https://github.com/golangci/golangci-lint-action) is not used
* nice features for CI check can not be used, e.g. show findings inline

Alternatively, there are 2 other PRs using [GitHub Action](https://github.com/golangci/golangci-lint-action):
* see #803 for using "matrix", which leads to multiple runners
* see #804 for using an additional build tag "lint" to filter out unsupported packages

One other alternative way for CI would be to collect all usable packages in another folder and set this as the working directory, but this would introduce other problems:
* inline annotations will be lost, because we check another file internally
* dependencies to "own" drivers package are ambiguous

TODO:
* [x] filter packages without dependency to "machine" imports
* [x] add new targets to Makefile
* [x] introduce linter configuration file
* [x] fix all serious issues in code, normally found by "go vet" ("make check")
* [x] tweak the configuration file
* [x] introduce files running for CI
* [ ] make a decision for this variant or one of the 2 others (see description above)
* [ ] fix all important issues in code
* [ ] discuss about fix of small problems, e.g. unused variables etc.

### DIFF
--- a/.github/workflows/golangci-lint-install.yml
+++ b/.github/workflows/golangci-lint-install.yml
@@ -1,0 +1,26 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - dev
+  pull_request:
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  # pull-requests: read
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    name: "Lint all"
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-go@v6
+        with:
+          go-version: '1.22'
+          cache: false
+      - name: "Install golangci-lint v2.5.0"
+        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.5.0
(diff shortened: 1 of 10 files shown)


soypat:
What's the output of running all those linters? Is the output actionable/useful? Keep in mind tinygo isn't exactly Go so we may run into some/many false positives.

gen2thomas:
@soypat as far I could see until now all output is very useful. Additionally most linters are customizable to fit our needs. You can simple try it by install the linter locally or use the docker version.

PR #758, opened 2025-05-11
Added alarm features for DS3231

mfulz:
I've implemented missing features for DS3231 RTC similar like [RTClib](https://github.com/adafruit/RTClib/tree/master)

These can be used to set Alarm1 (alarm2 is missing), read alarm1, activate alarm1, use sqw for interrupts, etc.

PR #805, opened 2025-10-29
DS3231 Alarm features

gKits:
It seems the original PR #758 was abandoned by @mfulz.

I tried to contact them and contribute the missing Alarm2 features to their PR but sadly I got no response.
Therefore I forked their fork and added the proposed changes in #758.

This PR adds:

- Alarm1 set, enable, read
- Alarm2 set, enable, read
- SQW set, read for interrupts


deadprogram:
Hello @gKits thanks for picking up on this work.

Perhaps you might want to replace the `legacy` package calls with the new `regmap` package, to avoid heap allocations.

Please see https://github.com/tinygo-org/drivers/blob/dev/internal/regmap/i2c-spi.go#L67

What do you think?

gKits:
@deadprogram Sure can do. I used the `regmap` package everywhere I could. Only in `SetTime` and `SetAlarm2` I used the `Tx` method of the bus directly because there I need to write chunks of data that would require more than 1 write because they have an odd number of bytes.

Maybe the `regmap.Device8I2C` type could provide a fitting `WriteData` method for the already existing `ReadData` method in the future?

deadprogram:
Looking very good @gKits 

I have just one more comment. It is a lot better to not use `fmt` package since it greatly increases the size of the resulting binary. Can you switch out the `fmt.Println()` calls in the examples for calls to `println()` instead?

Thank you!

gKits:
@deadprogram Done

deadprogram:
Thank you very much to @gKits for picking this up, and to @mfulz for getting it started, and everyone else who helped review/comment.

Now squash/merging.

PR #809, opened 2025-11-04
Add support for CEVA BNO08x 9DoF sensor

intermernet:
This PR adds support for the [Adafruit BNO085](https://www.adafruit.com/product/4754) ([CEVA BNO08x](https://www.ceva-ip.com/wp-content/uploads/BNO080_085-Datasheet.pdf)) 9DoF sensor. It's based off the Adafruit Arduino package at https://github.com/adafruit/Adafruit_BNO08x .

It implements most of the CEVA [SH-2](https://www.ceva-ip.com/wp-content/uploads/SH-2-Reference-Manual.pdf) / [SHTP](https://www.ceva-ip.com/wp-content/uploads/SH-2-SHTP-Reference-Manual.pdf) protocol whcih could be potentially moved into a separate package, but currently very few devices support it, so I have included it in this package

It has been tested with an Adafruit Metro 2350 and it seems to work well.


deadprogram:
Thank you very much for working on this @intermernet 

Perhaps you might want to look at the new `pin` package to avoid using the `machine` package directly.

Please see https://github.com/tinygo-org/drivers/blob/dev/internal/pin/internalpin.go

What do you think?

intermernet:
@deadprogram I'll look at the `pin` package and see what refactoring I can do. Thanks for the heads up! I didn't know about it!

intermernet:
@deadprogram I don't think the `pin` package applies to this driver, but I was erroneously using `machine.I2C` instead of the `drivers.I2C` interface. I have fixed this and the only remaining `machine` import is in the example code.

There *are* a bunch of linting errors related to unused functions/methods/consts/vars etc. These are mainly due to the implementation of the SH2 / SHTP protocols based on their respective specifications, and some of this isn't required in this driver.

There are also a few place-holders for more advanced functionality which isn't yet implemented.

What would you like me to do with these? I can comment them out and add TODOs, I can delete them altogether, or I can leave them in as they aren't actually causing errors and may help anyone else trying to extend this code in the future.

deadprogram:
I was referring to https://github.com/tinygo-org/drivers/pull/809/files#diff-97e43ef0dd25e88f79e809472ccaf81ff6cf2010147aac700799a5a1aea595bdR35-R62 that seems like better off removed and replaced by using `pin` package.

intermernet:
@deadprogram  I've updated the driver to use the `tinygo.org/x/drivers/internal/pin` package and it seems to work properly with my tests (I have a couple of short diagnostic programs that run through a few different functions of the driver).

I am, however, completely new to the `pin` package and would appreciate any feedback. I think I got it right, and things are working, but I'm not sure!

bno08x/types.go
```diff
@@ -0,0 +1,238 @@
...
+	ShakeDetector              ShakeDetector
+	StabilityClassifier        StabilityClassifier
+	StabilityDetector          uint8
+	ActivityClassifier         ActivityClassification
+	PersonalActivityClassifier PersonalActivityClassifier
+	SleepDetector              uint8
+	TiltDetector               uint8
+	PocketDetector             uint8
+	CircleDetector             uint8
+	HeartRateMonitor           uint16
```
soypat: I suggest unexporting the data fields of the SensorValue and exposing them via methods. This would allow rewriting the internals of SensorValue data fields as a `union` sort of data structure and saving a lot of memory

intermernet: @soypat That sounds reasonable, and I'm happy to do that, but I'd appreciate some guidance. Are you happy for me to just change each of the exported data fields and replace it with a method that returns the field? 

For example:

```
func (s SensorValue) Qauternion() Quaternion {
    return s.quaternion
}
```

and then using `s.Quaternion()` instead of `s.Quaternion`?

soypat: Yes! That'll do for now :)

soypat: Maybe do a check: if the method does not correspond to the ID of the sensor type then panic.

I think its worth unexporting all the fields  in the struct to give it a read only quality

intermernet: Ok, I've unexported the data fields and replaced with methods. I've added the SensorID check, which is useful in the simplest case, but when multiple sensor reports are enabled it's a requirement to check and switch on this value anyway. Still worth having it in for robustness.

deadprogram:
Another thing I just noticed @intermernet  is that you need to add the example to the smoke tests here: https://github.com/tinygo-org/drivers/blob/dev/smoketest.sh

Thanks!

intermernet:
@deadprogram Thanks for the heads up. I've added the example to the smoketest.

deadprogram:
Thanks for all your work on this @intermernet 

@soypat any other feedback on this?

intermernet:
I have a query regarding drivers working on different I/O busses. This driver is basically hard-coded to work on I2C (Mostly in the `Device` struct, which defines `bus drivers.I2C`), but the sensor is completely capable of using SPI or UART. Is there an interface for the bus, or another accepted way of making this more flexible in the future if someone wanted to use SPI or UART?

deadprogram:
@intermernet see https://github.com/tinygo-org/drivers/tree/dev/ssd1306 as example of device that comes in different flavors of interface (I2C/SPI).

intermernet:
> @intermernet see https://github.com/tinygo-org/drivers/tree/dev/ssd1306 as example of device that comes in different flavors of interface (I2C/SPI).

I see that device defines `NewI2C` and `NewSPI`. If I want to work on SPI support for the BNO08x, but don't want to break the API if this PR is merged, would you recommend I rename `New` to `NewI2C` now, and then add the SPI functionality in a seperate PR when it's tested and ready?

deadprogram:
> > @intermernet see https://github.com/tinygo-org/drivers/tree/dev/ssd1306 as example of device that comes in different flavors of interface (I2C/SPI).
> 
> I see that device defines `NewI2C` and `NewSPI`. If I want to work on SPI support for the BNO08x, but don't want to break the API if this PR is merged, would you recommend I rename `New` to `NewI2C` now, and then add the SPI functionality in a seperate PR when it's tested and ready?

Knowing that you already want to add SPI, I think that is probably a very sensible approach.

intermernet:
Ok, I've run into some issues with the dev board I'm using. It seems to have issues with SPI despite the sensor supporting it. I have done some refectoring of the code to allow the easy introduction of SPI in the future, but SPI support has been left out of this commit. The only change to the API is the renaming of `New` to `NewI2C`.

deadprogram:
Thanks very much for working on this @intermernet now squash/merging.

PR #810, opened 2025-11-08
si5351: add support for si5351

deadprogram:
Adds support for the si5351 I2C programmable clock generator using code from @chiefMarlin which used code from @conotto which somehow never got merged.

Thank you everyone!


si5351/si5351.go
```diff
@@ -0,0 +1,491 @@
+package si5351
+
+import (
+	"errors"
+	"fmt"
+	"math"
+
+	"tinygo.org/x/drivers"
+	"tinygo.org/x/drivers/internal/legacy"
```
ysoldak: This is known to cause heap allocations, shall we try and avoid this from start?

deadprogram: Good point @ysoldak I just added another commit to use the `regmap` package instead.

deadprogram: Updated again with the latest code from #801

si5351/si5351.go
```diff
@@ -0,0 +1,448 @@
...
+func (d *Device) Configure() error {
+	// // Disable all outputs setting CLKx_DIS high
+	d.rw.Write8(OUTPUT_ENABLE_CONTROL, 0xFF)
+
+	// Set the load capacitance for the XTAL
+	d.rw.Write8(CRYSTAL_INTERNAL_LOAD_CAPACITANCE, d.crystalLoad)
+
+	// Power down all output drivers
+	buf := []byte{0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80}
+	d.bus.Tx(uint16(d.Address), append([]byte{CLK0_CONTROL}, buf...), nil)
```
soypat: Potentially avoid an allocation by doing(not guaranteed):
```go
buf := []byte{CLK0_CONTROL, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80}
	d.bus.Tx(uint16(d.Address), buf nil)
```

si5351/si5351.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+
+	d.initialised = true
+
+	return nil
+}
+
+// Connected returns whether a device at SI5351 address has been found.
+func (d *Device) Connected() (bool, error) {
+	if err := d.bus.Tx(uint16(d.Address), []byte{}, []byte{0}); err != nil {
```
soypat, line 73: Maybe you can avoid an allocation here by calling `d.rw.Read8`. I understand you are just checking for an error from the I2C implementation for connected or not

si5351/si5351.go
```diff
@@ -0,0 +1,448 @@
...
+	case 0:
+		baseaddr = MULTISYNTH0_PARAMETERS_1
+	case 1:
+		baseaddr = MULTISYNTH1_PARAMETERS_1
+	case 2:
+		baseaddr = MULTISYNTH2_PARAMETERS_1
+	}
+
+	// Set the MSx config registers
+	data := [8]byte{}
```
soypat, line 388: You can avoid an allocation here by adding this buffer to Device.

```go
type Device struct {
    // fields...
    auxbuf [8]byte
}
```


And then use it here:
```go
data := d.auxbuf[:8]
data[0] = uint8((p3 & 0xFF00) >> 8)
// ...
err := d.bus.Tx(uint16(baseaddr), data, nil)
```

si5351/si5351.go
```diff
@@ -0,0 +1,448 @@
...
+	data[0] = uint8((p3 & 0x0000FF00) >> 8)
+	data[1] = uint8(p3 & 0x000000FF)
+	data[2] = uint8((p1 & 0x00030000) >> 16)
+	data[3] = uint8((p1 & 0x0000FF00) >> 8)
+	data[4] = uint8(p1 & 0x000000FF)
+	data[5] = uint8(((p3 & 0x000F0000) >> 12) | ((p2 & 0x000F0000) >> 16))
+	data[6] = uint8((p2 & 0x0000FF00) >> 8)
+	data[7] = uint8(p2 & 0x000000FF)
+	if err := d.bus.Tx(uint16(baseaddr), data[:], nil); err != nil {
```
soypat, line 231: suggestion on how to avoid allocation here below.

si5351/si5351.go
soypat: If you decide to add auxbuf and guaranteed allocation avoidance here make auxbuf length 9 and do:
```go
copy(d.auxbuf[:9], []byte{CLK0_CONTROL, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80})
d.bus.Tx(uint16(d.Address), d.auxbuf[:9], nil)
```

soypat (commented):
Very nice! Love to see regmap in use! FYI got a couple suggestions on how to avoid most if not all allocations

si5351/si5351.go
deadprogram: This already does not escape with this PR, so no further change needed:

```
$ tinygo build -size short -print-allocs=ConfigureMultisynth -o ./build/test.hex -target=pico ./examples/si5351/main.go
  23936     200    5232 |   24136    5432
```

si5351/si5351.go
deadprogram: Before any change:
```
$ tinygo build -size short -print-allocs=Configure -o ./build/test.hex -target=pico ./examples/si5351/main.go
/home/ron/Development/tinygo/drivers/si5351/si5351.go:59:43: object allocated on the heap: escapes at line 59
  23964     200    5232 |   24164    5432
```

After making the first proposed change (getting rid of the append):
```
$ tinygo build -size short -print-allocs=Configure -o ./build/test.hex -target=pico ./examples/si5351/main.go
  23936     200    5232 |   24136    5432
```

si5351/si5351.go
deadprogram: See my previous response.

deadprogram:
Appears that there are no further heap allocations now. Should be ready to go.

si5351/si5351.go
```diff
@@ -0,0 +1,448 @@
...
+func (d *Device) Configure() error {
+	// // Disable all outputs setting CLKx_DIS high
+	d.rw.Write8(OUTPUT_ENABLE_CONTROL, 0xFF)
+
+	// Set the load capacitance for the XTAL
+	d.rw.Write8(CRYSTAL_INTERNAL_LOAD_CAPACITANCE, d.crystalLoad)
+
+	// Power down all output drivers
+	buf := []byte{CLK0_CONTROL, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80}
+	d.bus.Tx(uint16(d.Address), buf, nil)
```
ysoldak, line 59: First real use of regmap in the wild!

I was trying to wrap my head around it and how could I use it in cases like #800 (as suggested).

I basically hit this same problem -- can't just use regmap functions everywhere, consistently, have to fall back to raw `d.bus.Tx` here and there => have to keep both `rw` and `bus`. And that sort of defies the purpose, to me at least... 

We may be better off implementing helper methods for reading and writing variable size slices, like this, but in `internal` package somewhere to be reusable?
https://github.com/tinygo-org/drivers/blob/dev/lsm9ds1/lsm9ds1.go#L216

si5351/si5351.go
deadprogram: I am just going to leave this as it, since it does not escape/allocate.

deadprogram:
```
$ tinygo build -size short -print-allocs=Connected -o ./build/test.hex -target=pico ./examples/si5351/main.go
  23936     200    5232 |   24136    5432
```

si5351/si5351.go
deadprogram: I think that @soypat is working on something like that. In the meantime, we can add in the future if we find it is needed.

deadprogram:
Any other feedback or can this get merged now?

ysoldak (approved):
LGTM

soypat:
Good stuff! Curious how the compiler is managing to not escape an array that is being passed into an interface. Very curious indeed

PR #811, opened 2025-11-09
Regmap Tx8 - heapless buffer writing for SPI and I2C

soypat:
Adds Device8Txer to the regmap package, providing transaction-based register access for I2C and SPI devices using pre-allocated buffers. This enables heapless multi-byte register writes by reusing fixed buffers across transactions, eliminating heap allocations during bus operations. The Device8Txer wraps Device8 and manages write/read buffer state through a transaction API:

- Tx() initiates a transaction with a register address
- AddWriteByte()/AddWriteData() stage bytes into the write buffer
- DoTxI2C()/DoTxSPI() execute the transaction over the respective bus

```go
// Initialization.
var dtx Device8Txer
dtx.SetTxBuffers(make([]byte, 256), make([]byte, 256))

// Usage.
const (
	defaultAddr = 65
	REG_WRITE   = 0x1f
	IOCTL_CALL  = 0xc0
)
tx, err := dtx.Tx(REG_WRITE)
if err != nil {
	panic(err)
}
err = tx.AddWriteData(IOCTL_CALL, 0x80, 0x80)
if err != nil {
	panic(err)
}
var bus drivers.I2C
readData, err := tx.DoTxI2C(bus, defaultAddr, 20)
if err != nil {
	panic(err)
}
fmt.Println(readData)
```

### DIFF
--- a/internal/regmap/device8tx.go
+++ b/internal/regmap/device8tx.go
@@ -1,0 +1,196 @@
+// Package regmap provides transaction-based interfaces for reading and writing
+// to device registers over I2C and SPI buses with pre-allocated buffers.
+package regmap
+
+import (
+	"errors"
+
+	"tinygo.org/x/drivers"
+)
+
+var (
+	// errNotInTx indicates an operation was attempted outside of an active transaction.
+	errNotInTx = errors.New("device not in Tx")
+
+	// errInTx indicates a transaction was started while another is still active.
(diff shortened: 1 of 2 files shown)


soypat:
@ysoldak @deadprogram check this out! Saw the pains in latest si5351 driver and came up with this!

PR #814, opened 2025-11-10
Pin HAL adapt - #753 with #795 semantics

soypat:
Just brought over changes in #753 modifying names in struct.
- #753 Changes are from this PR
- #795 Adapted to this PR's API

PR #815, opened 2025-11-10
Reduced pinhal port - "1-to-1 Ports"

soypat:
Is #814 but reduced in scope. I only added drivers which are simple to verify due to 1-to-1 porting logic because I imagine we won't be testing each driver individually- it should be enough to trust the compiler and several eyes.


deadprogram (approved):
Thank you for reducing scope to a set of more easily reviewed drivers!

PR #816, opened 2025-11-12
i2clegacy: avoid heap allocations in i2c legacy helper functions

ysoldak:
This theoretically shall instantly improve memory usage in majority of older drivers.
The buffer grows adapting for the need.

This is just a low hanging best-effort patch until we come around end erradicate legacy way of calling I2C in all the drivers.

What do you think?

### DIFF
--- a/internal/legacy/i2clegacy.go
+++ b/internal/legacy/i2clegacy.go
@@ -2,13 +2,18 @@ package legacy
 
 import "tinygo.org/x/drivers"
 
+var i2cBuf []uint8 = make([]uint8, 1)
+
 func ReadRegister(bus drivers.I2C, addr uint8, reg uint8, data []byte) error {
-	return bus.Tx(uint16(addr), []byte{reg}, data)
+	i2cBuf[0] = reg
+	return bus.Tx(uint16(addr), i2cBuf[:1], data)
 }
 
 func WriteRegister(bus drivers.I2C, addr uint8, reg uint8, data []byte) error {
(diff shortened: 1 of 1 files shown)


internal/legacy/i2clegacy.go
```diff
@@ -2,13 +2,18 @@ package legacy
 
 import "tinygo.org/x/drivers"
 
+var i2cBuf []uint8 = make([]uint8, 1)
+
 func ReadRegister(bus drivers.I2C, addr uint8, reg uint8, data []byte) error {
-	return bus.Tx(uint16(addr), []byte{reg}, data)
+	i2cBuf[0] = reg
```
deadprogram, line 8: Since this is a slice only being used inside this function, it could be declared as an array or a slice of a single byte, and I do not think it would be escaped. This would remove the need for the package level `i2cBuf` var.

ysoldak: we still need i2cBuf var in WriteRegister, see below

aykevl (changes requested):
This is not concurrency safe. For example, this will lead to data races on the rp2040 when two different goroutines use two different I2C peripherals through a driver, something that would otherwise be safe. It will also lead to data races on other chips that call `gosched()` while waiting for the I2C transfer to complete (therefore allowing concurrency).

ysoldak:
Alright, makes sense, thanks for giving this a look. Now closing, since we obviously don't want races.

PR #819, opened 2025-11-21
SCD4x: use standard sensor interface

aykevl:
This deprecates the older Read* methods and uses standard Update() and sensor data getters instead.

Apart from being more efficient, this also makes the driver more efficient when reading multiple sensors (since the SCD4x sensor won't get polled for new data at each Read* method call).

This PR also removes some dead code.


deadprogram:
Thanks for the improvements @aykevl now merging.

PR #822, opened 2025-12-07
si5351: add many missing functions needed for convenient use.

deadprogram:
This adds many missing functions needed for convenient use that are reinterpreted from the C code in https://github.com/dmalnati/picoinf

Thank you!


soypat (approved):
LGTM!

deadprogram:
Thanks for review @soypat now merging.

PR #823, opened 2025-12-10
sx127x: add functions used for FSK radio communication

deadprogram:
This PR adds a few missing functions to the `sx127x` that are used for FSK radio communication. It also cleans up a few existing functions to clarify the implementation. Also tested that Lora still works!


deadprogram:
Review requested please! :smile_cat:

sx127x/sx127x.go
```diff
@@ -81,23 +81,31 @@ func (d *Device) DetectDevice() bool {
...
-	//d.spiTxBuf = []byte{reg & 0x7f}
 	d.spiTxBuf = d.spiTxBuf[:0]
 	d.spiTxBuf = append(d.spiTxBuf, byte(reg&0x7f))
 	d.spi.Tx(d.spiTxBuf, nil)
 	// Read value
 	d.spiRxBuf = d.spiRxBuf[:0]
 	d.spiRxBuf = append(d.spiRxBuf, 0)
 	d.spi.Tx(nil, d.spiRxBuf)
-	d.controller.SetNss(true)
+	if d.controller != nil {
```
HattoriHanzo031, line 101: Is there a use-case of using the driver without the controller? If yes, maybe we could in `New` set the default `nopControler` that does nothing to avoid this checks.

sx127x/sx127x.go
```diff
@@ -244,9 +270,9 @@ func (d *Device) SetLowDataRateOptim(val uint8) {
 // SetLowFrequencyModeOn enables Low Data Rate Optimization
 func (d *Device) SetLowFrequencyModeOn(val bool) {
 	if val {
-		d.WriteRegister(SX127X_REG_OP_MODE, d.ReadRegister(SX127X_REG_OP_MODE)|0x04)
+		d.WriteRegister(SX127X_REG_OP_MODE, d.ReadRegister(SX127X_REG_OP_MODE)|(1<<SX127X_OPMODE_LOW_FREQUENCY_Pos))
```
HattoriHanzo031: We could also define `SX127X_OPMODE_LOW_FREQUENCY` constant and set it to `0x04` and then just do:
```suggestion
		d.WriteRegister(SX127X_REG_OP_MODE, d.ReadRegister(SX127X_REG_OP_MODE)|SX127X_OPMODE_LOW_FREQUENCY)
```

sx127x/sx127x.go
```diff
@@ -119,9 +130,24 @@ func (d *Device) SetOpMode(mode uint8) {
...
+
+// SetModulationType changes the modulation type (0=FSK, 1=OOK)
+func (d *Device) SetModulationType(typ uint8) {
+	switch typ {
+	case SX127X_OPMODE_MODULATION_FSK:
+		d.WriteRegister(SX127X_REG_OP_MODE, d.ReadRegister(SX127X_REG_OP_MODE)&^(1<<SX127X_OPMODE_MODULATION_Pos))
+	case SX127X_OPMODE_MODULATION_OOK:
+		d.WriteRegister(SX127X_REG_OP_MODE, d.ReadRegister(SX127X_REG_OP_MODE)|(1<<SX127X_OPMODE_MODULATION_Pos))
+	}
 }
```
HattoriHanzo031, line 152: Documentation says this is two bit field, although only 0b00 and 0b01 are used, it would maybe be safer to use mask like:
```suggestion
const (
	SX127X_OPMODE_MODULATION_MASK = uint8(0x60)
	SX127X_OPMODE_MODULATION_FSK = uint8(0x0)
	SX127X_OPMODE_MODULATION_OOK = uint8(0x20)
)
func (d *Device) SetModulationType(typ uint8) {
	cleared := d.ReadRegister(SX127X_REG_OP_MODE)&^SX127X_OPMODE_MODULATION_MASK
	d.WriteRegister(SX127X_REG_OP_MODE, cleared|typ)
}
```

HattoriHanzo031 (commented):
Just some small, non-functional adjustments for clarity/consistency.

sx127x/sx127x.go
```diff
@@ -25,9 +25,9 @@ type Device struct {
 	rstPin         machine.Pin          // GPIO for reset
 	radioEventChan chan lora.RadioEvent // Channel for Receiving events
 	loraConf       lora.Config          // Current Lora configuration
-	controller     RadioController      // to manage interactions with the radio
+	controller     RadioController      // to manage interrupts with the radio
 	deepSleep      bool                 // Internal Sleep state
-	deviceType     int                  // sx1261,sx1262,sx1268 (defaults sx1261)
+	deviceType     int                  // sx1272, sx1273, sx1276, sx1279 (defaults sx1276)
```
HattoriHanzo031, line 30: `deviceType` is actually not used anywhere

sx127x/sx127x.go
deadprogram: I would like to leave that question for a future PR.

sx127x/sx127x.go
deadprogram: Good idea! Done.

sx127x/sx127x.go
deadprogram: Done.

sx127x/sx127x.go
deadprogram: That was an oversight on my part, I think. I have added `SetDeviceType()` to make consistent with `sx126x`.

deadprogram:
Thanks for the feedback @HattoriHanzo031 I have made a few updates.

I also added a smoketest for the `sx127x` which was missing. So I think this PR is now ready.

sx127x/registers.go
```diff
@@ -96,6 +96,14 @@ const (
 	SX127X_OPMODE_RX_SINGLE = uint8(0x06)
 	SX127X_OPMODE_CAD       = uint8(0x07)
 
+	SX127X_OPMODE_MODULATION_Pos    = uint8(0x5)
+	SX127X_OPMODE_LOW_FREQUENCY_Pos = uint8(0x3)
```
HattoriHanzo031: Just remove these two unused constants.

HattoriHanzo031 (approved):
Looks good to me (after unused constants are removed)

PR #824, opened 2025-12-14
st7789: improvements

deadprogram:
This PR contains several improvements to the `st7789` driver:

It remove the dependency on the `machine` package by using the `internal/pin` package. It also defines a new `st7789.Bus` interface type to allow for both SPI and parallel connections to the display. Lastly, it switches the example to use the Adafruit Clue board, since that board actually does come with this display.


deadprogram:
Once the Tufty2040 board display is working with these changes, then we can say it is ready for review.

PR #826, opened 2025-12-15
build: use the latest TinyGo release container instead of the dev container for builds

deadprogram:
This PR is to use the latest TinyGo release container instead of the dev container for builds, as discussed by the team.


ysoldak (approved):

deadprogram:
Thanks for review @ysoldak now merging.

PR #827, opened 2025-12-21
st7735: remove dependency on the machine package

ysoldak:
Minimal, just to remove the dependency.


deadprogram:
Thank you @ysoldak now merging.

PR #828, opened 2025-12-21
rpio: basic implementation of Pins and SPI interfaces for Raspberry Pi

ysoldak:
Depends on #827 

This may help someone to jump-start their use of `drivers` package on RPi.
Possible now, as almost any driver can be made `machine`-independent.

### DIFF
--- a/examples/rpio/main.go
+++ b/examples/rpio/main.go
@@ -1,0 +1,91 @@
+package main
+
+// Example program for the ST7735 display (Waveshare 1.44" LCD HAT) using the rpio driver on a Raspberry Pi.
+// Build command:
+//   GOOS=linux GOARCH=arm64 go build -o rpio-example ./examples/rpio/
+
+import (
+	"fmt"
+	"image/color"
+	"time"
+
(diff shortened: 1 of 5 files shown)


examples/rpio/main.go
```diff
@@ -0,0 +1,68 @@
...
+	colors = [...]color.RGBA{
+		{255, 0, 0, 255},   // red
+		{0, 255, 0, 255},   // green
+		{0, 0, 255, 255},   // blue
+		{255, 255, 0, 255}, // yellow
+	}
+)
+
+func main() {
+	if err := go_rpio.Open(); err != nil {
```
ysoldak, line 29: We probably don't want to go as far as hiding even this (do automatically) in the `rpio` package. 
Leaving full control to user, `rpio` package is just a convenience wrapper implementing our interfaces.

amken3d:
Yuri,  if you are interested in extending the support to RPi5 then you may want to take a look at https://github.com/amken3d/GoRPI5.  I had worked on it a year ago and totally forgot about it till now. I just made it public.

PR #829, opened 2025-12-26
gps: improve implementation for UBX config commands

deadprogram:
This PR is to improve the GPS implementation for the use of structured UBX config commands.

### DIFF
--- a/gps/ublox.go
+++ b/gps/ublox.go
@@ -1,53 +1,227 @@
 package gps
 
 import (
-	"errors"
 	"time"
 )
 
-// flight mode disables the GPS COCOM limits
-var flight_mode_cmd = [...]byte{
-	0xB5, 0x62, 0x06, 0x24, 0x24, 0x00, 0xFF, 0xFF, 0x06, 0x03, 0x00, 0x00, 0x00,
-	0x00, 0x10, 0x27, 0x00, 0x00, 0x05, 0x00, 0xFA, 0x00, 0xFA, 0x00, 0x64, 0x00,
-	0x2C, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-	0x00, 0x00, 0x00, 0x16, 0xDC}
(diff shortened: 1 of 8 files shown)


gps/ublox.go
```diff
@@ -1,53 +1,227 @@
...
+
+// appendChecksum calculates UBX checksum and appends it to the message
+func appendChecksum(msg []byte) []byte {
+	var ckA, ckB byte
+	// Checksum covers class, ID, length, and payload (skip sync chars)
+	for i := 2; i < len(msg); i++ {
+		ckA += msg[i]
+		ckB += ckA
+	}
+	return append(msg, ckA, ckB)
```
ysoldak, line 214: This gonna heap allocate like crazy, gonna not? Probably fine in the supposed application?

deadprogram: I do not think so, since the slice passed in is actually the preallocated buffer. As long as `append` does not need to increase the capacity, it will not reallocate.

deadprogram:
```$ tinygo build -target pico -size short -print-allocs=gps ./examples/gps/uart/
/home/ron/Development/tinygo/drivers/gps/gpsparser.go:107:25: object allocated on the heap: escapes at line 107
/home/ron/Development/tinygo/drivers/gps/gps.go:67:17: object allocated on the heap: escapes at line 67
/home/ron/Development/tinygo/drivers/gps/gps.go:185:21: object allocated on the heap: escapes at line 185
  68716    1652    5416 |   70368    7068
```

@ysoldak looks like only the heap allocations are ones that were already there. Probably a good time to address that, so I will do so in a separate commit to this PR.

deadprogram: @ysoldak see additional commit that removed the heap allocation on error return.

```
$ tinygo build -target pico -size short -print-allocs=gps ./examples/gps/uart/
/home/ron/Development/tinygo/drivers/gps/gps.go:46:17: object allocated on the heap: escapes at line 46
  68372    1640    5416 |   70012    7056
```

Now only the allocation for the buffer remains.

ysoldak (approved):

deadprogram:
Thanks for the additions/reviews @ysoldak and @soypat now merging.

PR #834, opened 2026-01-11
gps: improvements and corrections for config commands

deadprogram:
This PR contains some improvements and corrections for the `gps` driver. It adds some additional functions for different modes (automobile, bike, etc) and also ignores the return results from any config commands.

Basically due to the fact that there is a constant stream of updates coming from NMEA messages, the results from sending any UBX commands are getting lost. Better to just ignore them for now.

### DIFF
--- a/gps/gpsparser.go
+++ b/gps/gpsparser.go
@@ -6,12 +6,28 @@ import (
 	"time"
 )
 
+type NMEASentenceType string
+
+const (
+	GSA NMEASentenceType = "GSA"
+	GGA NMEASentenceType = "GGA"
+	GLL NMEASentenceType = "GLL"
+	GSV NMEASentenceType = "GSV"
+	RMC NMEASentenceType = "RMC"
+	VTG NMEASentenceType = "VTG"
+	ZDA NMEASentenceType = "ZDA"
+	TXT NMEASentenceType = "TXT"
+)
+
 // Parser for GPS NMEA sentences.
 type Parser struct {
 }
 
 // Fix is a GPS location fix
 type Fix struct {
+	// Type is the NMEA sentence type that provided this fix.
+	Type NMEASentenceType
+
 	// Valid if the fix was valid.
 	Valid bool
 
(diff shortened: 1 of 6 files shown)


gps/ublox.go
```diff
@@ -29,8 +28,40 @@ var flightModeCmd = CfgNav5{
...
+	d.SendCommand(d.buffer[:42])
+	return nil
+}
+
+// SetBikeMode sends UBX-CFG-NAV5 command to set GPS into bike mode
+func (d *Device) SetBikeMode() (err error) {
+	nav5Cmd.DynModel = DynModeBike
+	nav5Cmd.FixMode = FixModeAuto
+	nav5Cmd.Put42Bytes(d.buffer[:])
+	d.SendCommand(d.buffer[:42])
```
soypat: Can't you return the SendCommend error? Here and everywhere else

deadprogram: > Basically due to the fact that there is a constant stream of updates coming from NMEA messages, the results from sending any UBX commands are getting lost. Better to just ignore them for now.

deadprogram: I realized the problem was not that an error was being returned, even though that error might be not actually an error. My problem was that the code was stopping from sending subsequent commands on an error.

Now have improved the implementation to return error, and count on the caller to ignore/handle as needed.

Thanks for the feedback on that @soypat

gps/gpsparser.go
```diff
@@ -6,12 +6,28 @@ import (
...
+
+const (
+	GSA NMEASentenceType = "GSA"
+	GGA NMEASentenceType = "GGA"
+	GLL NMEASentenceType = "GLL"
+	GSV NMEASentenceType = "GSV"
+	RMC NMEASentenceType = "RMC"
+	VTG NMEASentenceType = "VTG"
+	ZDA NMEASentenceType = "ZDA"
+	TXT NMEASentenceType = "TXT"
```
soypat, line 19: If these are not arguments or needed as user facing API I'd keep them internal, unexported

deadprogram: They are needed in user apps.

deadprogram: Example:

```go
		if !gpsFixAcquired && newfix.Type == gps.GSV {
			println("Waiting for fix... ", newfix.Satellites, "satellites in view")
		}
```

deadprogram:
Any other feedback here @soypat or anyone else?

soypat (approved):

deadprogram:
Thanks @soypat now merging.

Issue #703, opened 2024-08-16
pin name mappings for sx1262 LoRa radio?


gherlein:
@ofauchon  perhaps you know?

var (
	spi                        = machine.SPI1
	nssPin, busyPin, dio1Pin   = machine.GP13, machine.GP6, machine.GP7
	rxPin, txLowPin, txHighPin = machine.GP9, machine.GP8, machine.GP8
)

what are those pin names?  nss is CS, busyPin is busy, dio1 is dio1 - bit the others don't map to any pin names I can find on my part or in the datasheet.  Please advise?
![image](https://github.com/user-attachments/assets/0b6e2ec4-0e22-4c15-985d-9e7385d33e87)

deadprogram:
Probably comes from this board: https://wiki.seeedstudio.com/LoRa_E5_mini/

warrenguy:
@gherlein I think what you're looking for is the RXEN and TXEN pins? If there's no low power TX then set it to `machine.NoPin`.

jwetzell:
I know this is old but I noticed this when I was looking at this driver as well and it seems odd. 

I would think the base `radiocontrol_pins` driver should only include what is actually available on the SX126x chips (i.e no rxHigh, txHigh, etc and add DIO2). The driver already has additional `radiocontrol_xxx` options for tailoring things towards more specific boards like Wio stuff which include more things than just a raw SX126x chip like the RF switch being controlled by the rx/txHigh stuff.

I'll be looking at this when I get my LoRa boards in here soon.

PR #835, opened 2026-01-11
[sx126x] Wio sx1262 support

HattoriHanzo031:


deadprogram:
Anything else needed on this PR @HattoriHanzo031 ? 

Ignore my IRL questions about the `wireless` repo migration for the moment.

HattoriHanzo031:
> Anything else needed on this PR @HattoriHanzo031 ?
> 
> Ignore my IRL questions about the `wireless` repo migration for the moment.

I just did one last minute change to use new `pin.Output` abstraction instead of machine pin. 
If everyone is fine with breaking change in `New` function, I don't have anything more to add.

sx126x/sx126x.go
```diff
@@ -42,7 +41,7 @@ const (
 // Device wraps an SPI connection to a SX126x device.
 type Device struct {
 	spi            drivers.SPI          // SPI bus for module communication
-	rstPin         machine.Pin          // GPIO for reset pin
+	rstPin         pin.Output           // GPIO for reset pin
```
deadprogram, line 44: 🥇

Escalion:
I'll volunteer to test this as I'm working with this hardware right now, but the PR looks good.

HattoriHanzo031:
> I'll volunteer to test this as I'm working with this hardware right now, but the PR looks good.

@Escalion Thank you, this would be very helpful. Please report any bugs or improvements that could be implemented. 

P.S. what is the exact hardware you are using (MCU/module combination)?

jwetzell:
I've got a couple of the sx1262 boards coming in (LILYGO T3S3) that I could use to test. I saw another issue related to this driver in searching (#703) and brought up there that it would probably be nice (although a bit breaking) if the base `_pins` driver only has support for what the base sx126x chips have. Right now it looks like it is setup for some specific module that has some sort of RF switch alongside the sx126x. I think renaming the `_pins` file to reflect whatever kind of module that is and having a more basic `_pins` that actually reflects just the base sx126x chip would be a more accurate implementation. 

As well as the base driver doesn't expose DIO pins 2 and 3 (likely because whatever module was being used only has DIO1 connected) and the specific modules could simply use NoPins to reflect that they don't have that connection.

PR #841, opened 2026-02-21
make: add tasks to count and list drivers

deadprogram:
This PR is to add make tasks to count and list drivers. Should be helpful to keep the docs up to date.

```shell
$ make drivers-count 
Total drivers: 140 (root: 134, waveshare-epd: 6)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,3 +26,17 @@ unit-test:
 	@go test -v $(addprefix ./,$(TESTS))
 
 test: clean fmt-check unit-test smoke-test
+
+EXCLUDE_DIRS = build cmd examples internal lora ndir netdev netlink tester
+
+drivers-count:
+	@root_count=$$(find . -mindepth 1 -maxdepth 1 -type d | grep -vE '^\./($(subst $(space),|,$(EXCLUDE_DIRS)))$$' | wc -l); \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,3 +26,17 @@ unit-test:
 	@go test -v $(addprefix ./,$(TESTS))
 
 test: clean fmt-check unit-test smoke-test
+
+EXCLUDE_DIRS = build cmd examples internal lora ndir netdev netlink tester
+
+drivers-count:
+	@root_count=$$(find . -mindepth 1 -maxdepth 1 -type d | grep -vE '^\./($(subst $(space),|,$(EXCLUDE_DIRS)))$$' | wc -l); \
+	epd_count=$$(find ./waveshare-epd -mindepth 1 -maxdepth 1 -type d 2>/dev/null | wc -l); \
```
b0ch3nski, line 34: In receipt below, you are checking for existence of `./waveshare-epd` dir but here the potential error is dropped to /dev/null - do we actually need to protect against this dir not existing? If so, would be good to unify this ;)

deadprogram: > do we actually need to protect against this dir not existing?

No, we can be sure it is there.

deadprogram:
Now merging, thanks for review @b0ch3nski

PR #849, opened 2026-03-27
neopixel: add PIO-based WS2812B driver for RP2040/RP2350

gandarez:
## Summary
- Adds a new `neopixel` package that provides an Adafruit NeoPixel-like API for WS2812B RGB LED strips
- Uses **PIO** (via `tinygo-org/pio`) for hardware-timed control on RP2040/RP2350, fixing the issue where the existing `ws2812` bit-bang driver fails on RP2350's Cortex-M33
- Falls back to the existing `ws2812` bit-bang driver on all other platforms
- Includes an example targeting the Waveshare RP2350-LCD-1.47 board (GP22, 1 LED)

## Motivation
The existing `ws2812` driver uses CPU-timed bit-banging assembly tuned for Cortex-M0+. The RP2350's Cortex-M33 has a write buffer and different pipeline that breaks the nanosecond-level GPIO timing. The Arduino Adafruit_NeoPixel library works on the same hardware because it uses PIO.

## API
```go
pixel, err := neopixel.New(machine.GP22, 1)
pixel.SetBrightness(50)
pixel.SetPixel(0, color.RGBA{R: 255})
pixel.Show()
pixel.Fill(color.RGBA{G: 255})
pixel.Clear()
```

## Test plan
- [x] `tinygo flash -target=pico2 ./examples/neopixel/` — verify LED cycles Red → Green → Blue on Waveshare RP2350
- [x] Verify brightness control matches Arduino behavior at brightness=50
- [ ] Verify fallback compiles on non-RP targets

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/neopixel/main.go
+++ b/examples/neopixel/main.go
@@ -1,0 +1,36 @@
+// Demonstrates NeoPixel (WS2812B) RGB LED control.
+//
+// Tested on Waveshare RP2350-LCD-1.47 with onboard RGB LED on GP22.
+//
+// To flash:
+//
+//	tinygo flash -target=pico2 ./examples/neopixel/
+package main
+
+import (
+	"image/color"
+	"machine"
+	"time"
+
+	"tinygo.org/x/drivers/neopixel"
+)
+
+func main() {
+	pixel, err := neopixel.New(machine.GP22, 1)
+	if err != nil {
+		panic(err)
+	}
+	pixel.SetBrightness(50)
+
+	colors := []color.RGBA{
+		{R: 255, G: 0, B: 0},
+		{R: 0, G: 255, B: 0},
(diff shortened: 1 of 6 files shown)


soypat (approved):

soypat:
Looks good! Ready to merge?

deadprogram:
@gandarez I just wonder if this might be able to be added directly to the existing `ws2812` driver instead of adding another driver. I have not looked but seems like with the correct build tags you could bypass the CPU and use the PIO code in this PR instead.

What do you think?

gandarez:
Let me check if it's possible

gandarez:
Closing due to https://github.com/tinygo-org/drivers/pull/851

PR #851, opened 2026-03-30
ws2812: use PIO on RP2040/RP2350 transparently via build tags

gandarez:
## Summary
- On RP2040/RP2350, `NewWS2812()` now uses **PIO** (via `tinygo-org/pio`) for hardware-timed control instead of bit-banging
- Falls back to the existing bit-bang driver if PIO initialization fails, or on all other platforms
- No changes to the exported API surface — existing code works without modification

## Motivation
The existing `ws2812` driver uses CPU-timed bit-banging assembly tuned for Cortex-M0+. The RP2350's Cortex-M33 has a write buffer and different pipeline that breaks the nanosecond-level GPIO timing. The Arduino Adafruit_NeoPixel library works on the same hardware because it uses PIO.

## Implementation
- `ws2812_rp2_pio.go` (`rp2040 || rp2350`): `newWS2812Device()` claims a PIO0 state machine and sets `writeColorFunc` to a PIO-based closure. Falls back to bit-bang on failure.
- `ws2812_init_other.go` (`!rp2040 && !rp2350`): `newWS2812Device()` returns the standard bit-bang device.
- `NewWS2812()` now delegates to `newWS2812Device()`.

## Test plan
- [x] `tinygo flash -target=pico2 ./examples/ws2812/` — verify LEDs work on RP2350 via PIO
- [ ] `tinygo build -target=circuitplay-express ./examples/ws2812/` — verify bit-bang fallback compiles
- [ ] `tinygo build -target=arduino ./examples/ws2812/` — verify AVR still works
- [ ] `tinygo build -target=m5stamp-c3 ./examples/ws2812/` — verify Xtensa/RISC-V still works

### DIFF
--- a/ws2812/ws2812.go
+++ b/ws2812/ws2812.go
@@ -1,4 +1,7 @@
 // Package ws2812 implements a driver for WS2812 and SK6812 RGB LED strips.
+//
+// On most platforms NewWS2812 uses bit-banging.
+// On RP2040/RP2350 it uses PIO for hardware-timed control.
 package ws2812 // import "tinygo.org/x/drivers/ws2812"
 
 //go:generate go run gen-ws2812.go -arch=cortexm 16 48 64 120 125 150 168 200
@@ -24,14 +27,11 @@ func New(pin machine.Pin) Device {
 	return NewWS2812(pin)
 }
 
-// New returns a new WS2812(RGB) driver.
-// It does not touch the pin object: you have
-// to configure it as an output pin before calling New.
+// NewWS2812 returns a new WS2812(RGB) driver.
+// On RP2040/RP2350, it uses PIO for hardware-timed control.
+// On other platforms, you must configure the pin as output before calling this.
 func NewWS2812(pin machine.Pin) Device {
-	return Device{
-		Pin:            pin,
-		writeColorFunc: writeColorsRGB,
-	}
+	return newWS2812Device(pin)
 }
 
 // New returns a new SK6812(RGBA) driver.
--- a/go.mod
+++ b/go.mod
@@ -1,10 +1,8 @@
 module tinygo.org/x/drivers
 
-
 go 1.22.1
 
 toolchain go1.23.1
-
 
 require (
 	github.com/eclipse/paho.mqtt.golang v1.2.0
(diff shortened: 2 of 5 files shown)


.claude/settings.json
```diff
@@ -0,0 +1,43 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(find /Users/gandarez/development/git/gandarez/tinygo-drivers -type f -name *ws2812*)",
```
acouvreur: I don't think we want to keep this file ?

deadprogram:
I was not expecting you to add a new type to implement this, @gandarez 

Seems to me that you could wrap the existing `Device` type, and then use the `writeColorFunc` to do the PIO specific behavior.

.claude/settings.json
gandarez: oops!

gandarez:
@deadprogram is it ok?
https://github.com/tinygo-org/drivers/pull/851/changes/e569fc609894a37a71d71123355824bf15515306

deadprogram:
Sorry, but that was not what I had in mind @gandarez 

I do not see any reason to add a new `Strip` type. That changes the exported interface, when you really only need to extend it.

`NewWS2812(pin machine.Pin)` should be able to call the PIO code by some use of build tags, and possibly adding some init function to the `Device` type. You might be able to do lazy PIO init and avoid that entirely and do it all with build tags + the PIO code.

Hopefully that explains a bit better what I had in mind here.

gandarez:
@deadprogram I believe this is what you're looking for https://github.com/tinygo-org/drivers/pull/851/changes/544cd43fc7d7c53282f1e0960af5517aa126002e. Do you have any other device to try it out?

I also added brightness control in this commit https://github.com/tinygo-org/drivers/pull/851/changes/4b3ba69703ee865fca5aee2ff1758acb3608e988

deadprogram:
> @deadprogram I believe this is what you're looking for [544cd43](https://github.com/tinygo-org/drivers/commit/544cd43fc7d7c53282f1e0960af5517aa126002e). Do you have any other device to try it out?

I think that is a lot more like it.

> I also added brightness control in this commit [4b3ba69](https://github.com/tinygo-org/drivers/commit/4b3ba69703ee865fca5aee2ff1758acb3608e988)

That might be better to add in a separate PR, no?

ws2812/ws2812.go
```diff
@@ -1,4 +1,7 @@
 // Package ws2812 implements a driver for WS2812 and SK6812 RGB LED strips.
+//
+// On RP2040/RP2350, NewWS2812 uses PIO for hardware-timed control.
```
deadprogram: I would phrase this:

On most platforms NewWS2812 uses bit-banging.
On RP2040/RP2350 it uses PIO for hardware-timed control.

gandarez:
@deadprogram PR is ready

deadprogram:
Confirmed that it still works on `samd21` as well. Thank you for adding this, and for the clean implementation! Now merging.

PR #854, opened 2026-04-07
ws2812: add brightness control

gandarez:
## Summary

- Adds a `SetBrightness(b uint8)` method to `Device` for global brightness control (0–255, default 255)
- Introduces `applyBrightness()` helper that scales RGB channels using integer math (`uint16` multiply + bit shift to avoid float ops)
- Updates `writeColorsRGB`, `writeColorsRGBA`, and the RP2040/RP2350 PIO path to apply brightness before writing each color
- `brightness` field defaults to `255` (full brightness), ensuring backwards-compatible behavior for existing users

## Test plan

- [x] Flash to a WS2812 LED strip and verify full brightness at `255`
- [x] Call `SetBrightness(128)` and confirm ~50% perceived brightness
- [x] Call `SetBrightness(0)` and confirm LEDs are off
- [ ] Verify SK6812 RGBW strips: alpha channel is passed through unchanged (not dimmed)
- [x] Verify RP2040/RP2350 PIO path applies brightness correctly via `ws.PutRGB`

### DIFF
--- a/ws2812/ws2812.go
+++ b/ws2812/ws2812.go
@@ -18,7 +18,8 @@ var errUnknownClockSpeed = errors.New("ws2812: unknown CPU clock speed")
 // Device wraps a pin object for an easy driver interface.
 type Device struct {
 	Pin            machine.Pin
-	writeColorFunc func(Device, []color.RGBA) error
+	brightness     uint8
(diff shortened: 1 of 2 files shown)


deadprogram:
Looking good thanks for the addition @gandarez now merging.

PR #856, opened 2026-04-15
fix st7789 scroll on rotated displays

conejoninja:
Fix issue on the st7789 driver when it's small (<320) and is rotated


deadprogram:
Thanks for the fix @conejoninja now merging.

PR #857, opened 2026-04-16
add support extended id for mcp2515

nobonobo:
- add support extended-id frame for mcp2515
- remove depends legacy and pin
- add and use defined errors

### DIFF
--- a/examples/mcp2515/main.go
+++ b/examples/mcp2515/main.go
@@ -21,7 +21,7 @@ func main() {
 		SDI:       machine.SPI0_SDI_PIN,
 		Mode:      0})
 	can := mcp2515.New(spi, csPin)
-	can.Configure()
+	can.Configure(mcp2515.Configuration{})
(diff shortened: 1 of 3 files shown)


deadprogram:
Hello @nobonobo thank you for the PR.

However, I think it was not correct to replace the pin implementation with `machine.Pin`. We have been attempting to remove direct dependencies on the `machine` package, please see https://github.com/tinygo-org/drivers/pull/795 for more info.

nobonobo:
Understood. I'll restore the machine.Pin dependency later!

deadprogram:
> Understood. I'll restore the machine.Pin dependency later!

Why is it that you need `machine.Pin` here? It should no longer be required for use.

nobonobo:
It was "remove," not "restore."
dependency on "machine.Pin" was removed.

deadprogram:
Thanks you very much for the improvement @nobonobo now squash/merging.

Issue #858, opened 2026-04-16
SK6812's colors are off  (driver ws2812)


conejoninja:
This commit breaks my code, i'm using SK6812 MINI-E RGB and a nice!nano (nrf52840) using leds.WriteColors(xxx):

before this commit ( https://github.com/tinygo-org/drivers/commit/21a7d0a96aeb7f55b9c309606992b5350bb7ebd5 ) 
- I used ws2812.New and it worked fine
- ws2812.NewSK6812 worked fine on the first LED, but second LED's colors was off (no red)

after this commit ( https://github.com/tinygo-org/drivers/commit/21a7d0a96aeb7f55b9c309606992b5350bb7ebd5 ) 
- ws2812.New stopped working (LEDs are off)
- ws2812.NewSK6812 worked fine on the first LED, but second LED's colors was off (no red)

My board only have two leds, don't know what will happen with more LED's
In both cases SK6812 is off.

/cc @gandarez as the author of the commit
/cc @SoftTacos as the author of the SK6812 implementation

gandarez:
Thanks for fixing it

PR #859, opened 2026-04-16
fix issue #858

conejoninja:
This seems to fix the issue, but to be fair, I didn't deep into the SK6812 specs, nor I understand why it differs from WS2812.

So if anyone could check I did not break anything, that would be great.

Related issue #858


deadprogram:
Makes sense to me, thanks for the fix @conejoninja now merging.

PR #864, opened 2026-04-23
add SPI driver for semtech sx128x chips

jwetzell:
This is really only an interface directly for the chips themselves and does not provide much in the way of abstractions like LoRa or whatever else. The SPI stuff is heavily inspired by the existing semtech chip drivers but doesn't thrill me and open to any better way to handle all that.


sx128x/sx128x.go
```diff
@@ -0,0 +1,533 @@
...
+	time.Sleep(10 * time.Millisecond)
+	d.resetPin.Set(true)
+	time.Sleep(10 * time.Millisecond)
+}
+
+func (d *Device) WaitWhileBusy() {
+	// TODO(jwetzell): better way to do this?
+	for d.busyPin.Get() {
+	}
+}
```
soypat: Check out the hot-loops section of the driver design guide! https://tinygo.org/docs/guides/driver-design/#hot-loops

jwetzell: I should have searched for my own TODO lol. But that site definitely helped answer the question of "better way?". Did some measures on longest real world busy count and got mid 400s.

sx128x/sx128x.go
```diff
@@ -0,0 +1,637 @@
...
+func (d *Device) Reset() {
+	d.resetPin.Set(false)
+	time.Sleep(10 * time.Millisecond)
+	d.resetPin.Set(true)
+	time.Sleep(10 * time.Millisecond)
+}
+
+func (d *Device) WaitWhileBusy() error {
+	// largest busy period is on boot with around ~400ish this should be more than enough
+	retries := 1000
```
HattoriHanzo031: Timeout will be heavily platform dependent, I would suggest defining exact timeout duration, and then using `start := time.Now()` and `time.Since(start)` to check for timeout.

sx128x/sx128x.go
```diff
@@ -0,0 +1,637 @@
...
+	for retries > 0 && d.busyPin.Get() {
+		runtime.Gosched()
+		retries--
+	}
+	if retries == 0 {
+		return errors.New("busy pin timeout")
```
HattoriHanzo031: Define all static errors as a globals to avoid allocations.

sx128x/sx128x.go
```diff
@@ -0,0 +1,637 @@
...
+	}
+	return nil
+}
+
+func (d *Device) GetStatus() (uint8, uint8, error) {
```
HattoriHanzo031: Comments with the descriptions of all the methods and arguments would be useful. It could mostly be copied from datasheet.

sx128x/sx128x.go
```diff
@@ -0,0 +1,637 @@
...
+	d.spiRxBuf = d.spiRxBuf[:len(d.spiTxBuf)]
+	err = d.spi.Tx(d.spiTxBuf, d.spiRxBuf)
+	d.nssPin.Set(true)
+	if err != nil {
+		return nil, err
+	}
+	return d.spiRxBuf[3:], nil
+}
+
+func (d *Device) SetSleep(sleepConfig uint8) error {
```
HattoriHanzo031: It would be nicer if all of these parameters like sleepConfig, standbyConfig etc. would be custom types (enums) with defined values. It is not a small amount of work so I would like to hear your opinion (and also from other reviewers) if it would be worth it.

jwetzell: I did consider that but wasn't sure how far to take it. Should the commands and registers be types as well?

HattoriHanzo031: I would use types only for the arguments of exposed functions. If the types are defined, we could probably avoid sanity checks like `if sleepConfig > 3` because compiler would allow only defined constants and values explicitly casted to a type

jwetzell: But the type has to be something under the hood. Is there really a way to create a type that can only be the numbers 0-3?

HattoriHanzo031: Go does not have strong guarantee that the type will only have certain values, but I think the standard pattern for "enums" is good enough. For example if we have this type and function:
```
type SleepMode uint8

const (
	SleepModeRetainNone         = SleepMode(0)
	SleepModeRetainRam          = SleepMode(SLEEP_DATA_RAM_RETAIN)
	SleepModeRetainBuffer   = SleepMode(SLEEP_DATA_BUFFER_RETAIN)
	SleepModeRetainRamAndBuffer = SleepMode(SLEEP_DATA_RAM_RETAIN | SLEEP_DATA_BUFFER_RETAIN)
)

func (d *Device) SetSleep(mode SleepMode) error {
	...
```
it could be called like this:
```
d.SetSleep(SleepModeRetainRam)
```
and also like this: 
```
d.SetSleep(SleepMode(5))
```
but in my opinion, this is fine, because for the second case to happen, someone must explicitly cast a value.

What is not allowed in this case is:
```
d.SetSleep(5) // compile error
```
or even
```
d.SetSleep(1) // compile error
```
EDIT: added compile error examples

jwetzell: I don't think the stuff about compile errors [ is correct](https://go.dev/play/p/lDAQdcDUhl5).

HattoriHanzo031: > I don't think the stuff about compile errors [ is correct](https://go.dev/play/p/lDAQdcDUhl5).

Sorry, my bad, forgot about untyped constants. Lack of proper enums is the one thing I don't like about Go and it makes creating contracts like this very difficult.

Adding types would in any case increase user experience of the API. Would it remove the need for runtime value checks is questionable. I would lean a bit more on the side that it is safe to remove them, but it is ok either way.

jwetzell: I agree on the types, and I'll be working on that change.

sx128x/sx128x.go
jwetzell: Ok, I've switch to using a time based timeout in `WaitWhileBusy`

sx128x/sx128x.go
jwetzell: This has been replaced with a time based timeout

sx128x/sx128x.go
jwetzell: this has been addressed in the latest commit

sx128x/sx128x.go
jwetzell: this has been addressed in the latest commit

sx128x/errors.go
```diff
@@ -0,0 +1,19 @@
...
+	ErrInvalidStandbyConfig  = errors.New("invalid standby config")
+	ErrFrequencyTooLow       = errors.New("frequency below 2.4Ghz")
+	ErrFrequencyTooHigh      = errors.New("frequency above 2.5Ghz")
+	ErrPowerTooLow           = errors.New("power level below -18dBm")
+	ErrPowerTooHigh          = errors.New("power level above 13dBm")
+	ErrInvalidPeriodBase     = errors.New("invalid period base")
+	ErrInvalidPacketType     = errors.New("invalid packet type")
+	ErrInvalidRegulatorMode  = errors.New("invalid regulator mode")
+	ErrPayloadLengthTooShort = errors.New("payload length too short")
+	ErrPayloadLengthTooLong  = errors.New("payload length too long")
```
soypat: Don't export these

sx128x/constants.go
```diff
@@ -0,0 +1,410 @@
...
+
+type IRQMask = uint16
+
+type CircuitMode = uint8
+type CommandStatus = uint8
+
+// Packet Status
+type GFSKPacketInfo = uint8
+type BLEPacketInfo = uint8
+type FLRCPacketInfo = uint8
```
soypat: When you expose types and values you add more interaction dimension to the package and can add friction to package use. In Go you export identifiers like constants, types and functions by making them start with upper case.

Examples :
- Exported OpMode, useful for users: https://github.com/soypat/lora/blob/7b9fb67069ebcba68326d8410d1132c3d94d75db/sx127x/registers.go#L235-L256
- Unexported irqPosFlags type, not so useful for users :https://github.com/soypat/lora/blob/7b9fb67069ebcba68326d8410d1132c3d94d75db/sx127x/registers.go#L18-L53 
- Unexported registers, if API is well defined users should not depend on these. You may choose to export them but it can add so much package identifier noise that it makes looking for identifiers to understand package use very hard. A potential middle-ground solution is to have a package that stores registers, i.e `sx127xreg`: https://github.com/soypat/lora/blob/7b9fb67069ebcba68326d8410d1132c3d94d75db/sx127x/registers.go#L70

sx128x/errors.go
jwetzell: You are going to have to give a reason for me to be sold. My argument would be error comparison via something like `errors.Is()` would not be possible if these were unexported.

sx128x/constants.go
jwetzell: I could see the type itself not being exported but the constants from the type would almost definitely need to be or else the package is useless. Someone calling `ReadRegister(...)` if all the registers constants are internal is just going to have to go back to the datasheet because the convenient constants are all internal. The IRQ mask constants being unexported would also mean package users would not be able to use them to construct whatever mask they want when making calls to `SetDioIrqParams` or `ClearIrqStatus`.

jwetzell: The command constants (`CMD_*`) probably do fall into the camp of something that doesn't need to be exported as they are only used internally and there is no real use for them outside of the package.

soypat: > I could see the type itself not being exported but the constants from the type would almost definitely need to be or else the package is useless.

Well that's the excercise! Determine identifiers users need to get the expected functionality of the driver. If users are expected to interact with registers to use the device then I'd argue the HAL is not well defined and that the driver needs more work. We don't call out to MCU registers to use SPI bus in Tinygo because we've done the hard work of well defining the HAL.

> The command constants (CMD_*) probably do fall into the camp of something that doesn't need to be exported as they are only used internally and there is no real use for them outside of the package.

This is very likely the case.

sx128x/errors.go
soypat: How would users use these errors in a real world scenario?

sx128x/constants.go
jwetzell: This is simply an implementation of the SPI interface in the datasheet. This does not address what making a higher level radio device would look like (LoRa, GFSK, etc). The registers are user facing they are not used by this driver internally so if unexported they might as well be removed.

sx128x/errors.go
jwetzell: call function that might return an error if it does check it against known errors and react however the user wants to for specific error cases (this is very standard Go and all over the place in big Go see `fs`, `io`, `net`, etc). 

```
err := device.SetTx(....)

if err != nil {
  if errors.Is(err, sx128x.ErrBusyPinTimeout) {
    // something might be up with the chip user might reset it here and try again 
  } else {
   // user could do more error comparing as functions can return other kinds
  }
}
```

I'm open to suggestions on better errors and maybe distinguishing which ones are "user-actionable" and thus should be exported but I very much dislike when a function can return multiple different errors and I (the user) have no way to distinguish them besides string comparison.

jwetzell: I've changed to export only the busy pin timeout error as the others don't seem to have the same kind of user-"actionable" nature

sx128x/constants.go
jwetzell: Still working through cleaning this up found a couple of types that had identical constants so could be combined. Still not quite sure what the arguments for export vs not on types is. I could imagine a package user wanting to use the types themselves maybe in a config struct or something like that.

sx128x/sx128x.go
```diff
@@ -0,0 +1,768 @@
...
+		if time.Since(now) > timeout {
+			return ErrBusyPinTimeout
+		}
+		runtime.Gosched()
+	}
+	return nil
+}
+
+// Get tranceiver status, returns circuit mode and command status
+func (d *Device) GetStatus() (CircuitMode, CommandStatus, error) {
```
soypat, line 51: Why not make these Mode, Status and Config types their own types instead of type aliases of uint8s? I believe it would make the user's life easier to keep track of these values since the type would be lost on calling GetStatus in this PR's current state.

soypat:
Also it'd be great to have an example in the `examples/sx128x` directory since you already have experience using these!

sx128x/sx128x.go
jwetzell: What does that look like?

jwetzell:
The only example I have working is very crude LoRa Tx/Rx because of the state of LoRa in the tinygo universe. I can try to get one of them looking not too bad.

sx128x/sx128x.go
soypat: type alias:
```go
type SleepConfig = uint8
```
user/local type:
```go
type SleepConfig uint8
```

Try to make all type aliases local types if possible is what I'd suggest. Local types are more explcit. type aliases are very loosey goosey on use. See sx128x/constants.go for the location of most/all of your type aliases

soypat: this is how you'd rewrite some methods which use the uint8 argument receiving function `SetModulationParams`:
```go
func (d *Device) SetModulationParamsBLE(bitrateBandwidth GFSKBLEBitrateBandwidth, modulationIndex ModulationIndex, modulationShaping ModulationShaping) error {
	return d.SetModulationParams(uint8(bitrateBandwidth), uint8(modulationIndex), uint8(modulationShaping))
}
```

soypat:
Crude is OK! Should be a starting point for users

sx128x/sx128x.go
jwetzell: Oh dang to be honest the latter is what I thought I did. Honestly didn't realize there was a distinction.

soypat (approved):
Looks great!

HattoriHanzo031 (approved):
🚀

soypat:
Thanks for your contribution @jwetzell and @HattoriHanzo031 for reviews!

soypat:
On a side note, it'd be nice to have the example be more polished @jwetzell. That means making all those obscure WriteRegister calls more explicit. Ideally there should be a Configure method that takes all the parameters needed in struct form to configure the LoRa correctly for use.

jwetzell:
@soypat I'll make a global level struct that the `SetupLora` will look at

jwetzell:
@soypat I'll open the polished examples as a separate PR #865 since this already got merged. Also have to figure out how to fix my VS Code marking every commit as "Co-Authored by Copilot".

PR #866, opened 2026-05-05
TX power in eu868 regional parameters is now in line with specs

jhelberg:
the TX power was off. It is in line with the lora specs now.


deadprogram:
Thank you for the correction @jhelberg now squash/merging.

deadprogram:
I made a mistake in merging this to `release` instead of `dev` branch, but I corrected that via some force pushes. Now merged to the correct branch.

PR #870, opened 2026-06-04
bme688 driver and example

hoanga:
Hello!

This is a driver for the BME688 sensor that is included on the Pimoroni enviro indoor. I have tested this on an available enviro board and confirmed everything but the gas sensors are functional.

links:
https://github.com/pimoroni/enviro
https://shop.pimoroni.com/products/enviro-indoor

### DIFF
--- a/bme688/bme688.go
+++ b/bme688/bme688.go
@@ -1,0 +1,491 @@
+// Package bme688 provides a driver for the BME688 digital environmental sensor
+// by Bosch, which measures temperature, pressure, humidity and gas (VOC).
+//
+// # It is available in the Pimoroni Enviro indoor board
+//
+// Datasheet: https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bme688-ds000.pdf
+// Pimoroni Driver
+//
+//	https://github.com/pimoroni/pimoroni-pico/tree/main/examples/breakout_bme688
+package bme688
+
+import (
+	"time"
+
+	"tinygo.org/x/drivers"
+	"tinygo.org/x/drivers/internal/legacy"
+)
+
+// calibrationCoefficients holds the device calibration data read at startup.
+type calibrationCoefficients struct {
+	// Temperature
+	t1 uint16
+	t2 int16
+	t3 int8
+	// Pressure
+	p1  uint16
+	p2  int16
+	p3  int8
+	p4  int16
+	p5  int16
+	p6  int8
+	p7  int8
+	p8  int16
+	p9  int16
+	p10 uint8
+	// Humidity
+	h1 uint16
+	h2 uint16
+	h3 int8
+	h4 int8
+	h5 int8
+	h6 uint8
+	h7 int8
+	// Gas heater
+	gh1          int8
+	gh2          int16
+	gh3          int8
+	resHeatRange uint8
+	resHeatVal   int8
+	rangeSWErr   int8
+	// Running fine temperature value shared by T/P/H calculations
+	tFine int32
+}
+
+// Config holds sensor configuration settings.
+type Config struct {
+	Temperature Oversampling
+	Pressure    Oversampling
+	Humidity    Oversampling
+	IIR         Filter
+	// Gas heater settings
+	GasEnabled     bool
+	HeaterTemp     uint16 // target heater temperature in °C (max 400)
+	HeaterDuration uint16 // heater-on duration in ms
+	AmbientTemp    int8   // ambient temperature for heater resistance calculation
+}
+
+// Measurement holds a complete set of sensor readings.
+type Measurement struct {
+	Temperature   int32  // milli-degrees Celsius (e.g. 25000 = 25.000 °C)
+	Pressure      uint32 // Pascals
+	Humidity      uint32 // relative humidity × 1000 (e.g. 50000 = 50.000 %)
+	GasResistance uint32 // Ohms
+	GasValid      bool   // true when gas reading is valid
+	HeaterStable  bool   // true when heater reached target temperature
+}
+
+// Device wraps an I2C connection to a BME688 (or BME680) device.
+type Device struct {
+	bus       drivers.I2C
+	Address   uint16
+	variantID byte
+	calib     calibrationCoefficients
+	Config    Config
+	// Cached values from the last Update call.
+	lastTemp          int32
+	lastPressure      uint32
+	lastHumidity      uint32
+	lastGasResistance uint32
+	lastGasValid      bool
+	lastHeaterStable  bool
+}
+
+// New creates a new BME688 driver. The I2C bus must already be configured.
+// This only creates the Device object; it does not communicate with the sensor.
+func New(bus drivers.I2C) Device {
+	return Device{
+		bus:     bus,
+		Address: Address,
+	}
+}
+
+// Configure sets up the sensor with default settings and reads calibration data.
+// Defaults: 2× oversampling for T/P/H, no IIR filter, gas enabled at 320 °C / 150 ms.
+func (d *Device) Configure() {
+	d.ConfigureWithSettings(Config{})
+}
+
+// ConfigureWithSettings sets up the sensor with the provided configuration.
+// Zero-value fields revert to the defaults described in Configure.
+func (d *Device) ConfigureWithSettings(config Config) {
+	d.Config = config
+	if d.Config == (Config{}) {
+		d.Config = Config{
+			Temperature:    Sampling2X,
+			Pressure:       Sampling2X,
+			Humidity:       Sampling2X,
+			IIR:            FilterOff,
+			GasEnabled:     true,
+			HeaterTemp:     320,
+			HeaterDuration: 150,
+			AmbientTemp:    25,
+		}
+	}
+	if d.Config.AmbientTemp == 0 {
+		d.Config.AmbientTemp = 25
+	}
+
+	// Read variant ID to handle BME680 vs BME688 run_gas bit difference.
+	var vbuf [1]byte
+	legacy.ReadRegister(d.bus, uint8(d.Address), REG_VARIANT_ID, vbuf[:])
+	d.variantID = vbuf[0]
+
+	d.Reset()
+	time.Sleep(10 * time.Millisecond)
+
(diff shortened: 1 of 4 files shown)


deadprogram:
Hello @hoanga thank for the addition. Can you please add a smoketest to https://github.com/tinygo-org/drivers/blob/dev/smoketest.sh

deadprogram:
Also please remove `Assisted-by: Claude Code`

We only care about the human who submitted the PR, not the tools used.

hoanga:
Hello, I have made updates per suggestions. Hopefully, I understood the request properly. I left the note in an (over)attempt to keep attributions however from the explanation regarding tooling, makes sense.

If the PRs end up conflicting because of the smoketest add, I will can fix the aftermath depending on which PR goes in first.

deadprogram (commented):
Thanks for this driver. The structure follows the repo conventions well, and it keeps the same API shape as `bme280`. The calibration parse, `calcResHeat` and `calcGasWait` agree with the Bosch BME68x API.

I put the details in inline comments. Two of them explain why the gas sensor does not work: the driver asks the sensor for a high-range gas measurement, but it reads and calculates the low-range result. The other comments are smaller.

bme688/bme688.go
```diff
@@ -0,0 +1,491 @@
...
+	d.calib.rangeSWErr = int8((c3[4] & 0xF0) >> 4)
+}
+
+// parseField converts a raw 17-byte field into a Measurement.
+func (d *Device) parseField(f [LEN_FIELD]byte) Measurement {
+	adcP := uint32(f[2])<<12 | uint32(f[3])<<4 | uint32(f[4])>>4
+	adcT := uint32(f[5])<<12 | uint32(f[6])<<4 | uint32(f[7])>>4
+	adcH := uint16(f[8])<<8 | uint16(f[9])
+	adcG := uint16(f[13])<<2 | uint16(f[14]>>6)
+	gasRange := f[14] & GAS_RANGE_MSK
```
deadprogram, line 343: The gas result is always read from the low-range bytes. This is one of the two reasons the gas sensor does not work on a BME688.

`Configure` sets `RUN_GAS_688` (bit 5), so the sensor makes a high-range measurement. But for variant `0x01` the Bosch API reads the gas ADC from field bytes 15 and 16, `gas_range` from byte 16, and the `GASM_VALID` and `HEAT_STAB` bits also from byte 16. Bytes 13 and 14 hold the low-range result only.

See `read_field_data()` in the Bosch BME68x API. The parse must select the byte pair from `d.variantID`, and lines 353 to 354 must use the same byte.

bme688/bme688.go
```diff
@@ -0,0 +1,491 @@
...
+	}
+	if compHum > 100000 {
+		compHum = 100000
+	}
+	return uint32(compHum)
+}
+
+// calcGasResistance returns gas resistance in Ohms.
+// Formula from Bosch BME68x API using integer lookup tables.
+func (d *Device) calcGasResistance(adcG uint16, gasRange byte) uint32 {
```
deadprogram, line 425: This is the low-range formula (`calc_gas_resistance_low`), but the BME688 needs the high-range one. This is the second reason the gas sensor does not work.

The Bosch API uses `calc_gas_resistance_high` when `variant_id == 0x01`:

```go
var1 := uint32(262144) >> gasRange
var2 := (int32(adcG)-512)*3 + 4096
return uint32(1000000 * int32(var1) / var2)
```

Please keep the low-range version for the BME680 and select on `d.variantID`. Then `gasLookupTable1` and `gasLookupTable2` stay necessary.

bme688/bme688.go
```diff
@@ -0,0 +1,491 @@
...
+
+	// Gas heater coefficients
+	d.calib.gh2 = int16(uint16(c2[11])<<8 | uint16(c2[10]))
+	d.calib.gh1 = int8(c2[12])
+	d.calib.gh3 = int8(c2[13])
+
+	// Gas calibration extras (coeff3 block, from 0x00)
+	d.calib.resHeatVal = int8(c3[0])
+	d.calib.resHeatRange = (c3[2] & 0x30) >> 4
+	d.calib.rangeSWErr = int8((c3[4] & 0xF0) >> 4)
```
deadprogram, line 334: `range_sw_err` is signed, but this shift is unsigned, so the top nibble does not sign-extend. The Bosch API does a signed shift.

```go
d.calib.rangeSWErr = int8(c3[4]) >> 4
```

This changes the low-range gas result only, but it is a transcription error.

bme688/bme688.go
```diff
@@ -0,0 +1,491 @@
...
+	var field [LEN_FIELD]byte
+	for i := 0; i < 10; i++ {
+		legacy.ReadRegister(d.bus, uint8(d.Address), REG_FIELD_0, field[:])
+		if field[0]&NEW_DATA_MSK != 0 {
+			break
+		}
+		time.Sleep(5 * time.Millisecond)
+	}
+
+	return d.parseField(field), nil
```
deadprogram, line 205: If the new-data bit never becomes set, the loop ends and `parseField` runs on the still-zero `field` array. The caller then gets a nonsense temperature with a nil error.

Please return an error after the last try. At the moment `Read` never returns a non-nil error, so the error branch in the example is dead code.

bme688/bme688.go
```diff
@@ -0,0 +1,491 @@
...
+// Defaults: 2× oversampling for T/P/H, no IIR filter, gas enabled at 320 °C / 150 ms.
+func (d *Device) Configure() {
+	d.ConfigureWithSettings(Config{})
+}
+
+// ConfigureWithSettings sets up the sensor with the provided configuration.
+// Zero-value fields revert to the defaults described in Configure.
+func (d *Device) ConfigureWithSettings(config Config) {
+	d.Config = config
+	if d.Config == (Config{}) {
```
deadprogram, line 113: The comment and the code do not agree. The test is for the complete struct, so the defaults apply only when every field is zero.

If a user sets `Temperature: Sampling16X` and nothing else, pressure and humidity oversampling become off and the gas heater becomes disabled, without a message. Please either give each field its own default, or correct the comment.

bme688/bme688.go
```diff
@@ -0,0 +1,491 @@
...
+	tFine := var2 + var3
+	// (tFine*5+128)>>8 gives temperature in 0.01 °C; ×10 → milli-°C
+	return 10 * ((tFine*5 + 128) >> 8), tFine
+}
+
+// calcPressure returns pressure in Pascals.
+// Formula from Bosch BME68x API (integer variant).
+func (d *Device) calcPressure(adcP uint32) uint32 {
+	var1 := d.calib.tFine/2 - 64000
+	var2 := ((var1/4)*(var1/4))>>11*int32(d.calib.p6)/2 +
```
deadprogram, line 374: This term divides by 2, but the Bosch API shifts it right by 2 (divides by 4).

I compared the two versions with real calibration values. The difference is 1 to 2 Pa near 25 °C, and approximately 15 Pa at -20 °C.

bme688/bme688.go
```diff
@@ -0,0 +1,491 @@
...
+		legacy.WriteRegister(d.bus, uint8(d.Address), REG_RES_HEAT_0, []byte{resHeat})
+		legacy.WriteRegister(d.bus, uint8(d.Address), REG_GAS_WAIT_0, []byte{gasWait})
+
+		// Enable gas measurement on profile 0.
+		var runGasBit byte = RUN_GAS_688
+		if d.variantID == VARIANT_BME680 {
+			runGasBit = RUN_GAS_680
+		}
+		legacy.WriteRegister(d.bus, uint8(d.Address), REG_CTRL_GAS_1,
+			[]byte{runGasBit | 0x00}) // nb_conv = 0
```
deadprogram, line 163: `| 0x00` has no effect. The comment gives the necessary information.

bme688/bme688.go
```diff
@@ -0,0 +1,491 @@
...
+	sampleCounts := [6]int{0, 1, 2, 4, 8, 16}
+
+	clamp := func(o Oversampling) int {
+		if int(o) < len(sampleCounts) {
+			return sampleCounts[o]
+		}
+		return 16
+	}
+
+	µs := overhead +
```
deadprogram, line 481: Please use an ASCII name here, for example `us` or `durUS`. The rest of the repo keeps identifiers ASCII.

bme688/bme688.go
```diff
@@ -0,0 +1,491 @@
...
+	h7 int8
+	// Gas heater
+	gh1          int8
+	gh2          int16
+	gh3          int8
+	resHeatRange uint8
+	resHeatVal   int8
+	rangeSWErr   int8
+	// Running fine temperature value shared by T/P/H calculations
+	tFine int32
```
deadprogram, line 52: `tFine` is per-measurement state, not calibration data. It also makes `parseField` write to `d.calib`, which is not obvious.

`bme280` gives `tFine` to the calculation functions as a parameter. Please do the same here.

bme688/registers.go
```diff
@@ -0,0 +1,116 @@
...
+const (
+	Address    uint16 = 0x77
+	AddressAlt uint16 = 0x76
+)
+
+// Chip identifiers
+const (
+	CHIP_ID        = 0x61
+	VARIANT_BME680 = 0x00
+	VARIANT_BME688 = 0x01
```
deadprogram, line 13: These declarations have no user: `VARIANT_BME688`, `GAS_MEAS_MSK`, `MEASURING_MSK`, `GAS_IDX_MSK`, `ModeSleep`, and `REG_MEAS_STATUS_0` (same address as `REG_FIELD_0`).

`VARIANT_BME688` and `ModeSleep` become necessary if you add the high-range gas path. Please remove the others.

examples/bme688/main.go
```diff
@@ -0,0 +1,47 @@
...
+	// Uncomment if default does not work
+	//sensor.Address = bme688.AddressAlt
+	sensor.Configure()
+
+	connected := sensor.Connected()
+	if !connected {
+		println("BME688 not detected. Exiting...")
+	}
+
+	println("BME688 detected")
```
deadprogram, line 24: If the sensor does not answer, the example prints the message and then continues into the read loop. Please add a `return`.

PR #875, opened 2026-06-12
ws2812: fix PIO TX FIFO overflow on rp2040/rp2350 dropping LEDs

### LINKED ISSUE #869: [rp2040/pio] WS2812b not working properly in 0.35, fine in 0.34
I'm using a XIAO RP2040 I got 20 ws2812b LEDS (in reality is SK6812, but the driver is the same).

- Using 0.35.0 only 9 LEDs light up
- Using 0.34.0 all LEDs light up (as expected)

In 0.35.0 it was added PIO support, so maybe the issue is there

deadprogram:
PutRGB calls TxPut which is non-blocking and silently discards data when the TX FIFO is full.

Wait for FIFO space before each PutRGB using runtime.Gosched() to yield cooperatively, matching the pattern used by piolib.WriteRaw.

Should fix https://github.com/tinygo-org/drivers/issues/869

### DIFF
--- a/ws2812/ws2812_rp2_pio.go
+++ b/ws2812/ws2812_rp2_pio.go
@@ -5,6 +5,7 @@ package ws2812
 import (
 	"image/color"
 	"machine"
+	"runtime"
 
 	pio "github.com/tinygo-org/pio/rp2-pio"
 	"github.com/tinygo-org/pio/rp2-pio/piolib"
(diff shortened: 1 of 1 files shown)


soypat (approved):

deadprogram:
Thank you for review @soypat now merging.

PR #884, opened 2026-08-10
uc8151: fix LUT timing calculations

jakebailey:
I hit a few bugs while testing out the Badger 2040 W that @deadprogram gave me at GopherCon :)

The LUT timing calculations had a few small Python-to-Go translation errors (`**` vs `^`, `//` vs `%`, and a boundary condition). In particular, `FAST` caused a divide-by-zero panic.

Tested on the actual badge with fast, flicker-free page updates.


deadprogram:
Thank you for the fix @jakebailey now merging.

PR #887, opened 2026-08-31
Initial support for ADS1015 (4-channel 12-bit ADC)

conejoninja:
Initial support for ADS1015 (4-channel 12-bit ADC) used in the NiceCtrlr

It's been a while since I made a driver, so I'm a bit rusty, let me know if something is missing.

Disclaimer: I used the assistance of ClaudeCode

### DIFF
--- a/ads1015/ads1015.go
+++ b/ads1015/ads1015.go
@@ -11,13 +11,11 @@ import (
 	"tinygo.org/x/drivers"
 )
 
-// maxCode is the magnitude of a full-scale 12-bit conversion result
+// The datasheet defines the LSB as FS/2048, but we divide by the maximum
+// positive code (2047) so that the maximum ADC output maps to full scale.
 const maxCode = 2047
 
-var (
-	conversionPollInterval = 1 * time.Millisecond
-	conversionTimeout      = 100 * time.Millisecond
-)
+const conversionTimeout = 100 * time.Millisecond
 
 var (
 	ErrInvalidChannel = errors.New("ads1015: invalid channel")
@@ -43,6 +41,8 @@ type Config struct {
 	ComparatorPolarity ComparatorPolarity
 	ComparatorLatch    ComparatorLatch
 	ComparatorQueue    ComparatorQueue
+
+	conversionPollInterval time.Duration
 }
 
 // DefaultConfig selects the widest gain range (+/-6.144V), single-shot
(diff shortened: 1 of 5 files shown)


deadprogram:
Thanks for working on this @conejoninja

Here are a few items found during an automated review:

### Issues

1. **`Read()` waits too little in continuous mode** (`ads1015.go`). The code sleeps `conversionPollInterval` (1 ms) after a mux change. One conversion needs 7.8 ms at 128 SPS and 4 ms at 250 SPS, thus `Value()` can return the result of the previous mux setting. Calculate the delay from `d.config.DataRate`, or poll.

2. **`Ready()` doc is not correct** (`ads1015.go`). The comment says "In continuous mode it always reports true". The OS bit reads 0 while a conversion is in progress, and in continuous mode the device converts continuously, thus the function usually returns false. The code is not affected, because `Read()` does not call `Ready()` in continuous mode, but the doc can mislead.

3. **`ToVoltage()` divides by 2047** (`ads1015.go`). The datasheet gives an LSB of FS/2048 (3 mV at +/-6.144 V). The divisor 2047 adds an error of approximately 0.05% at full scale. The RobTillaart library does the same, thus this is possibly intentional. If it is, add a short note.

4. **Possible race in single-shot mode** (`ads1015.go`). `Read()` polls `Ready()` immediately after `startConversion()`. If the device did not yet clear the OS bit, the first poll can report ready and `Value()` returns the previous result. One sleep before the first poll prevents this.

5. **Dead commented-out constants** (`registers.go`). The two `Gain0256mV` lines (0b110, 0b111) and the second `DataRate3300SPS` line are commented-out code. Keep only a one-line note that these codes repeat the last setting.

6. **Address comment does not agree with the code** (`registers.go`). The comment says "see the datasheet for the other three options (0x49, 0x4A, 0x4B)", but `Address2`, `Address3`, and `Address4` define them on the next lines. Also, names that show the ADDR pin connection (for example `AddressADDRtoVDD`) are more clear than a number.

7. **Magic value** (`ads1015.go`). `SetThresholdHigh(-32768)` and `SetThresholdLow(0)` are the special conversion-ready values. Named constants make the intent more clear than the literal.

8. **No unit test.** 27 of 137 driver directories have tests, thus this is not a repo rule. A small fake I2C bus can test `Value()` sign extension, `ToVoltage()`, and the config word that `startConversion()` writes.

Item 1 is the only defect that gives incorrect data. Items 5 and 6 are quick cleanups.

conejoninja:
Thanks, issues has been addressed (I think). Claude helped me (a lot) with the tests, if that's not allowed, let me know and I'll rewrite them again.

PR #888, opened 2026-09-01
docs: add instructions to use ASD-STE-100 Simplified Technical English

deadprogram:
This PR adds instructions to use ASD-STE-100 Simplified Technical English like other TinyGo repos.


deadprogram:
Merging since it matches other TinyGo repos.

PR #890, opened 2026-09-10
examples: add 2.13in e-Paper V1 demo for RP2040-PiZero

davecheney:
Adds a hardware bring-up demo for the Waveshare 2.13in e-Paper HAT V1 driven by a Waveshare RP2040-PiZero.

IMG_2514

## What it does

The demo clears the panel, then draws a test pattern chosen so that a partial result localises the fault rather than just reporting failure:

- a 2 pixel border, which proves the full addressable area
- two diagonals, which show any row or column stride error
- a solid black block, which shows the darkest level the panel reaches
- an eight step dither ramp, which exercises the LUT

It logs progress over USB serial and then enters deep sleep, since an e-paper panel holds its image without power.

## Scope

The header comment is explicit about what this is verified against, because the driver is not interchangeable across panel revisions. `epd2in13` sends the IL3820 init sequence and a 30 byte LUT. V2 and later use the SSD1680, which needs a different init sequence and a 153 byte LUT, so this demo will not drive them. The comment lists the panel revisions and boards as known working, untested, and not supported, and explains how to identify a panel from the ribbon cable date code.

## Verification

Flashed to real hardware with `tinygo flash -target=pico -monitor`. The panel rendered the complete pattern, which confirms SPI1 clock and data, the CS, DC, RST and BUSY pins, the IL3820 init sequence, and the RAM write. Serial output was:

```
clear the display
draw the test pattern
done, the display holds the image without power
```

`gofmt` is clean and `tinygo build -target=pico` succeeds.

One caveat worth flagging for reviewers: only the RP2040-PiZero with a V1 panel is marked as tested. The suggestion in the comment to try `epd2in9v2` for SSD1680 panels is a starting point rather than a verified path, since that package targets a 2.9in panel and would need Width and Height overrides. I did not have the hardware to confirm it.

### DIFF
--- a/examples/waveshare-epd/epd2in13-pizero/main.go
+++ b/examples/waveshare-epd/epd2in13-pizero/main.go
@@ -1,0 +1,178 @@
+// Demo for the Waveshare 2.13in e-Paper HAT V1 on a Waveshare RP2040-PiZero.
+//
+// Panel version.
+//
+// This demo is only for the V1 panel, which uses the IL3820 controller. It
+// uses the epd2in13 driver, which sends the IL3820 init sequence and a 30
+// byte LUT. Later panels use the SSD1680 controller, which needs a different
+// init sequence and a 153 byte LUT, so this demo does not drive them.
+//
+// Identify the panel from the date code on the ribbon cable. V1 shipped from
+// 2015, V2 from 2019 and V3 and V4 from 2021.
+//
+//	V1 (IL3820,  122x250)  works, this demo
+//	V2 (SSD1680, 122x250)  not supported here, try the epd2in9v2 driver
+//	V3 (SSD1680, 122x250)  not supported here, try the epd2in9v2 driver
+//	V4 (SSD1680, 122x250)  not supported here, try the epd2in9v2 driver
+//	2.13in B, C and D variants (three colour or flexible) not supported here,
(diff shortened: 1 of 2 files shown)


examples/waveshare-epd/epd2in13-pizero/main.go
```diff
@@ -0,0 +1,159 @@
...
+func fill(x, y, w, h int16, c color.RGBA) {
+	for i := x; i < x+w; i++ {
+		for j := y; j < y+h; j++ {
+			setPixel(i, j, c)
+		}
+	}
+}
+
+func setPixel(x, y int16, c color.RGBA) {
+	if x < 0 || x >= epdWidth || y < 0 || y >= epdHeight {
```
conejoninja: display.SetPixel already does this check (with rotation if needed), maybe it could be removed from the example and use display.SetPixel directly instead

davecheney: let me give that a try, and also i'll double check all the rotations work

davecheney:
@conejoninja PTAL, i also verified rotation was working 

https://github.com/user-attachments/assets/338156b4-6733-4f9a-9580-5f2246f1c269

conejoninja:
Thanks for your contribution! Merged

PR #891, opened 2026-09-10
waveshare-epd: add epd2in7 driver and RP2040-PiZero slideshow demo

davecheney:
Add a driver for the Waveshare 2.7in V1 e-Paper HAT. This panel uses an IL91874-style controller, different from the IL3820 and SSD1680 controllers used by the other panels in this package. The controller also needs a GET_STATUS command before each BUSY read, or the BUSY pin never updates.

Add a slideshow example for the Waveshare RP2040-PiZero. Each picture is a packed 1 bit per pixel asset file under assets/, embedded with go:embed, and the firmware cycles through the files in name order, one image every 10 seconds.

The scaffold ships with four public domain pictures, dithered to 1 bit: the 1984 Macintosh hello screen, The Great Wave off Kanagawa, a 1972 Soviet space era stamp, and a guinea pig photo.

Verified on real hardware: RP2040-PiZero with the 2.7in V1 HAT on the standard 40 pin header.

https://github.com/user-attachments/assets/80eee7ed-e7e7-4af1-bf55-80d2c326268e


davecheney:
@conejoninja @soypat /wave

waveshare-epd/epd2in7/epd2in7.go
```diff
@@ -0,0 +1,372 @@
...
+	d.SendData(0x00)
+
+	d.SendCommand(POWER_ON)
+	d.WaitUntilIdle()
+
+	d.SendCommand(PANEL_SETTING)
+	d.SendData(0xAF) // KW-BF   KWR-AF    BWROTP 0f
+
+	d.SendCommand(PLL_CONTROL)
+	d.SendData(0x3A) // 3A 100Hz   29 150Hz   39 200Hz   31 171Hz
```
conejoninja, line 133: Does it make sense to put those values in a constant? like PLL_CONTROL_100HZ and such? this only happens once and can not be changed later, so I'm not sure what is the best action. Same with KW-BF, KWR-AF,...

deadprogram: @davecheney agree with @conejoninja about magic numbers. :magic_wand:

PR #895, opened 2026-09-21
scd4x: adhere to the datasheet better

sparques:
I was having issues using the scd4x driver; compared what it was doing to what the datasheet says should be done and fixed all the places where the two differed. 

- Checks DataReady() before reading.
- Returns ErrDataNotReady without restarting the sensor.
- Validates CRCs on all returned words.
- Waits the required 30 ms after reinit.
- Uses the correct 65535 conversion denominator.
- Waits 30 ms for power-up before starting measurements.
- Removes the unnecessary startup Configure() call.
- Adds CRC and conversion tests.


conejoninja:
I prefer the int32 measure, I think the precision lose in under the margin error of the sensor, and is less ram/cycles hungry. What does anyone else think?

